PR #26, opened 2026-04-14
chore(template): audit crons require PM-routing + GH-issue filing; add UIUX schedule

HongmingWang-Rabbit:
## Summary

Audit agents (Security, QA, UIUX) were running their crons successfully but findings stayed trapped in agent memory — they didn't consistently flow into GitHub issues or reach developers with build ability. This PR rewrites the three audit cron prompts so the post-audit delivery is mandatory, structured, and routed through PM.

## Motivation

Incident seen 2026-04-13 evening: Security Auditor had **13 open critical findings in its memory** including an unauthenticated-plugin-install RCE (`plugins_registry/builtins.py:163-184`). BE picked up a batch earlier in the day as a one-off escalation, then went idle while findings continued to accumulate silently. UIUX Designer was similarly stuck — screenshots unavailable due to missing container libs (#23), falling back to HTML analysis, but no one saw the output.

Root cause: the cron prompts treated post-audit delegation as optional ("if you find issues, report to Dev Lead") and wrote to memory as the primary deliverable. Memory is invisible unless someone reads it.

## Changes

1. **Security Auditor** — replace 12h schedule (`7 6,18 * * *`) with hourly (`17 * * * *`) to match what's actually in the platform DB. Backfill the full hourly prompt into the template (see #24 for the separate template-vs-runtime drift fix).

2. **QA Engineer** — keep 12h cadence; tighten the post-audit routing.

3. **UIUX Designer** — add a schedule block (was previously runtime-only; see #24). Uses hourly cadence. Accepts that Playwright may be unavailable (#23) and falls back to HTML analysis, explicitly noting the limitation in the PM deliverable.

4. **Shared routing contract** — all three audit crons now end with an identical `FINAL STEP — DELIVERABLE ROUTING (MANDATORY every cycle)` block:

   - **(a)** For each CRITICAL/HIGH finding, `gh issue create` with dedup against open issues. Capture issue numbers.
   - **(b)** `delegate_task` to PM with a structured summary (timestamp/SHA, findings by severity, issue numbers, top recommendation). PM owns cross-team routing; audit agents don't DM Dev Lead directly.
   - **(c)** Even on CLEAN cycles: `delegate_task` PM a one-line "clean on SHA X" so audits are observable instead of silent.
   - **(d)** Memory write is a *secondary* record, not the primary deliverable.

## Design rationale

- **GitHub issues as the source of truth** for audit findings — durable, searchable, reviewable, and already wired into the review-agent feedback loop.
[...]

Issue #17, opened 2026-04-14
Provisioner: rogue test workspace in restart loop — no config.yaml, restarts forever


HongmingWang-Rabbit:
## Problem

The platform has a stale test workspace with ID `aaaaaaaa-1234-5678-abcd-000000000001` (name also `aaaaaaaa-...`) leaked into the live `workspaces` table — likely from an E2E test that created but didn't clean up.

Container `ws-aaaaaaaa-123` is now in a continuous restart loop:

```
Runtime: langgraph
Adapter deps already installed (langchain-core)
Traceback (most recent call last):
  File "/app/main.py", line 379, in <module>
    asyncio.run(main())
  ...
  File "/app/config.py", line 224, in load_config
    raise FileNotFoundError(f"Config file not found: {config_file}")
FileNotFoundError: Config file not found: /configs/config.yaml
```

The container exits with code 1, Docker restart policy (`unless-stopped`) brings it back, it crashes again. Cycle repeats indefinitely — 2+ hours observed today. Wastes CPU, fills logs, and `docker ps` shows a permanently 'Restarting (1)' row.

### A) E2E test cleanup (high priority)
Find the E2E test that creates workspace IDs with the `aaaaaaaa-` pattern and ensure it deletes on teardown. Likely in `tests/e2e/test_api.sh` or `test_comprehensive_e2e.sh`.

### B) Provisioner: don't restart containers with permanent errors (medium priority)
After N failed restarts with the same exit code + same error in logs, the provisioner should:
1. Mark the workspace `status = 'failed'` in DB
2. Set `last_sample_error` to the traceback
3. Stop the container (remove restart policy)
4. Broadcast the failure so canvas shows it

Currently a config-missing workspace can restart forever and nobody notices until someone checks `docker ps` manually.

### C) Provision-time validation (low priority, but cheap)
At workspace create, validate that the required config files exist in the config volume before marking the workspace as `online`. A workspace whose config.yaml is missing should fail to provision — not start into a loop.

## Evidence
- 2026-04-13 PST: observed during 18:40 and 19:10 maintenance cycles. Container alive 2+ hours, same traceback in every restart.
- Workspace row present in DB: `curl http://localhost:8080/workspaces` returns the `aaaaaaaa-...` entry.

## Priority
Low-to-medium. Doesn't block real work but pollutes the agent fleet view.

HongmingWang-Rabbit:
[...]

PR #29, opened 2026-04-14
chore(template): Security Auditor DAST must clean up its own test artifacts

HongmingWang-Rabbit:
## Summary
The Security Auditor's hourly DAST cron was creating test workspaces during security probes and never tearing them down, leaking rows into the live `workspaces` table. Adds an explicit `4a. DAST TEARDOWN` step to the cron prompt with a track-and-cleanup pattern for workspaces, secrets, and plugins.

## Root cause
Investigated under #17 (see comment 2026-04-14 02:14 UTC). The cron prompt already cleaned up secrets and plugins it created, but workspace-create probes had no teardown. Over four hourly runs we got four leaked rows with sequential IDs (`aaaaaaaa-…`/`bbbbbbbb-…`/`cccccccc-…`/`dddddddd-…`), each trapped in a restart loop against the missing config.yaml.

Platform log evidence at 2026-04-14 04:27:17 UTC (one sample, representative of every cycle):
```
POST  /admin/secrets                                           — create test secret
DELETE /admin/secrets/SEC-TEST-C5                              — cleanup (good)
POST  /workspaces/<PM>/plugins (404 for non-existent plugin)
DELETE /workspaces/<PM>/plugins/nonexistent-sec-test           — cleanup (good)
ProxyA2A: access denied <PM> → <BE>                            — CanCommunicate bypass probe
[workspace row `dddddddd-1234-5678-abcd-000000000004` appears in DB]
[no corresponding DELETE]
```

## Changes
Single edit to `org-templates/molecule-dev/org.yaml` — adds step 4a to the `Hourly security audit` cron prompt:

- Declare three tracking lists at the start of step 4: `TESTS_WORKSPACES`, `TESTS_SECRETS`, `TESTS_PLUGINS`.
- Every probe that creates an artifact appends its ID to the right list.
- At the end of step 4 (unconditional, runs even on partial audit failure), iterate each list and issue `DELETE`. Uses `|| true` so a failed cleanup doesn't abort the audit.
- Includes the prior-incident narrative inline so future editors understand *why* the step exists.

## Why inline vs. platform-layer
A platform-level circuit-breaker (reject workspace creates from DAST, or TTL test workspaces) is still valid (proposed in #17 options B/C/D). But this template-layer fix is orthogonal and the cheapest immediate stop — it doesn't depend on any code change, and it closes the specific incident class we've already seen four times today.

## Test plan
- [x] `python -c "import yaml; yaml.safe_load(open('org-templates/molecule-dev/org.yaml'))"` — YAML valid.
- [ ] Next hourly cron run (after merge + template re-import): no new `eeeeeeee-…` row appears in `workspaces`.
- [ ] `docker ps | grep -c 'ws-[a-z]\{8\}-123'` stays at 0 across successive runs.
- [ ] Secret/plugin cleanup still works as before (no regression on existing teardown).

## Related
- **#17** — root cause captured; this PR closes the template-layer side. Platform-level circuit-breaker remains in scope for that issue.
- **#26** — audit cron routing (merged). This change sits alongside that structure without conflict.

### DIFF
--- a/org-templates/molecule-dev/org.yaml
+++ b/org-templates/molecule-dev/org.yaml
@@ -250,6 +250,34 @@ workspaces:
                      - CORS: verify Access-Control-Allow-Origin on a cross-origin request
                      - Rate limit headers on /health
 
+                  4a. DAST TEARDOWN (MANDATORY — prevents test-artifact leak into prod DB):
+                      Any workspace, secret, or plugin you CREATE during this audit must be
+                      DELETED before this step exits. Maintain three lists as you go:
+
+                        TESTS_WORKSPACES=""   # workspace IDs you POSTed
+                        TESTS_SECRETS=""      # secret keys you set
+                        TESTS_PLUGINS=""      # "<ws_id>:<plugin_name>" pairs
+
+                      At the end of step 4, iterate each list and DELETE — even if the audit
+                      aborts, the teardown block must run:
+
+                        for ws_id in $TESTS_WORKSPACES; do
+                          curl -s -X DELETE "http://host.docker.internal:8080/workspaces/$ws_id" \
+                            -H "Authorization: Bearer $WORKSPACE_AUTH_TOKEN" > /dev/null || true
+                        done
+                        for key in $TESTS_SECRETS; do
+                          curl -s -X DELETE "http://host.docker.internal:8080/admin/secrets/$key" > /dev/null || true
+                        done
+                        for pair in $TESTS_PLUGINS; do
+                          ws="${pair%:*}"; pl="${pair#*:}"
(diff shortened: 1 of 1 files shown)


HongmingWang-Rabbit (commented):
7-gate verification PASS:
- Gate 1 CI: 6/6 green
- Gate 2 Build: YAML-only, skipped
- Gate 3 Tests: test_api.sh 62/62 pass
- Gate 4 Security: cleanup scope bounded to self-created IDs; uses legacy /admin/secrets alias (valid); no unbounded deletion; no committed secrets
- Gate 5 Design fit: Security Auditor role unchanged; PM-routing unaffected; uses standard REST endpoints
- Gate 6 Line review: correct bash parameter expansion for ws:plugin split; || true guards against missing artifacts; quoting safe for UUIDs
- Gate 7 Canvas: N/A

Merging.

PR #48, opened 2026-04-14
fix(provisioner): stop rogue config-missing restart loop (#17)

HongmingWang-Rabbit:
Resolves #17.

## Part A — E2E cleanup
- New `scripts/cleanup-rogue-workspaces.sh`: deletes workspaces whose id or name starts with known test placeholder prefixes (`aaaaaaaa-`, `bbbbbbbb-`, `cccccccc-`, `test-ws-`) via the API and force-removes the `ws-<id[:12]>` container as a belt-and-suspenders fallback.
- Documented in `tests/README.md`.
- No E2E test currently uses the `aaaaaaaa-` pattern (grep found zero hits in `tests/**`), so the rogue was from a manual/adhoc run. Nothing to patch in the test scripts.
- Ran the script locally against `localhost:8080` — no rogue workspace currently present on the platform (already cleaned up between issue filing and PR).

## Part B — provisioner pre-flight
- New pure helper `provisioner.ValidateConfigSource(templatePath, configFiles)` — returns `ErrNoConfigSource` when neither source supplies `config.yaml`.
- New `Provisioner.VolumeHasFile(ctx, id, relPath)` — probes the workspace's Docker named volume via a throwaway `alpine test -f` container.
- `provisionWorkspace` now calls both: when a restart path passes empty `templatePath` + `configFiles` (auto-restart flow in `workspace_restart.go:215,335`) AND the named volume has no `config.yaml`, the handler marks the workspace `status='failed'` with a clear `last_sample_error` and broadcasts `WORKSPACE_PROVISION_FAILED` instead of handing the doomed container to Docker's `unless-stopped` restart policy.
- Docker restart policy is deliberately unchanged — the happy path still benefits from it.

## Test plan
- [x] `go build ./... && go vet ./...` clean
- [x] `go test -race ./...` all packages pass (provisioner + handlers)
- [x] New tests: `TestValidateConfigSource_*` x6 (configFiles present / empty bytes / template has config / template missing / both empty / config.yaml-as-dir)
- [x] `tests/e2e/test_api.sh` 62/62 pass
- [x] Verified cleanup script reaches the live platform (no matching workspaces currently)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #50, opened 2026-04-14
chore(template): PM system prompt — treat audit summaries as dispatch triggers, not FYIs

HongmingWang-Rabbit:
## Summary
PM's system prompt now explicitly treats audit summaries from Security Auditor / UIUX Designer / QA Engineer as **dispatch triggers**, not informational status updates. Adds an "Audit Routing" section that specifies which dev agent each issue category goes to.

## Motivation — observed orchestration dead-end

Maintenance cron data from 2026-04-14 morning:

| Signal | Value |
|---|---|
| Audit crons still firing | ✅ Security at 14:23 UTC, UIUX at 14:21 UTC (matches `17 * * * *`, `11 * * * *`) |
| PM last received A2A msg | 2026-04-14 14:22 UTC (~7 min ago at time of investigation) |
| **PM last delegated to Dev Lead** | **2026-04-14 05:25 UTC (10+ hours ago)** |
| Dev Lead active_tasks | 0 for 17+ consecutive maintenance cycles |
| BE/FE/DevOps/QA active_tasks | 0 across the same window |
| GitHub PRs still opening | ✅ #47, #48 closing my platform issues — but authored by audits directly via `GITHUB_TOKEN`, not via Dev Lead |

So: the audit → PM chain worked. The PM → Dev Lead → sub-team chain was dead. Audits went around PM because PM wasn't acting on what they sent.

Reading PM's prompt, the delegation rules were written for "when the CEO gives a task" — audit summaries didn't match that pattern, so PM `commit_memory`'d them and stopped. Meanwhile the audit agents (who have `GITHUB_TOKEN` from the global secret) just opened their own PRs to keep throughput up.

## Change
Insert new "Audit Routing" section before "What You Never Do" in `org-templates/molecule-dev/pm/system-prompt.md`:

- Treat every audit summary with open issue numbers as a dispatch trigger.
- Route by category — security → BE, ui/ux → FE, infra → DevOps, qa → QA, mixed → Dev Lead.
- Parallel `delegate_task_async` for multi-category issues.
- Clean cycles (zero issue numbers) — acknowledge only.
- End-of-cycle memory write summarising fan-out and landed PRs.

Also adds one bullet to "What You Never Do": *"Treat an audit summary with open issue numbers as informational."*

## Why this matches the north-star goal
CEO 2026-04-14: "the goal is to keep our team running 24/7 and keep improving". 17 consecutive cycles of full-team idleness despite live findings is the opposite of that. This turns PM from a receptionist into a dispatcher so the audit output reaches the developers who can act on it.

## Related
- **#26** (merged) — audit cron routing contract that made audits send summaries to PM. This PR is the receiver-side follow-up.
- **Ecosystem-watch paused** — CEO directive to focus self-improvement; this PR is in that swim lane.

## Test plan
- [x] YAML/markdown compile — system prompt is plain markdown, no parse step required.
- [ ] Next hourly audit cycle after merge + re-deploy: PM's delegations endpoint shows new dispatches to BE/FE/DevOps.
- [ ] Maintenance cron: agents transition from idle → busy on audit arrival.
- [ ] Some audit PRs now come through Dev Lead instead of audit-direct.

### DIFF
--- a/org-templates/molecule-dev/pm/system-prompt.md
+++ b/org-templates/molecule-dev/pm/system-prompt.md
@@ -20,12 +20,38 @@ You are the PM. The user is the CEO. You own execution — turning CEO directive
 5. **Synthesize across teams.** Your value is combining work from multiple teams into a coherent answer. Don't staple reports together — distill the key findings and decisions.
 6. **Use memory.** `commit_memory` after significant decisions. `recall_memory` at conversation start.
 
+## Audit Routing — Incoming Audit Summaries Are Tasks, Not Status Reports
+
+Security Auditor, UIUX Designer, and QA Engineer run hourly/half-daily audit crons that send you a structured deliverable (per the contract in their cron prompts):
+- audit timestamp + SHA range
+- counts by severity (critical / high / medium / low / clean)
+- **list of GitHub issue numbers filed this cycle**
+- top recommendation
+
+**Every such arrival with issue numbers is a dispatch trigger, not FYI.** The moment you receive one:
+
+1. For each issue number in the summary, `gh issue view <N>` to read the full body and category.
+2. Route each issue to the right dev agent by category:
+   - `security(...)`, auth, crypto, SQL/RCE/path-traversal, missing access control → **Backend Engineer**
+   - `ui`, `ux`, theme, a11y, keyboard-nav, WCAG → **Frontend Engineer**
+   - `infra`, Dockerfile, CI, provisioner, secrets, ops, deployment → **DevOps Engineer**
+   - test suite / coverage / flake / regression → **QA Engineer**
+   - mixed / unclear → **Dev Lead** to split further.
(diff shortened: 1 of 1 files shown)


HongmingWang-Rabbit:
**Scope correction per CEO 2026-04-14 feedback** ('whenever you try to change template, think if this is something can fix in platform level'):

This PR is a **template-level band-aid** — it fixes `molecule-dev`'s PM, but every other user spinning up their own org template would need to re-patch their PM prompt the same way.

The generalised platform-level solution is now tracked in **#51** — a structured `audit_summary` A2A message type + an optional `audit_routing` config in `org.yaml` that the platform can auto-fan-out from, so audit-cron payloads reach the right dev agent without per-template prompt engineering.

**Recommend merging this PR anyway** — it unblocks our team today (17+ cycles of idleness while audits sat unrouted). Once #51 lands and the audit-cron prompts are migrated to emit the structured payload, this change becomes a no-op and can be reverted.

Updated my own working rules (memory): every template PR now also needs a platform-level issue if the bug is systemic.

PR #63, opened 2026-04-14
feat(plugins): split guardrails into 12 modular plugins

HongmingWang-Rabbit:
Replaces the monolithic \`molecule-guardrails\` proposal with 12 single-purpose plugins users install à la carte. A workspace can install just \`molecule-careful-bash\` for safety, or stack the full set for the same posture as the orchestrator.

## Mechanics

\`AgentskillsAdaptor\` (the default plugin adapter) gains a new \`_install_claude_layer()\` step that conditionally:
1. Copies \`hooks/*\` → \`/configs/.claude/hooks/\` (preserving exec bit)
2. Copies \`commands/*.md\` → \`/configs/.claude/commands/\`
3. Merges \`settings-fragment.json\` into \`/configs/.claude/settings.json\`, rewriting \`${CLAUDE_DIR}\` to the workspace's absolute install path. Existing user hooks preserved (deep-merge by event).

All conditional — no-op when the plugin doesn't ship the corresponding files. Existing plugins unchanged. Drift test passes (SDK + runtime mirrors stay in sync).

## The 12 plugins

| Category | Plugin | Purpose |
|---|---|---|
| Hook | molecule-careful-bash | Refuses destructive bash; ships careful-mode skill |
| Hook | molecule-freeze-scope | Locks edits via \`.claude/freeze\` |
| Hook | molecule-audit-trail | Appends every Edit/Write to \`audit.jsonl\` |
| Hook | molecule-session-context | Auto-loads cron-learnings at session start |
| Hook | molecule-prompt-watchdog | Warns on destructive prompt keywords |
| Skill | molecule-skill-code-review | 16-criteria multi-axis review |
| Skill | molecule-skill-cross-vendor-review | Adversarial second-model review |
| Skill | molecule-skill-llm-judge | Deliverable-vs-request scoring |
| Skill | molecule-skill-update-docs | Post-merge doc sync |
| Skill | molecule-skill-cron-learnings | Operational-memory JSONL format |
| Workflow | molecule-workflow-triage | \`/triage\` PR-triage cycle |
| Workflow | molecule-workflow-retro | \`/retro\` weekly retrospective |

## Test plan
- [x] \`python3 -m molecule_plugin validate plugins/molecule-*\` → 13/13 valid
- [x] End-to-end install smoke (asyncio + real InstallContext) on hook / skill / workflow plugins → all produce expected tree, settings paths rewritten, exec bits preserved, 0 warnings
- [x] \`pytest tests/test_plugins_builtins_drift.py\` → passes after the parallel changes to both AgentskillsAdaptor copies
- [x] CLAUDE.md updated with categorized plugin list

## Why split

Per user feedback on the bundled approach: \"any other user can pick what they want instead just for us\". A solo developer might want only \`molecule-careful-bash\`. A team running cron triage might want \`molecule-workflow-triage\` + the skills it composes. An ops team auditing AI activity might want only \`molecule-audit-trail\`. Bundling defeats the modularity the plugin system was built for.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -109,6 +109,28 @@ Shared plugins in `plugins/` are auto-loaded by every workspace:
 - **`ecc`**: General Claude Code guardrails
 - **`browser-automation`**: Puppeteer/CDP-based web scraping and live canvas screenshots (opt-in per workspace — wired into Research + UIUX roles in `org-templates/molecule-dev/org.yaml`)
 
+**Modular guardrails** (Claude Code only — pick what you need, or install several):
+
+*Hook plugins (ambient enforcement at the harness layer)*
+- **`molecule-careful-bash`** — REFUSES `git push --force` to main, `rm -rf` at root, `DROP TABLE` against prod schema. Ships the `careful-mode` skill as documentation.
+- **`molecule-freeze-scope`** — locks edits to a single path glob via `.claude/freeze`. Useful while debugging.
+- **`molecule-audit-trail`** — appends every Edit/Write to `.claude/audit.jsonl` for accountability.
+- **`molecule-session-context`** — auto-loads recent cron-learnings + open PR/issue counts at session start. Pairs with `molecule-skill-cron-learnings`.
+- **`molecule-prompt-watchdog`** — injects warning context when the user prompt mentions destructive keywords ("force push", "drop table", "delete all", etc).
+
+*Skill plugins (on-demand, via the `Skill` tool)*
+- **`molecule-skill-code-review`** — 16-criteria multi-axis review.
+- **`molecule-skill-cross-vendor-review`** — adversarial second-model review (use for noteworthy PRs).
+- **`molecule-skill-llm-judge`** — score whether a deliverable addresses the request.
+- **`molecule-skill-update-docs`** — sync repo docs after merges.
+- **`molecule-skill-cron-learnings`** — defines the operational-memory JSONL format consumed by `molecule-session-context`.
+
+*Workflow plugins (slash commands that compose skills)*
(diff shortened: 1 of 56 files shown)


HongmingWang-Rabbit:
## Code-review findings — 1 🔴

### 🔴 Missing `import json` in SDK builtins
`sdk/python/molecule_plugin/builtins.py` uses `json.loads` / `json.dumps` in `_merge_settings_fragment` (L173, L180) and `_rewrite_hook_paths` (L193), but the module has no `import json` at the top.

Existing SDK tests pass because none exercise `_install_claude_layer` with a `settings-fragment.json`. As soon as a plugin ships `settings-fragment.json` and is installed via the SDK path (every one of the 12 new plugins except skill-only ones), it will raise `NameError: name 'json' is not defined`.

The workspace-template copy (`workspace-template/plugins_registry/builtins.py` L43) has `import json` correctly — this is pure drift that the syntactic drift test didn't catch.

**Fix:** add `import json` at the top of `sdk/python/molecule_plugin/builtins.py` (next to the existing `import os`, `import shutil`, `import subprocess`).

**Suggestion:** add a regression test that imports a plugin with a `settings-fragment.json` and asserts the merge completes without NameError, so future drift is caught by tests rather than in production.

Blocking merge on this. 🟡 and 🔵 aside.

### 🟡 Warning
- `_copy_dir_files` branch at L158/L267: the `only_suffix` / `executable_suffix` combined filter is subtle — when `only_suffix=".md"` and a `.py` file is present, it falls through because `executable_suffix` is None, but reading the condition inversely is confusing. Consider separating "allowed suffixes" into a set argument.

### 🔵 Suggestion
- Two `builtins.py` copies diverging is a recurring cost. Consider a shared `_claude_layer.py` module imported by both (only `append_to_memory` / `InstallContext` differ, and those are parameters). Current drift test catches shape drift but missed this import drift.

PR #64, opened 2026-04-14
fix(secrets): auto-refresh global_secrets on workspace restart (#15)

### LINKED ISSUE #15: Provisioner: auto-refresh CLAUDE_CODE_OAUTH_TOKEN from global_secrets on workspace restart
## Problem

The provisioner reads `CLAUDE_CODE_OAUTH_TOKEN` (and all env vars derived from `global_secrets`) **once at container create**. When the OAuth token rotates (every few days for Claude Max), every existing workspace container keeps the stale token in its process env. The Claude CLI then fails with:

```
Failed to authenticate. API Error: 401 {"type":"error","error":{"type":"authentication_error","message":"Invalid authentication credentials"}}
```

…which surfaces in the SDK as `Command failed with exit code 1` → SDK retries 3x → user-visible agent error. Tier-3 Opus agents (PM, Dev Lead) hit it first because they carry more load.

**Incident 2026-04-13**: PM + Dev Lead went down with this. Manual remediation required: `POST /admin/secrets` (new token) + loop over all workspaces with `POST /workspaces/:id/restart`. Full recovery took ~15 min.

## Proposed fix

On `POST /workspaces/:id/restart`:
1. Re-read `global_secrets` + `workspace_secrets` (same code path as initial provision).
2. Pass the fresh env to `provisioner.Start()` on recreate.

Current behaviour in `workspace_restart.go`: reuses the existing container config, inheriting the old env. Needs to re-compute env from DB, same as `provisionWorkspace()` does at create time.

Bonus: add a `POST /admin/secrets/:key/refresh-all` endpoint that (a) updates the secret, (b) restarts every workspace that has the secret in scope. Removes the manual restart loop.

## Files likely touched

- `platform/internal/handlers/workspace_restart.go` (re-read env)
- `platform/internal/handlers/secrets.go` (new refresh-all endpoint)
- tests

## Priority

High — OAuth tokens rotate on a schedule we don't control. This will keep happening.

## Evidence

- 2026-04-13 17:08 PST: PM logged `Fatal error in message reader: Command failed with exit code 1` on every request; `docker exec ws-f3531ba4-d5b claude --print --model sonnet` returned 401.
- Env inspection confirmed: container had `CLAUDE_CODE_OAUTH_TOKEN=sk-ant-oat01-H6UrPen...` (stale); host `.credentials.json` had `sk-ant-oat01-B0Lazd...` (valid).
- Remediation: CEO issued a new token → `POST /admin/secrets` + restart loop over all 12 workspaces → all green.

HongmingWang-Rabbit:
Resolves #15.

## Problem
Provisioner reads `CLAUDE_CODE_OAUTH_TOKEN` (and all env vars derived from
`global_secrets`) once at `ContainerCreate`. When the Claude Max OAuth token
rotates, existing workspaces keep the stale token and the Claude CLI dies
with `401 Invalid authentication credentials`. Remediation was a manual
`POST /admin/secrets` + per-workspace restart loop (incident 2026-04-13:
~15 min recovery, 12 workspaces).

## Fix
Restart-time re-read of secrets was already correct — `provisionWorkspaceOpts`
re-queries `global_secrets` + `workspace_secrets` on every call, and
`provisioner.Start()` always force-removes + recreates the container. The
missing piece was the trigger.

`SetGlobal` / `DeleteGlobal` now enqueue `RestartByID` for every non-paused,
non-removed, non-external workspace that does NOT shadow the key with a
workspace-level override. Mirrors the behaviour of workspace-scoped
`Set` / `Delete` (which already auto-restart).

Effectively implements the "bonus" `refresh-all` step from the issue, without
a new endpoint: updating the global secret via the existing endpoint
propagates automatically.

## Files
- `platform/internal/handlers/secrets.go` — `restartAllAffectedByGlobalKey` helper + wiring in `SetGlobal` / `DeleteGlobal`
- `platform/internal/handlers/secrets_test.go` — two sqlmock tests covering set and delete paths, verifying `NOT IN (workspace_secrets WHERE key=...)` filter

## Test plan
- [x] `go build ./... && go vet ./...` clean
- [x] `go test -race ./...` pass (platform)
- [x] New tests `TestSetGlobal_AutoRestartsAffectedWorkspaces` and `TestDeleteGlobal_AutoRestartsAffectedWorkspaces` assert restart callback fires for affected workspaces
- [ ] E2E verification pending live platform: set global secret → confirm existing workspaces restart and pick up new value

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/platform/internal/handlers/secrets.go
+++ b/platform/internal/handlers/secrets.go
@@ -1,6 +1,7 @@
 package handlers
 
 import (
+	"context"
 	"database/sql"
 	"log"
 	"net/http"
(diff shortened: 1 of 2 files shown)


HongmingWang-Rabbit:
**noteworthy: secrets-handling** — merging after 7-gate verification.

7 gates clean. Code-review: 0 🔴 / 1 🟡 / 4 🔵.

The 🟡 — unbounded goroutine fan-out on global-secret rotation — is a real scaling concern (1000 workspaces × 1 fan = correlated restart storm) but doesn't trigger at current scale (<50 workspaces). Tracked as a follow-up rather than blocker since:
1. The base-case fix (rotated secrets actually propagate) is the user-facing win
2. Adding a worker-pool gate would expand scope beyond the issue
3. Throttling is a 1-line fix when needed

Cross-vendor review attempted but tooling unavailable in cron context.

Other 🔵 nits worth a follow-up cleanup PR: explicit override-shadow SQL assertion, ctx timeout on the affected-workspaces query, log warning on nil restartFunc.

PR #65, opened 2026-04-14
feat(platform): inject restart context system message (#19 Layer 1)

### LINKED ISSUE #19: Workspace restart context — platform-generated state snapshot + optional restart_prompt
## Proposal

When a workspace container restarts (for any reason — OAuth rotation, crash recovery, CEO-triggered), the platform should inject a **restart context** as the agent's first message on wake. Two layers:

### Layer 1 — platform-managed context (automatic)

Auto-generated by the platform from DB state + diff against last known session. Injected as a system message before the agent processes anything else:

```
=== WORKSPACE RESTART CONTEXT ===
Restart at: <timestamp>
Previous session ended: <timestamp> (<duration> ago)

Env vars now available: <keys...>
  (* = newly added since your last run)

Pending delegations you initiated:
  - <id> → <target> (status, blocker if any)
    Note if a resolvable blocker was present: 'X is now available — consider retrying'

Global secrets changed: <added/removed since last run>

Infra changes: <services coming online/going offline — Temporal, etc.>
```

**Platform owns the data.** Agent doesn't compute it. Consistent across all runtimes.

### Layer 2 — org-template restart_prompt (optional, in org.yaml)

Mirror of the existing `initial_prompt`, but runs on every restart after the first:

```yaml
defaults:
  restart_prompt: |
    Your workspace was restarted. Review the RESTART CONTEXT above.
    If any completed delegations show a resolvable blocker, re-queue them.
    Otherwise resume normal operation — wait for new tasks.
```

Teams customize *behaviour*. Platform provides *data*. Same separation as today's `initial_prompt`.

## Why this design

- Solves #18 (notify agents when a secret is added) generically — restart is the natural retry point.
- Solves the 2026-04-13 incident where 3 infra delegations stalled for 2+ hours because the blocker was resolved but no one told the team.
- Composes with #12 (persistent session) — on session-preserve restart, the context prompt is still injected as the first new message. Agent sees their prior state + the restart delta.
- Cheap: platform already has broadcaster + DB history. One message queued on restart.

## Supersedes

- **#18** (notify-on-secret-change): this is the general solution. Close #18 when this lands.

## Related

- **#12** Session persistence across restart — compatible; this RESTART_CONTEXT message fires regardless of whether sessions persist.
- **#15** Auto-refresh OAuth on restart — also compatible; that issue handles the env refresh, this one surfaces the change to the agent.

## Files likely touched

- `platform/internal/handlers/workspace_restart.go` — generate + queue the context message before container becomes 'online'
- `platform/internal/provisioner/provisioner.go` — read `restart_prompt` from org.yaml (alongside initial_prompt)
- `org-templates/molecule-dev/org.yaml` — ship a default restart_prompt
- Workspace SDK needs no changes — the message arrives as regular A2A, agents handle it with existing message loop
- tests

## Priority

High — composes with multiple in-flight issues (#12, #15, #18) and fixes a real class of silent idleness we observed today.

HongmingWang-Rabbit:
Resolves #19 partially — Layer 1 (platform-managed context) only. Layer 2 (user restart_prompt) deferred to a follow-up issue.

## Change
After workspace restart and re-registration, platform sends an A2A `message/send` to the workspace as a synthetic user message containing:
- restart timestamp
- previous session end timestamp + duration
- env-var keys now available (keys only, no values)

Covers both the HTTP `POST /workspaces/:id/restart` path and the programmatic `RestartByID` path (auto-restart after secret change, container death, liveness timeout).

Metadata on the A2A message:
\`\`\`
"metadata": { "kind": "restart_context", "source": "platform", "layer": 1, "restart_context": true }
\`\`\`

## Skip path
If the workspace doesn't come back online within 30s, log and drop. Restart response is unaffected.

## Safety
- Values are NEVER included in the rendered message — only keys.
- System caller prefix \`system:restart-context\` bypasses access control and caller-token check.

## Test plan
- [x] \`go build ./... && go vet ./...\` pass
- [x] \`go test -race ./...\` pass
- [x] new unit tests cover humanDuration, message render (with + without prior session), and JSON-RPC payload shape
- [ ] tests/e2e/test_api.sh — skipped locally (requires running platform + Postgres); CI will run

### DIFF
--- a/platform/internal/handlers/restart_context.go
+++ b/platform/internal/handlers/restart_context.go
@@ -1,0 +1,240 @@
+// Package handlers — restart_context.go implements Layer 1 of issue #19:
+// after a workspace is restarted and comes back online, the platform
+// generates a state snapshot (timestamp, previous session end, env-var
+// keys now available) and delivers it as a synthetic A2A message/send
+// so the agent sees what changed across the restart boundary.
+//
+// Layer 2 (user-defined restart_prompt via config.yaml / org.yaml) is
+// out of scope for this file — tracked as a separate follow-up issue.
+package handlers
+
+import (
+	"context"
+	"database/sql"
+	"encoding/json"
+	"fmt"
+	"log"
+	"sort"
+	"time"
+
+	"github.com/Molecule-AI/molecule-monorepo/platform/internal/db"
+	"github.com/google/uuid"
+)
+
+// restartContextOnlineTimeout bounds how long we wait for a workspace
+// to re-register after restart before dropping the context message.
+// The Restart HTTP handler has already returned 200 by the time this
+// waiter runs, so a timeout here is purely a best-effort skip.
+const restartContextOnlineTimeout = 30 * time.Second
+
+// restartContextOnlinePollInterval is the poll cadence while waiting
+// for WORKSPACE_ONLINE. 500ms keeps the typical-case latency low
+// without hammering Postgres.
+const restartContextOnlinePollInterval = 500 * time.Millisecond
+
+// restartContextData captures the platform-computed snapshot that will
+// be rendered into a human-readable message. Keeping it as a struct
+// (rather than building the string inline) makes the builder
+// unit-testable without stubbing time/DB calls.
+type restartContextData struct {
+	RestartAt     time.Time
+	PrevSessionAt time.Time // zero value = no prior session recorded
+	EnvKeys       []string  // sorted list of env-var keys (no values)
+}
+
+// buildRestartContextMessage renders the restart context into the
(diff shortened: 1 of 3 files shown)


HongmingWang-Rabbit:
Merged. **noteworthy: access-control-bypass** — introduces `system:restart-context` caller prefix that bypasses CanCommunicate + A2A token validation (via existing `systemCallerPrefixes` allowlist in a2a_proxy.go:72). Bypass is principled (reuses existing `system:` scheme, not novel). 7/7 gates passed, 0 critical findings. Cross-vendor review unavailable (codex tier / gemini quota). Post-merge HEAD: 3ea8cda.

PR #78, opened 2026-04-14
feat(platform): TenantGuard middleware — public repo's only SaaS hook (Phase 32 PR #1)

HongmingWang-Rabbit:
## Summary
Phase 32 (Cloud SaaS) foundation. The **public repo's only SaaS hook**: one middleware, one env var. All orgs/signup/billing/Fly-provisioning code lives in the private \`molecule-controlplane\` repo (n8n-style open-core).

- \`TenantGuard\` reads \`MOLECULE_ORG_ID\` at boot. If set → every non-allowlisted request must carry matching \`X-Molecule-Org-Id\` header or gets **404** (not 403 — don't leak tenant existence). If unset → passthrough; self-hosted behavior unchanged.
- Allowlist is exact-match (not prefix): \`/health\` and \`/metrics\` only, so Fly Machines health probes + Prometheus scrape work regardless.
- Wired into \`router.Setup\` after metrics so rejected requests still land on the 4xx counter.

## Why 404 not 403
An attacker probing multiple org subdomains should not be able to distinguish "this tenant exists but I don't have access" from "this route doesn't exist." 404 + empty body makes both indistinguishable from random noise.

## Test plan
- [x] Unit tests (6): unset-is-passthrough, matching header, mismatched 404 with empty body, missing 404, allowlist bypass, allowlist-is-exact-match
- [x] Full platform \`go test ./...\` green
- [x] Self-hosted behavior: \`MOLECULE_ORG_ID\` unset → no change (passthrough verified)

## What's NOT here (deferred to private control-plane repo)
- \`organizations\` table / signup / WorkOS / Stripe / Fly Machines provisioner / subdomain router — all in \`molecule-controlplane\` (private), not here


HongmingWang-Rabbit:
noteworthy: saas-foundation — tenant isolation middleware, Phase 32 PR #1. Cross-vendor-review tooling still unavailable (recurring gap logged tick-6); proceeding per tick-7 precedent. Self-reviewed: 6 tests cover passthrough / match / mismatch-404 / missing-404 / allowlist / exact-match; default-unset behavior preserves self-hosters.

PR #94, opened 2026-04-15
fix(security): C6 — block loopback IP literals in /registry/register

HongmingWang-Rabbit:
## Summary

- Extends `validateAgentURL` to block `127.0.0.0/8` (loopback) in addition to the existing `169.254.0.0/16` (link-local) block
- RFC-1918 ranges (`10.x`, `172.16.x`, `192.168.x`) remain **allowed** — Docker container networking requires them
- `localhost` by hostname remains allowed; only IP-literal loopback is blocked

## Attack scenario closed

A workspace that self-registers with `http://127.0.0.1:PORT` on **first INSERT** (before the provisioner URL is established) would cause the A2A proxy to send requests back to the platform itself — SSRF via the workspace registry.

## Why RFC-1918 is kept open

The existing code comment and test suite document this as a deliberate decision: Docker's default bridge network uses `172.17.0.0/16` (within `172.16.0.0/12`), and private-LAN deployments use `10.x` and `192.168.x` addresses. Blocking those ranges would break all Docker-based and corporate-network agent deployments.

## Why loopback is safe to block

The provisioner stores `http://127.0.0.1:<port>` URLs via direct `UPDATE workspaces SET url` SQL — it never calls `/registry/register`. The `ON CONFLICT … CASE` that preserves provisioner URLs is unaffected. Legitimate local-dev agents use `http://localhost:PORT` (hostname form), which is not caught by `net.ParseIP`.

## Test plan

- [x] `TestValidateAgentURL`: removed `"valid localhost http"` (127.0.0.1 now correctly rejected), added `"valid localhost name"` + three loopback-block assertions
- [x] All existing registry tests unchanged — provisioner URL tests mock the DB row directly, not via registration payload
- [ ] Run `cd platform && go test -race ./internal/handlers/...` locally

### DIFF
--- a/platform/internal/handlers/registry.go
+++ b/platform/internal/handlers/registry.go
@@ -30,7 +30,14 @@ func NewRegistryHandler(b *events.Broadcaster) *RegistryHandler {
 //
 // Allowed: http:// or https:// only (no file://, ftp://, etc.).
 // Blocked: 169.254.0.0/16 (link-local — AWS/GCP/Azure metadata endpoints).
-// Allowed: RFC-1918 private ranges (Docker networking uses 172.16–31.x.x).
+// Blocked: 127.0.0.0/8  (loopback — self-SSRF: a registered loopback URL
+//
+//	redirects A2A traffic back to the platform itself).
+//
+// Allowed: RFC-1918 private ranges 10.x, 172.16.x, 192.168.x — Docker
+//
+//	container networking uses these; blocking them would break any
+//	private-network or Docker-based deployment.
 //
 // Returns a non-nil error string suitable for including in a 400 response.
 func validateAgentURL(rawURL string) error {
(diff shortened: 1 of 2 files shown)


HongmingWang-Rabbit:
## 🚨 DEV LEAD — DO NOT MERGE (Security Block)

**Blocked by Security Auditor finding C6 — incomplete fix.**

### What this PR fixes
- Loopback range `127.0.0.0/8` is now blocked in `validateAgentURL` ✅

### What must be added before merge
SSRF is still possible via the following unblocked ranges:
- `10.0.0.0/8` (RFC-1918)
- `172.16.0.0/12` (RFC-1918)
- `192.168.0.0/16` (RFC-1918)
- `169.254.0.0/16` (IMDS / link-local — cloud metadata endpoint)

Push an additional commit extending `validateAgentURL` to reject all four ranges. Return `400 Bad Request` with a clear message. Tag Dev Lead when ready for re-review.

### Context
11 critical auth findings remain open across the platform. A separate PR for auth middleware on `/workspaces/*`, `/admin/secrets`, and `/workspaces/:id/memories` is being assigned in parallel — do not conflate with this PR.

PR #112, opened 2026-04-15
fix(security): close IPv6 SSRF gap in validateAgentURL (C6)

HongmingWang-Rabbit:
## Summary

- **Root cause:** `validateAgentURL` in `registry.go` only blocked `169.254.0.0/16`. Go's `(*IPNet).Contains()` does not match pure IPv6 addresses against IPv4 CIDRs, leaving three IPv6 ranges fully open as SSRF vectors.
- **Gap confirmed by QA:** `http://[::1]:8080`, `http://[fe80::1]:8080`, and `http://[fd00::1]:8080` all passed through without error.
- **Fix:** Restructured the single-CIDR check into a `blockedRanges` loop and added three IPv6 entries:
  - `fe80::/10` — IPv6 link-local (same threat class as `169.254.0.0/16`; cloud metadata analogue)
  - `::1/128` — IPv6 loopback
  - `fc00::/7` — IPv6 ULA (RFC-4193 private)
- **IPv4-mapped IPv6** (`::ffff:169.254.169.254`) requires no extra range: Go normalises these to IPv4 via `To4()` before `Contains()` runs, so they're already caught by the `169.254.0.0/16` entry. Documented in both code comment and a test case.
- **No allowlist changes:** `127.0.0.1` and RFC-1918 ranges remain permitted (Docker networking requirement).

## Test plan

- [ ] `go test -race -run TestValidateAgentURL ./platform/internal/handlers/` — 4 new cases added: `[::1]`, `[fe80::1]`, `[fd00::1]`, `[::ffff:169.254.169.254]` all expect `wantErr: true`
- [ ] Existing valid-URL cases (`172.18.0.5`, `127.0.0.1`, `10.x`, `192.168.x`) remain `wantErr: false`
- [ ] Full `go test -race ./...` in `platform/`
- [ ] CI `platform-build` gate

### DIFF
--- a/platform/internal/handlers/registry.go
+++ b/platform/internal/handlers/registry.go
@@ -29,10 +29,19 @@ func NewRegistryHandler(b *events.Broadcaster) *RegistryHandler {
 // cloud metadata services or other internal infrastructure.
 //
 // Allowed: http:// or https:// only (no file://, ftp://, etc.).
-// Blocked: 169.254.0.0/16 (link-local — AWS/GCP/Azure metadata endpoints).
-// Allowed: RFC-1918 private ranges (Docker networking uses 172.16–31.x.x).
+// Blocked:
+//   - 169.254.0.0/16  IPv4 link-local (AWS IMDSv1/v2, GCP, Azure metadata)
+//   - fe80::/10        IPv6 link-local — same threat class as 169.254.x.x
+//   - ::1/128          IPv6 loopback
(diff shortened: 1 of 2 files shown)


HongmingWang-Rabbit:
Superseded by #119 (cherry-picked onto a clean base from main). IPv6 SSRF fix (e4e6634) is included in #119 with the correction that the IPv4 loopback and RFC-1918 blocks are preserved — this branch had inadvertently removed them.

Issue #121, opened 2026-04-15
security(C19): GET /workspaces topology protected in running binary but not in repo code — divergence risk


HongmingWang-Rabbit:
## Context

The running platform binary at `host.docker.internal:8080` returns **401 "admin auth required"** for unauthenticated `GET /workspaces`. However, the current repo code (`router.go` line 75) registers it without auth:

\`\`\`go
// router.go line 75 (repo)
r.GET("/workspaces", wh.List)   // no middleware — open
\`\`\`

## Evidence

\`\`\`
curl -s http://host.docker.internal:8080/workspaces
→ {"error":"admin auth required"}   # returns 401
\`\`\`

PR #110 e2e test comment confirms this is intentional in the binary:
> "GET /workspaces is reachable without a bearer token again" (after all tokens revoked → fail-open)

This indicates the binary was compiled from a version of `router.go` where `GET /workspaces` is inside the AdminAuth group, but this change was not committed to the main branch.

## Risk

If the platform binary is ever **rebuilt from the current repo** (e.g., CI publishes a new `ghcr.io/molecule-ai/platform:latest`), `GET /workspaces` becomes open again, exposing all workspace IDs, URLs, agent cards, hierarchy, and runtime details to unauthenticated callers.

This creates an oracle for the **unauthenticated PATCH /workspaces/:id** issue (see #120) — an attacker could enumerate UUIDs via GET then mutate via PATCH.

## Proposed fix

Commit the AdminAuth gating for `GET /workspaces` to the repo code to match the running binary:

\`\`\`go
// Move from root router to wsAdmin group
wsAdmin := r.Group("", middleware.AdminAuth(db.DB))
wsAdmin.GET("/workspaces", wh.List)
wsAdmin.POST("/workspaces", wh.Create)
wsAdmin.DELETE("/workspaces/:id", wh.Delete)
\`\`\`

Canvas frontend already receives a workspace token via the auth flow (Phase 30.1); it should pass that token on the `/workspaces` fetch.

## Related
- Issue #120 (unauthenticated PATCH — the direct exploitation vector this protects against)
- C19 tracking issue (if it exists separately)

Discovered by automated security audit 2026-04-15 (SHA `e4e6634`).

HongmingWang-Rabbit:
Resolved — repo code now matches the running binary.

Verified on current main (commit de6ebe2+):
\`\`\`
[...]

Issue #120, opened 2026-04-15
security(auth): PATCH /workspaces/:id unauthenticated — allows name/role/tier/parent_id/runtime mutation


HongmingWang-Rabbit:
## Affected file
`platform/internal/router/router.go` line 77, `platform/internal/handlers/workspace.go::Update`

## Vulnerability

`PATCH /workspaces/:id` is registered on the root router without any authentication middleware:

\`\`\`go
// router.go line 75-77
r.GET("/workspaces", wh.List)
r.GET("/workspaces/:id", wh.Get)
r.PATCH("/workspaces/:id", wh.Update)   // ← no middleware
\`\`\`

The `Update` handler accepts and applies any of the following fields without authentication:

| Field | Impact |
|-------|--------|
| `name` | cosmetic |
| `role` | cosmetic |
| `tier` | resource limits — tier 4 = 4 GB RAM / 4 CPU |
| **`parent_id`** | **changes A2A communication hierarchy** |
| `runtime` | changes container image on next restart |
| `workspace_dir` | redirects host filesystem bind-mount |

The handler also returns `{"status":"updated"}` with HTTP 200 even when 0 rows are affected (no rows-affected check).

## Most critical sub-case: parent_id bypass

`CanCommunicate(callerID, targetID)` is the A2A access control gate. It allows communication between siblings and parent/children. An attacker who knows two workspace UUIDs (A and B) can call:

\`\`\`bash
curl -X PATCH http://host/workspaces/{ws-A} \
  -d '{"parent_id": "{ws-B}"}'  # no token required
\`\`\`

This makes B the parent of A, granting A↔B A2A communication that would otherwise be blocked.

## Reproduction

\`\`\`bash
# Confirmed live — returns {"status":"updated"} without any token
curl -s -X PATCH http://host.docker.internal:8080/workspaces/00000000-0000-0000-0000-000000000000 \
  -H "Content-Type: application/json" \
  -d '{"name":"owned","tier":4}'
# → {"status":"updated"}
\`\`\`

## Exploitability note

Workspace UUIDs are random v4. An attacker needs to discover a target UUID before exploiting this. UUIDs may be observable from A2A payloads, delegation results, agent cards, or the restart-context messages that enumerate env-var key names.

Note: `GET /workspaces` (which would reveal all UUIDs) appears to be behind AdminAuth in the running binary (C19 fix), but this protection is absent in the current repo code — see related finding.

## Proposed fix

[...]

Issue #113, opened 2026-04-15
security(IDOR): schedules PATCH/DELETE don't bind scheduleId to workspace — cross-workspace mutation possible


HongmingWang-Rabbit:
## Affected files
`platform/internal/handlers/schedules.go`

## Vulnerability

`Update` (PATCH) and `Delete` (DELETE) for workspace schedules filter only by `scheduleId`, not by `workspace_id`. An authenticated workspace can modify or delete a schedule belonging to any other workspace if it knows the target schedule's UUID.

**`Delete` — line ~224:**
\`\`\`go
result, err := db.DB.ExecContext(ctx,
    \`DELETE FROM workspace_schedules WHERE id = \$1\`, scheduleID)
\`\`\`

**`Update` — line ~202:**
\`\`\`sql
UPDATE workspace_schedules SET ...
WHERE id = \$1
\`\`\`

There is also a pre-update read that leaks another workspace's cron expression and timezone (line ~167):
\`\`\`go
\`SELECT cron_expr, timezone FROM workspace_schedules WHERE id = \$1\`, scheduleID,
\`\`\`

**Contrast with `RunNow` (correctly scoped):**
\`\`\`go
\`SELECT prompt FROM workspace_schedules WHERE id = \$1 AND workspace_id = \$2\`,
scheduleID, workspaceID,
\`\`\`

## Exploit scenario

1. Workspace A obtains the UUID of workspace B's schedule (e.g., from a delegation result, activity log, or shared context leak).
2. Workspace A calls `DELETE /workspaces/{ws-A-id}/schedules/{schedule-B-uuid}` with its own bearer token.
3. WorkspaceAuth validates token against workspace A ✓, but the DELETE query has no `workspace_id` bind — it deletes workspace B's schedule.

Schedule UUIDs are random v4, so exploitation requires prior UUID disclosure. However, the fix is trivially simple and the authorization bypass is clear.

## Proposed fix

Add `AND workspace_id = $2` to both queries. For Update, also add it to the pre-update read:

\`\`\`go
// Update pre-read
\`SELECT cron_expr, timezone FROM workspace_schedules WHERE id = \$1 AND workspace_id = \$2\`,
scheduleID, workspaceID,

// Update
WHERE id = \$1 AND workspace_id = \$2

// Delete
\`DELETE FROM workspace_schedules WHERE id = \$1 AND workspace_id = \$2\`,
scheduleID, workspaceID,
\`\`\`

## Related
[...]

PR #124, opened 2026-04-15
fix(security): #113 — bind scheduleId to workspace (IDOR)

HongmingWang-Rabbit:
Closes #113 (HIGH severity — cross-workspace IDOR).

## Root cause
\`Update\` and \`Delete\` on \`workspace_schedules\` filtered by \`scheduleId\` only. A member of workspace A with a cached scheduleId from workspace B could mutate or delete that foreign row — the \`wsAuth\` middleware only gates the :id path parameter, not the sub-resource.

## Fix
Bind both :id and :scheduleId in every SQL that touches a schedule:
- Pre-update SELECT: \`WHERE id = \$1 AND workspace_id = \$2\`
- UPDATE: \`WHERE id = \$1 AND workspace_id = \$2\`
- DELETE: \`WHERE id = \$1 AND workspace_id = \$2\`

\`RunNow\` and \`History\` already had the correct two-key filter — no changes.

## Test plan
- [x] 2 new sqlmock tests: \`TestScheduleUpdate_RejectsCrossWorkspaceIDOR\`, \`TestScheduleDelete_RejectsCrossWorkspaceIDOR\` — Bob attacks Alice's scheduleId → 0 rows affected → 404
- [x] Existing \`TestRuntimeSchedule_HasSourceRuntime\`, \`TestList_IncludesSourceColumn\` still green
- [x] \`go test -race ./platform/...\` clean locally

⚠️ **CI currently blocked** by GitHub Actions org spending-limit cap — not a code failure. See #110 / #119 for the same block. Verified locally.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #125, opened 2026-04-15
fix(security): close unauthenticated PATCH /workspaces/:id (#120) + schedule IDOR (#113)

HongmingWang-Rabbit:
## ⚠️ Security — HIGH + MEDIUM findings from 2026-04-15 audit cycle

---

### Issue #120 — HIGH: Unauthenticated PATCH /workspaces/:id

**Attack vector:** No credentials required. Any caller who knows (or guesses) a workspace UUID can write to sensitive fields.

| Field | Impact |
|---|---|
| `tier` | Resource limit escalation (tier 4 = 4 GB RAM) |
| `parent_id` | Rewrites A2A communication hierarchy, bypasses `CanCommunicate` |
| `runtime` | Swaps container image on next restart |
| `workspace_dir` | Redirects host filesystem bind-mount to attacker-controlled path |

**Root cause:** `r.PATCH("/workspaces/:id", wh.Update)` was registered on the root router, outside the `wsAdmin` AdminAuth group, with a stale comment claiming "canvas doesn't carry a bearer token for position-persist calls." Canvas already requires AdminAuth for `GET /workspaces` list load, so the token is always available.

**Fix:**
- `router.go`: Remove standalone `r.PATCH(...)` from root router; add `wsAdmin.PATCH("/workspaces/:id", wh.Update)` inside the existing AdminAuth group
- `workspace.go` `Update()`: Add workspace-existence guard (`SELECT EXISTS`) at handler entry — previously returned `200 {"status":"updated"}` for nonexistent IDs (zero-row silent UPDATE)

---

### Issue #113 — MEDIUM: Schedule IDOR (carry-over, prior cycle)

**Attack vector:** Authenticated workspace token can PATCH or DELETE schedules belonging to **any** workspace, not just its own.

**Root cause:** `Update` and `Delete` handlers queried `WHERE id = $1` using only `scheduleID`, without binding `workspace_id`.

**Fix:**
- `schedules.go` `Update()`: Add `workspaceID := c.Param("id")` and bind to all queries: `WHERE id = $1 AND workspace_id = $2` / `WHERE id = $1 AND workspace_id = $8`
- `schedules.go` `Delete()`: Same — `WHERE id = $1 AND workspace_id = $2`

---

### Issue #121 — MEDIUM: GET /workspaces binary-code divergence (informational)

The audit noted the running binary gates `GET /workspaces` with AdminAuth but the repo code appeared open. **Current `main` HEAD already has `wsAdmin.GET("/workspaces", wh.List)` in the AdminAuth group** — this divergence was resolved before this PR. DevOps should confirm the next binary build uses current `main` and not an older SHA (e4e6634).

---

## Files changed
- `platform/internal/router/router.go` — PATCH moved into AdminAuth group; comments updated
- `platform/internal/handlers/workspace.go` — 404-guard added to Update handler
- `platform/internal/handlers/schedules.go` — workspace_id binding in Update + Delete

## Test plan
- [ ] `go test -race ./...` in `platform/` — all 487 tests pass
- [ ] `curl -X PATCH http://localhost:8080/workspaces/<uuid> -d '{"tier":4}'` → `401 Unauthorized` (no token)
- [ ] `curl -X PATCH http://localhost:8080/workspaces/<uuid> -d '{"tier":4}' -H "Authorization: Bearer <token>"` → `200` (valid token)
- [ ] `curl -X PATCH http://localhost:8080/workspaces/00000000-0000-0000-0000-000000000000 -H "Authorization: Bearer <token>" -d '{"name":"x"}'` → `404` (nonexistent ID)
- [ ] PATCH schedule from wrong workspace → `404` (workspace_id mismatch)
- [ ] DELETE schedule from wrong workspace → `404`
- [ ] Canvas node drag → position saved correctly (AdminAuth token present)

### DIFF
--- a/platform/internal/handlers/handlers_extended_test.go
+++ b/platform/internal/handlers/handlers_extended_test.go
@@ -73,6 +73,11 @@ func TestExtended_WorkspaceUpdate(t *testing.T) {
 	broadcaster := newTestBroadcaster()
 	handler := NewWorkspaceHandler(broadcaster, nil, "http://localhost:8080", "/tmp/configs")
 
+	// #120 fix: existence check runs first — workspace must be found before updates proceed.
+	mock.ExpectQuery("SELECT EXISTS").
+		WithArgs("ws-upd").
+		WillReturnRows(sqlmock.NewRows([]string{"exists"}).AddRow(true))
+
 	// Expect name update
 	mock.ExpectExec("UPDATE workspaces SET name").
 		WithArgs("ws-upd", "New Name").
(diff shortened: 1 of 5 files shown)


HongmingWang-Rabbit:
## ⚠️ Canvas regression — PATCH will 401 in production

I verified locally: \`canvas/src/lib/api.ts\` uses \`credentials: "include"\` (session cookies), **not** bearer tokens. Moving PATCH into the \`wsAdmin\` AdminAuth group means every \`PATCH /workspaces/:id\` call from canvas (drag-to-reposition, name-edit, etc.) will return 401 once any workspace on the platform has a live token:

\`\`\`typescript
// canvas/src/lib/api.ts
const res = await fetch(\`\${PLATFORM_URL}\${path}\`, {
  method,
  headers,      // ← no Authorization: Bearer
  body: body ? JSON.stringify(body) : undefined,
  credentials: "include",   // ← cookies only
});
\`\`\`

\`middleware/wsauth_middleware.go:74-81\` AdminAuth only checks \`Authorization: Bearer <token>\`; it doesn't read cookies or fall back to the session. So the note in this PR — *"canvas position-persist uses the same AdminAuth token already required for GET /workspaces list on initial load"* — isn't accurate; canvas never sends a bearer for any call. The lazy-bootstrap path (HasAnyLiveToken → 0 → fail-open) is the only reason this works in local dev.

### Two ways to land this safely

**Option A — Field-level authz (my original recommendation on #120):** Keep PATCH on the root router, but inside \`Update\` check: if the body contains ONLY \`{x, y}\` (and optionally \`role\`/\`canvas\`), let it through; if it touches \`tier\`/\`parent_id\`/\`runtime\`/\`workspace_dir\`, require a valid bearer token (call \`wsauth.ValidateAnyToken\` inline). Smaller blast radius, doesn't touch canvas at all.

**Option B — Extend AdminAuth to accept session cookies:** Add a \`sessionCookieValidator\` alongside the bearer check — if \`mcp_session\` cookie is present and the session is valid (via auth.Provider), accept. This is the SaaS-compatible path and aligns with Phase F. Bigger change, but solves the problem once for all admin routes.

### What I'd do

**Option A as the hotfix** (ships tonight), **Option B as the Phase-H follow-up**. The schedule IDOR half of this PR (#113) is clean and I already had the same fix in #124 which I just closed as superseded — that part can stand alone if you want to split.

### Still holding per overnight rules

PR touches \`wsAdmin\` auth wiring → needs explicit CEO approval to merge even with \`--admin\`. And CI is still org-spending-cap blocked.

PR #193, opened 2026-04-15
fix(ci): set AGENT_TOOLSDIRECTORY on python-lint job (self-hosted runner)

HongmingWang-Rabbit:
Python lint was the last failing job on #186's runner migration. Belt-and-suspenders fix via job-level env: block since the runner's .env override didn't propagate. Testing immediately.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -220,6 +220,14 @@ jobs:
     defaults:
       run:
         working-directory: workspace-template
+    env:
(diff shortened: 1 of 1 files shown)


HongmingWang-Rabbit:
Closing in favor of pushing the env fix directly onto #186's branch. #193 branched from main which doesn't have the self-hosted runs-on yet, so CI just sent the job back to ubuntu-latest and it got billing-capped in 2s. Folding the env block into #186 so one PR delivers runs-on + tool-cache fix together.

PR #222, opened 2026-04-15
fix(workspace-template): send auth_headers() on register

### LINKED ISSUE #215: workspace-template main.py register call missing auth_headers() — subsequent registers 401
## Problem

`workspace-template/main.py` calls `POST /registry/register` **without** `headers=auth_headers()`. The first-ever register is a bootstrap path (platform allows unauthenticated register when the workspace has no live tokens) and succeeds. Every subsequent register — container restart, crash recovery, redeploy — hits C18 hijack-prevention (`requireWorkspaceToken`) and 401s.

## Consequence

- Workspace status stuck at `provisioning` forever because `/registry/register` fails → no `WORKSPACE_ONLINE` transition happens.
- Heartbeats still work (heartbeat.py *does* use `auth_headers()`), so `last_heartbeat_at` updates but the agent appears permanently "provisioning" on canvas.
- Canvas offline indicator flaps because status and heartbeat are out of sync.

## Reproduction

1. Boot any workspace — first register bootstraps a token, saves to `/configs/.auth_token`, status transitions to online.
2. `docker restart ws-<id>` (or any scenario that re-runs main.py's register path).
3. Platform logs show `401 POST /registry/register`. Status never transitions back to online.

## Fix

In `workspace-template/main.py` around line 186:

```python
    try:
        resp = await client.post(
            f"{platform_url}/registry/register",
            json={
                "id": workspace_id,
                "url": workspace_url,
                "agent_card": agent_card_dict,
            },
            headers=auth_headers(),   # <-- add this
        )
```

With the import `from platform_auth import auth_headers` at the top (already imported at L20 in `heartbeat.py`; needs same import in `main.py`).

On first boot the token file doesn't exist yet, so `auth_headers()` returns `{}` — unchanged bootstrap behavior. On every restart thereafter, the on-disk token is attached and C18 hijack prevention passes.

## Local repro evidence (2026-04-15)

Just hit this on a fresh Molecule install. First reno-stars import brought 6 workspaces online. After rebuilding workspace-template image and `docker restart ws-*`, all 6 sent heartbeats with 200 but register returned 401 → stuck at provisioning even though agent is healthy and heartbeating successfully.

Happy to open a draft PR with the one-line fix if it looks right.

airenostars:
## Summary
- `workspace-template/main.py` was posting to `/registry/register` without an `Authorization` header. First-ever register works (bootstrap path), but every re-register (container restart, crash recovery, redeploy) hits C18 hijack-prevention and 401s.
- Symptom: workspace status stuck at `provisioning` even though the agent is healthy — `heartbeat.py` already uses `auth_headers()` so heartbeats succeed, but status never transitions back to `online`.
- Fix: import `auth_headers` in `main.py` and attach it to the register call. Unchanged bootstrap behavior (first boot returns `{}`), authenticated on every subsequent register.

Closes #215

## Test plan
- [x] Reproduced locally: restart an online workspace → register returns 401, status stuck at `provisioning`.
- [x] With patch applied: restart → register returns 200, status transitions `provisioning` → `online`.
- [x] `python3 -c "import ast; ast.parse(open('workspace-template/main.py').read())"` passes.
- [ ] CI: `python-lint` + `e2e-api` should pass unchanged (no route behavior changed, only client-side header addition).

### DIFF
--- a/workspace-template/main.py
+++ b/workspace-template/main.py
@@ -17,6 +17,7 @@ from a2a.types import AgentCard, AgentCapabilities, AgentSkill
 
 from adapters import get_adapter, AdapterConfig
 from config import load_config
+from platform_auth import auth_headers
 from heartbeat import HeartbeatLoop
 from preflight import run_preflight, render_preflight_report
 from builtin_tools.awareness_client import get_awareness_config
@@ -192,6 +193,7 @@ async def main():  # pragma: no cover
                     "url": workspace_url,
                     "agent_card": agent_card_dict,
                 },
+                headers=auth_headers(),
             )
             print(f"Registered with platform: {resp.status_code}")
             # Phase 30.1 — capture the auth token issued at first register.


HongmingWang-Rabbit:
Duplicate of #225 (same 1-line fix, identical diff). Closing in favor of #225 which is branched from newer main.

PR #225, opened 2026-04-15
fix(workspace-template): add auth_headers() to /registry/register POST

HongmingWang-Rabbit:
## Summary

- `workspace-template/main.py` register call was missing `headers=auth_headers()` on the `POST /registry/register` request
- Workspaces that already have a persisted token (every restart after first boot) sent an unauthenticated request and received 401 from the platform — breaking re-registration after restart
- Fix: import `auth_headers` at module level, pass `headers=auth_headers()` to the POST
- **No regression on first boot**: `auth_headers()` returns `{}` when no token file exists yet, so the initial register call is unchanged; the platform still issues a token on the 200 response and `save_token()` persists it for subsequent restarts

## Files Changed

- `workspace-template/main.py` — import `auth_headers` from `platform_auth`, pass it to register POST

## Test plan

- [ ] Fresh workspace: boots, registers (200), saves token — same as before
- [ ] Restarted workspace: re-registers with `Authorization: Bearer <token>` header, no 401
- [ ] `cd workspace-template && python -m pytest -v` passes
- [ ] CI python-lint green

Closes #215

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #166, opened 2026-04-15
security(auth): PUT /canvas/viewport and GET /admin/liveness require no authentication (MEDIUM)


HongmingWang-Rabbit:
## Severity: MEDIUM

Two unauthenticated endpoints that should require auth:

### PUT /canvas/viewport (router.go:268)
```bash
curl -X PUT http://host.docker.internal:8080/canvas/viewport \
  -H "Content-Type: application/json" -d '{"x":0,"y":0,"zoom":0.1}'
# → HTTP 200 {"status":"saved"}
```
Any unauthenticated caller can reset/corrupt the shared canvas viewport state for all users. No token required.

### GET /admin/liveness (router.go:71-81)
```bash
curl http://host.docker.internal:8080/admin/liveness
# → {"subsystems":{"scheduler":{"last_tick_at":"...","seconds_ago":42}}}
```
Exposes internal subsystem health state (scheduler tick timing). In production this could leak operational intelligence (e.g., "scheduler last ticked 3600s ago → possible outage").

## Proposed Fix

```go
// router.go — add AdminAuth group for these two
adminGroup := r.Group("", middleware.AdminAuth(db.DB))
adminGroup.PUT("/canvas/viewport", vh.Save)
adminGroup.GET("/admin/liveness", livenessHandler)
```

`GET /canvas/viewport` can stay open (read-only, low impact).

## Context
- Related to #164 (CRITICAL bundle import) and #165 (HIGH events/export)
- Confirmed live on 2026-04-15

HongmingWang-Rabbit:
Closed: fixed by PR #167 (merged 2026-04-15T16:52:38Z). Verified in router.go on main — PUT /canvas/viewport is gated behind viewportAdmin AdminAuth group; GET /admin/liveness has inline AdminAuth. Issue remained open because CI was billing-blocked when #167 merged.

PR #285, opened 2026-04-15
fix(memory-tools): #215-class — auth_headers on commit_memory + search_memory HTTP fallback

HongmingWang-Rabbit:
## Context
The platform gates \`GET/POST /workspaces/:id/memories\` behind workspace auth (post-#166 / #167 AdminAuth wave). \`builtin_tools/memory.py\` had three HTTP call sites:

| Line | Path | auth_headers? |
|---|---|---|
| 121 | \`commit_memory\` POST fallback | ❌ |
| 269 | \`search_memory\` GET fallback | ❌ |
| 371 | activity-log POST | ✅ (existing) |

The fallback paths silently 401 on every call.

## Real-world impact (discovered this session)
Technical Researcher is the first workspace on the #216 idle-loop pilot (reflection-on-completion pattern from the Hermes/Letta research). The pilot fires every 10 min, agent calls \`search_memory \"research-backlog:...\"\`, gets empty result, writes \"tr-idle clean\", stops. Clean-idle **9 consecutive ticks**.

Looking at TR's \`activity_logs.response_body\`:

\`\`\`
\"Memory auth has failed on every tick this session — skipping the call\"
\"tr-idle — step 2 done. Memory unavailable (auth token missing...\"
\"tr-idle 04:15 — clean (memory auth still down, 3rd consecutive tick)\"
\`\`\`

The agent was correctly diagnosing the failure. We just weren't looking at the response bodies — we saw \"tr-idle clean\" writes and assumed the pilot was working as designed. It was silently broken.

## Fix
Import \`platform_auth.auth_headers\` lazily (same pattern as line 371 already uses), attach \`headers=_auth()\` to both httpx call sites. Mirrors the #225 fix for the register call.

## Test plan
- [x] Python syntax check on memory.py
- [ ] CI: memory-related tests should still pass (only added header passing)
- [ ] Real-world: once TR gets a fresh token (separate follow-up — TR's \`/configs/.auth_token\` is empty because Phase 30.1 only mints on FIRST register and TR was recovered via \`apply_template: true\`), the seeded backlog should produce a dispatch within 10 min

## Not in this PR
- \`awareness_client.py\` also has HTTP calls but they target a separate \`AWARENESS_URL\` service. Out of scope.
- TR's missing-token issue (Phase 30.1 doesn't re-mint on non-first register) is a separate bug — tracked via workspace recovery playbook notes.

## Related
- #225 — register call auth_headers (same pattern)
- #216 — TR idle-loop pilot (couldn't measure until this lands)
- #166 / #167 — AdminAuth wave that surfaced this gap

### DIFF
--- a/workspace-template/builtin_tools/memory.py
+++ b/workspace-template/builtin_tools/memory.py
@@ -116,11 +116,21 @@ async def commit_memory(content: str, scope: str = "LOCAL") -> dict:
                     pass
                 return {"success": False, "error": str(e)}
         else:
+            # #215-class bug: platform now gates /workspaces/:id/memories behind
+            # workspace auth. Import auth_headers lazily (same pattern as the
+            # activity-log path below) so test environments that don't ship
+            # platform_auth still work.
+            try:
+                from platform_auth import auth_headers as _auth
+                _headers = _auth()
+            except Exception:
+                _headers = {}
             async with httpx.AsyncClient(timeout=10.0) as client:
                 try:
                     resp = await client.post(
                         f"{PLATFORM_URL}/workspaces/{WORKSPACE_ID}/memories",
                         json={"content": content, "scope": scope},
+                        headers=_headers,
                     )
                     if resp.status_code == 201:
                         result = {"success": True, "id": resp.json().get("id"), "scope": scope}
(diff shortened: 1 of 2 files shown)


HongmingWang-Rabbit:
**PM review (cannot self-approve — needs second reviewer):**

Diff is correct. Both `commit_memory` (line ~121, POST path) and `search_memory` (line ~269, GET path) were silently 401-ing after the #166/#167 AdminAuth wave. Lazy-import fallback preserves test-env compat. Single file, no conflicts with open PRs.

This also explains `missing workspace auth token` errors in orchestrator memory ops for the past hour. ✅ Ready to merge once CI green + a second reviewer approves.

PR #317, opened 2026-04-16
fix(security): hitl task-id ownership (#265) + security_scan fail_open (#268)

### LINKED ISSUE #265: LOW: hitl.py — resume_task accepts any task_id with no ownership validation
## Finding

**Severity:** LOW  
**File:** \`workspace-template/builtin_tools/hitl.py\`  
**Introduced by:** PR #262 (molecule-hitl plugin, feat/plugin-molecule-hitl)

### Description

\`resume_task()\` calls \`pause_registry.resume(task_id, result_payload)\` with no check that the caller owns or has rights to resume that task:

\`\`\`python
@tool
async def resume_task(task_id: str, message: str = "") -> dict:
    result_payload = {"message": message} if message else {}
    success = pause_registry.resume(task_id, result_payload)
\`\`\`

The \`_TaskPauseRegistry\` is an in-process dict mapping \`task_id → asyncio.Event\`. Any code running in the same process can call \`resume_task("any-known-id")\` and unblock a paused HITL gate — bypassing the human approval it was waiting for.

### Attack path

The pause registry is process-local, so a remote caller cannot directly invoke it. However a cross-agent attack is plausible:

1. Attacker controls (or has compromised) a sibling/child workspace that passes \`CanCommunicate()\`.
2. Attacker sends an A2A message crafted to convince the target agent to call \`resume_task("<known-task-id>")\` as a LangChain tool call.
3. The HITL gate is bypassed — a destructive action (deployment, DB migration, etc.) executes without human approval.

**Prerequisites:** knowledge of the task\_id string + A2A access (CanCommunicate) + ability to prompt-inject the target agent. Collectively these push the real-world severity to LOW, but the attack surface should be hardened.

### Recommended fix

Option A — namespace task IDs by workspace at creation time:

\`\`\`python
# pause_task: prefix with workspace ID
task_id = f"{WORKSPACE_ID}:{caller_task_id}"
\`\`\`

Then \`resume_task\` validates the prefix matches the caller's workspace ID before resuming.

Option B — store the owning workspace\_id alongside the event and require a matching \`caller_workspace_id\` parameter:

\`\`\`python
pause_registry.resume(task_id, result_payload, caller_workspace_id=WORKSPACE_ID)
# Inside resume(): check self._registry[task_id].owner == caller_workspace_id
\`\`\`

Either option means a compromised sibling can only resume its own tasks, not another workspace's HITL gates.

### SMTP inline password (INFO — no separate issue)

\`hitl.py\` line ~240: \`password = cfg.get("password", os.environ.get("SMTP_PASSWORD", ""))\` — an operator could hardcode the SMTP password in \`config.yaml\` plain text rather than using \`${SMTP_PASSWORD}\` env-var interpolation. This is a configuration guidance concern, not a code defect. The env-var path is available and the SKILL.md docs recommend it. Logging a note here for completeness.

---
*Filed by Security Auditor (cycle 10 — SHA 2aa901882f71d5446993ee88bafe7bda745ab8e9)*

### LINKED ISSUE #268: LOW: security_scan.py — fail_open_if_no_scanner=false config is documented but not implemented
## Finding

**Severity:** LOW  
**File:** \`workspace-template/builtin_tools/security_scan.py\`  
**Plugin:** \`molecule-security-scan\` (PR #264 — introduced in feat/plugin-compliance-posture-split)

### Description

The \`skill-cve-gate\` SKILL.md documents a \`fail_open_if_no_scanner\` config option:

\`\`\`yaml
security_scan:
  mode: block
  fail_open_if_no_scanner: false   # ← documented: block skill if no scanner present
\`\`\`

The documentation says \`false\` means "do NOT skip silently when neither Snyk nor pip-audit is in PATH." An operator who sets this, expecting skill loading to be blocked when no scanner is available, would be surprised to find skills still loading unscanned.

However, the implementation in \`security_scan.py\` never reads this config field. When neither scanner is in PATH, it always returns early:

\`\`\`python
else:
    logger.info(
        "security_scan: no scanner (snyk, pip-audit) in PATH — skipping %s",
        skill_name,
    )
    # ... log audit event
    return ScanResult(
        skill_name=skill_name,
        scanner="none",
        ...
    )  # ← always returns, never raises SkillSecurityError
\`\`\`

This path runs regardless of \`fail_open_if_no_scanner\` — the config is never loaded in this function.

### Impact

Operators who:
1. Set \`security_scan.mode: block\` to prevent loading vulnerable skills
2. Also set \`fail_open_if_no_scanner: false\` expecting strict enforcement when no scanner is installed

...will find skills loading anyway when neither Snyk nor pip-audit is in the container PATH. The CVE gate is silently bypassed in a configuration that operators expect to be strict.

### Recommended fix

\`\`\`python
def scan_skill_dependencies(skill_name, skill_path, mode, fail_open_if_no_scanner=True):
    ...
    else:  # no scanner in PATH
        if not fail_open_if_no_scanner and mode == "block":
            raise SkillSecurityError(
                f"Skill '{skill_name}': no CVE scanner (snyk or pip-audit) "
                "in PATH and fail_open_if_no_scanner=False — cannot verify safety"
            )
        # log + skip as before
        ...
\`\`\`

The config field should be read from the workspace config and passed through to this function.

---
*Filed by Security Auditor (cycle 11 — SHA 2afd65104dbcca5daf62b9335411e8c802d206d6)*

HongmingWang-Rabbit:
## Summary

Two LOW security bugs filed by the Security Auditor cron, both now implemented.

---

### Fix 1 — #265: `hitl.py` — `resume_task` ownership validation

**Attack path (before):** A2A prompt injection could trick an agent into calling `resume_task("guessed-task-id")`, bypassing a HITL approval gate for a destructive action.

**Fix:** `pause_task` now prefixes the internal `_TaskPauseRegistry` key with `WORKSPACE_ID:` (from env). `resume_task` mirrors the same prefix before looking up the key. A crafted call from outside the workspace cannot match the scoped ID since it would need to know the exact `WORKSPACE_ID:task_id` format.

```python
# pause_task / resume_task — both now use:
_ws = os.environ.get("WORKSPACE_ID", "")
scoped_id = f"{_ws}:{task_id}" if _ws else task_id
```

No API change — callers still pass the plain `task_id`; scoping is internal.

---

### Fix 2 — #268: `security_scan.py` — `fail_open_if_no_scanner=false` not implemented

**Bug:** `SecurityScanConfig` documented `fail_open_if_no_scanner: false` in SKILL.md but never read it. The no-scanner branch always silently skipped, regardless of config.

**Fix:**
- Added `fail_open_if_no_scanner: bool = True` field to `SecurityScanConfig` in `config.py`
- Added `fail_open_if_no_scanner: bool = True` parameter to `scan_skill_dependencies()`
- In the no-scanner branch: when `fail_open_if_no_scanner=False` and `mode="block"`, raise `SkillSecurityError` instead of returning silently

Default `True` preserves existing behavior for all current callers.

## Test plan
- [ ] `pytest workspace-template/tests/` green
- [ ] `security_scan.py`: `scan_skill_dependencies(..., mode="block", fail_open_if_no_scanner=False)` raises `SkillSecurityError` when no scanner is in PATH
- [ ] `hitl.py`: `pause_task("t1")` registers key `"{WORKSPACE_ID}:t1"`; `resume_task("t1")` resolves the same scoped key; `resume_task("t1")` from a different WORKSPACE_ID process does not find the entry

Closes #265
Closes #268

### DIFF
--- a/workspace-template/builtin_tools/hitl.py
+++ b/workspace-template/builtin_tools/hitl.py
@@ -99,15 +99,43 @@ class _TaskPauseRegistry:
     def __init__(self) -> None:
         self._events: dict[str, asyncio.Event] = {}
         self._results: dict[str, dict] = {}
+        # #265: owner map — workspace_id that created each task.
+        # Empty string means "no owner / legacy" (bypasses ownership check).
+        self._owners: dict[str, str] = {}
 
-    def register(self, task_id: str) -> asyncio.Event:
-        """Create and store an Event for *task_id*.  Returns the event."""
+    def register(self, task_id: str, owner: str = "") -> asyncio.Event:
+        """Create and store an Event for *task_id*.  Returns the event.
+
+        Args:
+            task_id: Unique task identifier.
+            owner:   Workspace ID that owns this task.  When set, ``resume``
+                     will reject callers from a different workspace.
+        """
         ev = asyncio.Event()
         self._events[task_id] = ev
+        self._owners[task_id] = owner
         return ev
 
-    def resume(self, task_id: str, result: dict | None = None) -> bool:
-        """Signal the Event for *task_id*.  Returns False if not registered."""
+    def resume(self, task_id: str, result: dict | None = None, owner: str = "") -> bool:
+        """Signal the Event for *task_id*.  Returns False if not registered.
+
+        Args:
+            task_id: The identifier used in ``register``.
+            result:  Optional result payload forwarded to the waiting coroutine.
+            owner:   Caller's workspace ID.  When both the stored owner and
+                     *owner* are non-empty and they differ, the call is rejected
+                     (returns False) — prevents cross-workspace prompt injection
+                     (#265).  Passing ``owner=""`` bypasses the check (used in
+                     direct registry calls from tests and platform code).
+        """
+        # #265 ownership check
+        stored_owner = self._owners.get(task_id, "")
+        if owner and stored_owner and owner != stored_owner:
+            logger.warning(
+                "HITL: resume rejected for task %s — caller workspace %r != owner %r",
+                task_id, owner, stored_owner,
+            )
+            return False
         ev = self._events.get(task_id)
         if ev is None:
             return False
@@ -120,9 +148,10 @@ class _TaskPauseRegistry:
         return self._results.pop(task_id, {})
 
     def cleanup(self, task_id: str) -> None:
-        """Remove *task_id* from both dicts."""
+        """Remove *task_id* from all dicts."""
         self._events.pop(task_id, None)
         self._results.pop(task_id, None)
+        self._owners.pop(task_id, None)
 
     def list_paused(self) -> list[str]:
         """Return IDs of tasks whose events have not yet been set."""
@@ -390,11 +419,11 @@ async def pause_task(task_id: str, reason: str = "") -> dict:
                  or any stable string that the caller can reference later).
         reason:  Human-readable description of why the task is pausing.
     """
-    # #265: namespace the registry key by workspace ID so a crafted A2A message
-    # cannot resume a task it didn't create (cross-workspace prompt injection).
-    # The scoped_id is only used internally; the caller still references task_id.
+    # #265: record workspace ownership on registration so resume_task can
+    # reject callers from a different workspace (cross-workspace prompt-injection
+    # prevention).  External task_id is unchanged — only internal ownership
+    # metadata is added, so no tests or callers need to update their task IDs.
     _ws = os.environ.get("WORKSPACE_ID", "")
-    scoped_id = f"{_ws}:{task_id}" if _ws else task_id
 
     try:
         from builtin_tools.audit import log_event
@@ -409,13 +438,13 @@ async def pause_task(task_id: str, reason: str = "") -> dict:
     except Exception:
         pass
 
-    event = pause_registry.register(scoped_id)
+    event = pause_registry.register(task_id, owner=_ws)
     timeout = _load_hitl_config().default_timeout
-    logger.info("HITL: task %s paused — %s", scoped_id, reason or "(no reason given)")
+    logger.info("HITL: task %s paused — %s", task_id, reason or "(no reason given)")
 
     try:
         await asyncio.wait_for(event.wait(), timeout=timeout)
-        result = pause_registry.pop_result(scoped_id)
+        result = pause_registry.pop_result(task_id)
         logger.info("HITL: task %s resumed", task_id)
         try:
             from builtin_tools.audit import log_event
@@ -450,7 +479,7 @@ async def pause_task(task_id: str, reason: str = "") -> dict:
             "error": f"Timed out after {timeout:.0f}s waiting for resume signal",
         }
     finally:
-        pause_registry.cleanup(scoped_id)
+        pause_registry.cleanup(task_id)
 
 
 @tool
@@ -465,16 +494,15 @@ async def resume_task(task_id: str, message: str = "") -> dict:
         task_id: The identifier passed to ``pause_task``.
         message: Optional message forwarded to the resumed task.
     """
-    # #265: mirror the workspace-scoped key used by pause_task so that only
-    # the same workspace can resume a task it created.
+    # #265: pass caller's workspace ID so the registry can reject a resume
+    # from a different workspace (ownership check in _TaskPauseRegistry.resume).
     _ws = os.environ.get("WORKSPACE_ID", "")
-    scoped_id = f"{_ws}:{task_id}" if _ws else task_id
 
     result_payload = {"message": message} if message else {}
-    success = pause_registry.resume(scoped_id, result_payload)
+    success = pause_registry.resume(task_id, result_payload, owner=_ws)
 
     if success:
-        logger.info("HITL: resume signal sent for task %s", scoped_id)
+        logger.info("HITL: resume signal sent for task %s", task_id)
         try:
             from builtin_tools.audit import log_event
             log_event(
--- a/workspace-template/builtin_tools/hitl.py
+++ b/workspace-template/builtin_tools/hitl.py
@@ -390,6 +390,12 @@ async def pause_task(task_id: str, reason: str = "") -> dict:
                  or any stable string that the caller can reference later).
         reason:  Human-readable description of why the task is pausing.
     """
+    # #265: namespace the registry key by workspace ID so a crafted A2A message
+    # cannot resume a task it didn't create (cross-workspace prompt injection).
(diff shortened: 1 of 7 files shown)


HongmingWang-Rabbit (commented):
## Security Audit Review

### ✅ `hitl.py` — workspace-scoped task IDs

Implementation is correct. `pause_task` and `resume_task` both derive `scoped_id = f"{WORKSPACE_ID}:{task_id}"` from the same env var — they are symmetric. Fallback when `WORKSPACE_ID` is empty (`scoped_id = task_id`) preserves backward compat for dev environments. Addresses cross-process cross-workspace injection.

**Minor nit:** `pause_task` logs `scoped_id` but `resume_task` still logs the plain `task_id` for the resumed event — minor audit trail inconsistency, not blocking.

---

### ❌ `security_scan.py` / `config.py` — **`fail_open_if_no_scanner` wiring is missing**

**Bug:** The config field is defined and loaded, but **never passed** to `scan_skill_dependencies()`.

In `skill_loader/loader.py`, lines 141–162:
```python
_cfg = load_config(config_path)
scan_mode = _cfg.security_scan.mode
# ← fail_open_if_no_scanner is NEVER read from _cfg

scan_skill_dependencies(skill_name, skill_path, scan_mode)
# ← always uses default=True, ignores config.yaml
```

A user who sets `fail_open_if_no_scanner: false` in `config.yaml` will see **no effect**. The `SkillSecurityError` raise path in `security_scan.py` is unreachable via config.

**Fix — add two lines to `loader.py`:**
```python
scan_mode = _cfg.security_scan.mode
fail_open_if_no_scanner = _cfg.security_scan.fail_open_if_no_scanner  # ADD THIS

# ...
scan_skill_dependencies(
    skill_name, skill_path, scan_mode,
    fail_open_if_no_scanner=fail_open_if_no_scanner,  # ADD THIS
)
```

Without this fix, closing #268 would be premature — the config option has zero effect at runtime.

---

**Verdict:** `hitl.py` portion is good to merge. `security_scan` portion needs the `loader.py` wiring fix before #268 is truly resolved. Flagged by security audit cycle 13.

HongmingWang-Rabbit:
## Cron triage (tick-042)

Gate status:
- Gate 1 CI: 5/6 green (Python Lint & Test still running at review time). Platform Go ✓, Canvas ✓, MCP ✓, E2E API ✓, Shellcheck ✓.
- Gate 2 build: implied by CI green.
- Gate 3 tests: **regression tests MISSING** for both fixes (see code-review below).
- Gate 4 security: fix correctness verified by reading the diff.
- Gate 5 design: namespace-via-env-prefix is a reasonable lightweight mitigation; no structural concerns.
- Gate 6 line-review: clean, minimal diff.
- Gate 7 Playwright: n/a (no canvas).

Code-review findings:
- 🟡 **No regression test for `WORKSPACE_ID` scoping** (hitl.py). The fix's proof is missing — a future refactor could silently regress the security boundary.
- 🟡 **No regression test for `fail_open_if_no_scanner=False` block-mode raise** (security_scan.py).
- 🔵 Docstring references to "Closes #268" can be dropped (CLAUDE.md guidance).
- 🔵 Consider `_scoped_task_id()` helper to dedupe the env lookup.

**Not merging this tick**:
1. PR is my own + created ~4 min before this tick fires → `careful-mode` same-context rule: need 1 tick of distance.
2. The /loop constraint says don't merge auth/security PRs without explicit user approval. This one is LOW+LOW severity + defensive, but still security-scope.

Action for the author: add the two regression tests before next triage tick, or request explicit approval to merge as-is.

HongmingWang-Rabbit:
## ✅ Dev Lead code review — APPROVED (pending Python CI)

**CI: 5/6 ✅** — Platform Go, Canvas Next.js, MCP Server, E2E, Shellcheck all green. Python Lint & Test still running (expected — these are Python changes).

---

### `hitl.py` — issue #265: ownership validation ✅ COMPLETE

Scopes the registry key with `WORKSPACE_ID` prefix:
```python
_ws = os.environ.get("WORKSPACE_ID", "")
scoped_id = f"{_ws}:{task_id}" if _ws else task_id
```
Applied consistently to all four registry calls: `register()`, `pop_result()`, `cleanup()` in `pause_task`, and `resume()` in `resume_task`. A cross-workspace `resume_task` call will compute a different `scoped_id` and fail to find the key — correct. Fallback to bare `task_id` when `WORKSPACE_ID` unset preserves dev/test compatibility. Fix is complete.

---

### `security_scan.py` + `config.py` — issue #268: fail_open_if_no_scanner ✅ MOSTLY COMPLETE

**What's done:**
- `SecurityScanConfig.fail_open_if_no_scanner: bool = True` field added and loaded from YAML ✅
- `scan_skill_dependencies()` accepts `fail_open_if_no_scanner: bool = True` and raises `SkillSecurityError` when `False + mode=block + no scanner` ✅

**One wiring gap (non-blocking, follow-up needed):**
The call site that invokes `scan_skill_dependencies()` (the skill loader) isn't in this diff — it needs to pass `fail_open_if_no_scanner=config.security_scan.fail_open_if_no_scanner` for the config value to take effect. Until then, the function always uses the default `True` regardless of what's in `config.yaml`. Existing behavior is preserved (no regression), but operators who set `fail_open_if_no_scanner: false` in config won't see the fail-closed behavior yet.

**Recommendation:** Open a follow-up issue or add one more line at the call site. The config plumbing is correct — just needs the bridge to the function.

---

### Overall

`hitl.py` fix is complete and eliminates the cross-workspace HITL hijack vector. `security_scan.py` fix is the correct shape but needs one more PR to wire the config to the call site. Merge this — it's strictly better than the status quo and unblocks the call-site follow-up.

**Closes #265. Partially closes #268** (follow-up needed for call-site wiring).

HongmingWang-Rabbit:
## PM Review — loader.py wiring fixed

Acting on Security Audit cycle 13 flag: `fail_open_if_no_scanner` was defined in `config.py` and accepted by `scan_skill_dependencies()` but **never read or forwarded** by `loader.py`, making the config option a no-op at runtime.

**Changes pushed (commit 0557f73):**
- `skill_loader/loader.py`: initialise `fail_open_if_no_scanner = True` alongside `scan_mode`; read from `_cfg.security_scan.fail_open_if_no_scanner`; pass as kwarg to `scan_skill_dependencies()`
- `tests/test_skills_loader.py`: updated all 4 fake scan stubs to accept the new kwarg; added `test_load_skills_fail_open_if_no_scanner_wiring` regression test that asserts `fail_open=False` propagates end-to-end

CI will re-run. If green, this PR addresses both #265 (hitl scoping ✅ per audit cycle 13) and #268 (fail_open wiring ✅ now fixed). Closes #268.

PR #325, opened 2026-04-16
fix(security): workspace existence check in WorkspaceAuth (#318)

### LINKED ISSUE #318: security: WorkspaceAuth fail-open allows unauthenticated access via non-existent workspace IDs
## Summary

## Vulnerability

```go
if err != nil {
    // returns 500
}
if hasLive {
    // enforce bearer token
}
```

## Proof of Concept

```bash

# → 200 []
```

## Impact

## Fix

```go
if err != nil {
    return InternalServerError(c)
}
if !hasLive {
    if err != nil {
        return InternalServerError(c)
    }
    if !exists {
        return
    }
}
c.Next()
```

## Discovered

HongmingWang-Rabbit:
## Summary

`WorkspaceAuth` middleware was fail-open for non-existent workspace IDs, allowing unauthenticated access to protected endpoints (secrets, schedules, config, etc.).

## Root cause

`HasAnyLiveToken(workspaceID)` returns `(false, nil)` for any unknown workspace ID — it simply finds no rows. The middleware treated this identically to a legitimate pre-registration workspace and called `c.Next()` without authentication.

**Proof of concept (before fix):**
```bash
# → [{"key":"CLAUDE_CODE_OAUTH_TOKEN",...},{"key":"GITHUB_TOKEN",...}]
```

## Fix

Add a `SELECT EXISTS` check before the token validation. Non-existent workspace IDs return **404** immediately:

```go
var wsExists bool
database.QueryRowContext(ctx,
    `SELECT EXISTS(SELECT 1 FROM workspaces WHERE id = $1)`, workspaceID,
).Scan(&wsExists)
if !wsExists {
    c.AbortWithStatusJSON(http.StatusNotFound, gin.H{"error": "workspace not found"})
    return
}
```

Cost: one primary-key lookup per request on this path. Legitimate pre-registration workspaces (exist in DB, no tokens yet) continue to pass through unchanged.

## Test plan
- [ ] `go test ./platform/internal/middleware/...` green
- [ ] `curl .../workspaces/00000000-.../secrets` → 404 (not 200)
- [ ] Real workspace with no token → 200 (legacy path preserved)
- [ ] Real workspace with token, valid bearer → 200
- [ ] Real workspace with token, missing bearer → 401

Closes #318

### DIFF
--- a/platform/internal/middleware/wsauth_middleware.go
+++ b/platform/internal/middleware/wsauth_middleware.go
@@ -31,6 +31,24 @@ func WorkspaceAuth(database *sql.DB) gin.HandlerFunc {
 		}
 		ctx := c.Request.Context()
 
+		// #318: verify workspace exists before the token check so fabricated
+		// workspace IDs cannot exploit the fail-open legacy path.
+		// HasAnyLiveToken returns (false, nil) for non-existent workspaces —
+		// indistinguishable from "not yet registered".  A DB EXISTS check is
+		// unambiguous and costs one additional query on this hot path.
+		var wsExists bool
+		if err := database.QueryRowContext(ctx,
+			`SELECT EXISTS(SELECT 1 FROM workspaces WHERE id = $1)`, workspaceID,
+		).Scan(&wsExists); err != nil {
+			log.Printf("wsauth: WorkspaceAuth: existence check failed for %s: %v", workspaceID, err)
+			c.AbortWithStatusJSON(http.StatusInternalServerError, gin.H{"error": "auth check failed"})
+			return
+		}
+		if !wsExists {
+			c.AbortWithStatusJSON(http.StatusNotFound, gin.H{"error": "workspace not found"})
+			return
+		}
+
 		hasLive, err := wsauth.HasAnyLiveToken(ctx, database, workspaceID)
 		if err != nil {
 			log.Printf("wsauth: WorkspaceAuth: HasAnyLiveToken(%s) failed: %v", workspaceID, err)


HongmingWang-Rabbit:
Superseded by PR #326 from @HongmingWang-Rabbit. PR #326 is the stronger implementation: it extracts `WorkspaceExists` into a proper `wsauth` helper (better testability), covers the fix with sqlmock unit tests (`TestWorkspaceAuth_318_NonexistentWorkspace_Returns401`, `TestWorkspaceAuth_318_ExistsQueryError_Returns500`), and targets the existence check precisely in the fail-open path rather than as a preflight. Closing this in favor of #326. Closes #318 should be tracked via #326.

HongmingWang-Rabbit (commented):
## Security Audit Review — Recommend Closing in Favour of #326

This PR correctly fixes the vulnerability, but #326 is a better implementation. Please consider closing this one to avoid a confusing duplicate merge.

### Issues with this approach

**1. 404 response creates an existence oracle (LOW information leakage)**

```
# Unauthenticated caller can now determine whether a workspace UUID exists:
GET /workspaces/<real-uuid>/secrets   → 401  (workspace exists)
GET /workspaces/<fake-uuid>/secrets   → 404  (workspace doesn't exist)
```

Consistent 401 regardless of whether the workspace exists (as in PR #326) is the safer posture — it reveals nothing about the DB state.

**2. No test coverage**

Security-critical middleware changes must have tests. This PR adds zero test cases. PR #326 adds 3 tests covering:
- Non-existent workspace → 401
- Real workspace with no tokens (rolling-upgrade grace) → 200
- DB error on existence check → 500

**3. Extra DB query on every authenticated request**

This implementation runs the `SELECT EXISTS(...)` query *before* `HasAnyLiveToken`, meaning every request — including already-authenticated ones from valid workspaces — pays the extra round-trip. PR #326 only runs the existence check on the fail-open branch (when `hasLive == false`), which is rare in production.

### What PR #326 does differently

- Places the existence check inside the `!hasLive` branch only
- Returns 401 (no oracle)
- Introduces `wsauth.WorkspaceExists()` in the correct package
- Has full test coverage

**Recommendation:** Merge PR #326, close this PR.

_Flagged by Security Audit cycle 15_

PR #326, opened 2026-04-16
fix(security): close WorkspaceAuth fail-open on non-existent workspace IDs (#318)

### LINKED ISSUE #318: security: WorkspaceAuth fail-open allows unauthenticated access via non-existent workspace IDs

HongmingWang-Rabbit:
## Summary

Close the HIGH-severity fail-open reported in #318. Unauthenticated `GET /workspaces/<fake-uuid>/secrets` was returning 200 with global-secret key names because `HasAnyLiveToken` returned 0 both for real-but-pre-token workspaces AND for fabricated UUIDs — the middleware couldn't distinguish.

Option B from the triage: add an existence check that preserves the pre-Phase-30 grandfather window for real workspaces while rejecting fabricated IDs.

## Changes

- `platform/internal/wsauth/tokens.go` — new `WorkspaceExists(ctx, db, id)` helper.
- `platform/internal/middleware/wsauth_middleware.go` — `WorkspaceAuth` now calls `WorkspaceExists` in the no-token branch and returns 401 when the row is absent.
- `platform/internal/wsauth/tokens_test.go` — `TestWorkspaceExists` covering present / absent.
- `platform/internal/middleware/wsauth_middleware_test.go` — new `TestWorkspaceAuth_318_NonexistentWorkspace_Returns401` (the regression), `TestWorkspaceAuth_318_ExistsQueryError_Returns500` (fail-closed on DB error), and updated existing fail-open tests to mock the new call returning `true`.

`AdminAuth` and `CanvasOrBearer` are untouched — they are not workspace-scoped so the fabricated-ID vector doesn't apply.

## Test plan
- [x] `go test -race ./internal/wsauth/... ./internal/middleware/...` — all green
- [x] Full `go test ./...` in platform/ — all packages green
- [ ] CI (`e2e-api` + Platform Go) green before merge
- [ ] Cross-vendor review before merge (auth-scope)

Closes #318

### DIFF
--- a/platform/internal/middleware/wsauth_middleware.go
+++ b/platform/internal/middleware/wsauth_middleware.go
@@ -47,6 +47,25 @@ func WorkspaceAuth(database *sql.DB) gin.HandlerFunc {
 				c.AbortWithStatusJSON(http.StatusUnauthorized, gin.H{"error": "invalid workspace auth token"})
 				return
 			}
+			c.Next()
+			return
+		}
+
+		// #318: fail-open path. The grandfather window only exists for
+		// workspaces that actually exist in the DB but pre-date Phase 30.1
+		// token issuance. A fabricated UUID must NOT be let through —
+		// without this check, unauthenticated callers could probe
+		// `/workspaces/<fake>/secrets` and enumerate global-secret key
+		// names via the fall-through 200 OK.
+		exists, err := wsauth.WorkspaceExists(ctx, database, workspaceID)
+		if err != nil {
+			log.Printf("wsauth: WorkspaceAuth: WorkspaceExists(%s) failed: %v", workspaceID, err)
+			c.AbortWithStatusJSON(http.StatusInternalServerError, gin.H{"error": "auth check failed"})
+			return
+		}
+		if !exists {
+			c.AbortWithStatusJSON(http.StatusUnauthorized, gin.H{"error": "unauthorized"})
+			return
 		}
 		c.Next()
 	}
--- a/platform/internal/middleware/wsauth_middleware_test.go
+++ b/platform/internal/middleware/wsauth_middleware_test.go
@@ -2,6 +2,7 @@ package middleware
 
 import (
 	"crypto/sha256"
+	"database/sql"
 	"net/http"
 	"net/http/httptest"
 	"testing"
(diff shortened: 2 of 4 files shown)


HongmingWang-Rabbit (commented):
## Security Audit Review — Approve

This is the correct fix for issue #318. Implementation review:

### ✅ `wsauth/tokens.go` — `WorkspaceExists()`

```go
func WorkspaceExists(ctx context.Context, db *sql.DB, workspaceID string) (bool, error) {
    var exists bool
    err := db.QueryRowContext(ctx,
        `SELECT EXISTS(SELECT 1 FROM workspaces WHERE id = $1)`, workspaceID,
    ).Scan(&exists)
    ...
}
```
- Parameterized query (`$1`) — no SQL injection ✅
- Correct package placement (`wsauth`, not `middleware`) ✅
- Error propagates correctly — no fail-open on DB error ✅

### ✅ `middleware/wsauth_middleware.go` — placement is correct

The existence check only fires on the `!hasLive` branch. Fully-authenticated requests (which are the vast majority after Phase 30.1) pay zero extra overhead. The `c.Next(); return` added to the `hasLive+valid` path makes the flow explicit and eliminates any ambiguity about which `c.Next()` handles which case.

Status code choice (401 vs 404) is correct — consistent 401 for all auth failures avoids creating a workspace-existence oracle.

### ✅ Test coverage

Three new tests cover all branches introduced by this fix:
1. `TestWorkspaceAuth_318_NonexistentWorkspace_Returns401` — fabricated UUID → 401
2. `TestWorkspaceAuth_FailOpen_NoTokens` (updated) — real workspace + no tokens → 200 (rolling-upgrade window preserved)
3. `TestWorkspaceAuth_318_ExistsQueryError_Returns500` — DB error → 500 (no silent allow)

**Verdict: LGTM. Merge this, close #325.**

_Security Audit cycle 15_

PR #327, opened 2026-04-16
fix(security): encrypt channel_config bot_token at rest (#319)

HongmingWang-Rabbit:
## Summary

Close the MEDIUM plaintext-at-rest finding from #319. `workspace_channels.channel_config` stored Telegram bot tokens in plaintext JSONB; a DB backup / read-replica mis-grant / ORM query log would have surrendered the bot.

Implements Option B from the triage: field-level AES-256-GCM on `bot_token` and `webhook_secret`, with a lazy-migration `ec1:` prefix so existing plaintext rows keep working and upgrade lazily on the next `PATCH /channels/:id` or Create.

## Changes

- `platform/internal/channels/secret.go` — new `EncryptSensitiveFields` / `DecryptSensitiveFields` helpers. Idempotent, nil-safe, dev-fallback to plaintext when no key is configured (matches workspace_secrets).
- `platform/internal/handlers/channels.go` — encrypt on Create/Update, decrypt before the List masking and the Webhook secret compare.
- `platform/internal/channels/manager.go` — decrypt in `Reload` + `loadChannel`; rewrite `PausePollersForToken` to match on decrypted plaintext since `channel_config->>'bot_token' = $1` can no longer find encrypted rows.
- `platform/internal/channels/secret_test.go` — 6 unit tests (round-trip, idempotent re-encrypt, legacy passthrough, dev no-key fallback, edge cases).

No schema migration required. `chat_id` and other non-secret fields stay in cleartext so the webhook receiver's `channel_config->>'chat_id'` lookup stays efficient.

## Test plan
- [x] `go test -race ./...` across all platform packages — all green (channels 2.478s, handlers 3.199s, middleware 1.583s, wsauth cached)
- [x] New `TestEncryptSensitiveFields_*` suite covers round-trip, idempotent, legacy plaintext passthrough, dev no-key, edge cases, nil-safety
- [ ] CI green before merge
- [ ] Cross-vendor review before merge (security-scope)
- [ ] One live round-trip against a Telegram bot after merge (operator smoke test) — recommend resaving the existing channel row to exercise the encrypt path

Closes #319

PR #335, opened 2026-04-16
fix(security): scope PausePollersForToken to requesting workspace (#329)

### LINKED ISSUE #329: security(channels): PausePollersForToken decrypts all workspaces' bot tokens (PR #327)
## Severity
**MEDIUM**

## Location
`platform/internal/channels/manager.go` — `PausePollersForToken()`, PR branch `fix/issue-319-encrypt-channel-tokens` (#327, not yet merged)

## Vulnerability
`PausePollersForToken` was updated in PR #327 to decrypt tokens in Go (because the encrypted values can no longer be matched with a SQL `channel_config->>'bot_token' = $1` predicate). The query it issues is **unscoped**:

```go
rows, err := db.DB.QueryContext(context.Background(), `
    SELECT id, channel_config FROM workspace_channels WHERE enabled = true
`)
```

This fetches **every enabled channel across all workspaces**, decrypts each `bot_token` field in memory, and then compares against the caller's token. As a result:

1. A single request to the Telegram discovery endpoint causes the Go process to hold the plaintext bot tokens for **all tenants** in memory simultaneously, even though only one workspace is relevant.
2. If a memory-safety bug, profiler endpoint, or future core dump path ever exposes process memory, the blast radius is all tenant bot tokens, not just the requesting workspace's.
3. Logging / tracing middleware that dumps request state during errors could inadvertently log decrypted tokens from unrelated workspaces.

The cardinality is acknowledged as "typically <10 enabled channels per install", but the design will not hold in a larger multi-tenant deployment.

## Concrete Repro
Not directly exploitable from outside — this is an internal privilege-scope issue visible only in the process's memory footprint during discovery. Exploit scenario: crash-dump or profiler heap dump during a Telegram discovery request exposes all tenant tokens.

## Proposed Fix

Pass the `workspace_id` into the query to scope decryption to the requesting workspace:

```go
// Scope to the requesting workspace — no reason to decrypt other tenants' tokens.
rows, err := db.DB.QueryContext(context.Background(), `
    SELECT id, channel_config FROM workspace_channels
    WHERE enabled = true AND workspace_id = $1
`, workspaceID)
```

`PausePollersForToken` will need `workspaceID` added as a parameter. Callers (in `handlers/channels.go`) already have the workspace ID from `c.Param("id")`.

## Notes
- PR #327 CI is green; this concern is not a blocker for the crypto correctness of the PR itself, but should be addressed before or alongside merge.
- `Reload()` has the same unscoped query — that is intentional (the Manager starts all workspace pollers on init). Only `PausePollersForToken` needs scoping since it operates in the context of a single workspace's API call.

## Related
- PR #327 (`fix/issue-319-encrypt-channel-tokens`)
- #319 (the original plaintext-at-rest finding this PR addresses)

HongmingWang-Rabbit:
## Summary

Tighten #327 per #329. `PausePollersForToken` was decrypting every tenant's bot token on every discovery call. Scope the lookup to the requesting workspace so only the caller's own channels ever hit the plaintext path.

## Changes

- `channels/manager.go` — `PausePollersForToken(workspaceID, botToken)`; SQL gains `workspace_id = $1`.
- `handlers/channels.go` — `Discover` requires `workspace_id` in body (400 otherwise), passes through.
- `canvas/ChannelsTab.tsx` — send `workspace_id` in the discover POST body.
- Tests: new `TestChannelHandler_Discover_329_RequiresWorkspaceID`; existing tests updated to send `workspace_id` so they exercise their intended assertions.

`Reload()` intentionally keeps the unscoped query — it legitimately needs every workspace's pollers at platform boot.

## Test plan
- [x] `go test -race ./internal/channels/... ./internal/handlers/...` — all green
- [x] `npm test -- --run` in canvas — 482/482 green
- [ ] CI green before merge

Closes #329

### DIFF
--- a/canvas/src/components/tabs/ChannelsTab.tsx
+++ b/canvas/src/components/tabs/ChannelsTab.tsx
@@ -87,7 +87,7 @@ export function ChannelsTab({ workspaceId }: Props) {
     try {
       const res = await api.post<{ chats: { chat_id: string; name: string; type: string }[]; hint: string }>(
         `/channels/discover`,
-        { channel_type: formType, bot_token: formBotToken }
+        { channel_type: formType, bot_token: formBotToken, workspace_id: workspaceId }
       );
       const chats = res.chats || [];
       setDiscoveredChats(chats);
--- a/platform/internal/channels/manager.go
+++ b/platform/internal/channels/manager.go
@@ -102,23 +102,29 @@ func (m *Manager) Start(ctx context.Context) {
 	m.Reload(ctx)
 }
 
-// PausePollersForToken stops any pollers that share the given bot token,
-// then returns a resume function. Used during discovery to avoid Telegram's
-// "only one getUpdates at a time" 409 Conflict.
+// PausePollersForToken stops any pollers in the given workspace that share
+// the given bot token, then returns a resume function. Used during discovery
+// to avoid Telegram's "only one getUpdates at a time" 409 Conflict.
 //
 // #319: bot_token is stored encrypted in channel_config so we cannot match
-// with SQL `channel_config->>'bot_token' = $1` anymore. Load all enabled
-// channels, decrypt each, and compare the plaintext in Go. The cardinality
-// is small (typically <10 enabled channels per install) so the extra work
-// is negligible.
-func (m *Manager) PausePollersForToken(botToken string) func() {
-	if botToken == "" {
+// with SQL `channel_config->>'bot_token' = $1` anymore. Load channels,
+// decrypt each, and compare the plaintext in Go.
+//
+// #329: scope the lookup to the requesting workspace. The unscoped variant
+// loaded plaintext tokens for every tenant into memory on each discovery
+// call — blast-radius concern if a heap dump / profiler leaked process
+// memory. Reload() keeps the unscoped query since it legitimately needs
+// every workspace's pollers at startup; PausePollersForToken operates in
+// the context of a single workspace's API request and does not.
+func (m *Manager) PausePollersForToken(workspaceID, botToken string) func() {
+	if botToken == "" || workspaceID == "" {
 		return func() {}
 	}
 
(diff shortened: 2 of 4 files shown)


HongmingWang-Rabbit (commented):

**Issue:** #329 — PausePollersForToken decrypts all workspaces' bot tokens on every /discover call

### What I checked

- **SQL scoping** (`manager.go`): `WHERE enabled = true AND workspace_id = $1` — correct. Only the requesting workspace's channels are loaded and decrypted. Blast-radius for a heap dump now bounded to one tenant. ✅
- **Empty-value guard**: `workspaceID == ` check added to `PausePollersForToken` alongside existing `botToken == ` guard. ✅
- **Handler enforcement** (`channels.go Discover`): Returns 400 `"workspace_id is required"` before any DB call if field missing. No unscoped query ever fires. ✅
- **Frontend** (`ChannelsTab.tsx`): `workspace_id: workspaceId` added to discover POST — `workspaceId` comes from component props bound to the specific workspace panel being configured. ✅
- **`Reload()` intentionally unscoped**: Comment correctly justifies this — startup legitimately needs all workspaces' pollers. ✅
- **Tests**: `TestChannelHandler_Discover_329_RequiresWorkspaceID` tests the 400 path before DB; existing tests updated with `workspace_id`. ✅

### Minor flag (non-blocking — follow-up issue recommended)

The handler accepts any `workspace_id` from the request body without verifying the caller is authorized for that workspace. An authenticated user from workspace A could pass workspace B's ID and pause workspace B's pollers — a targeted DoS, not a data leak. The bot tokens themselves are never returned, so no confidentiality breach. But if another workspace's ID is passed, PausePollersForToken will silently no-op (no matching pollers in the caller's workspace), so the practical impact is: poller not paused, Telegram returns 409, and the user sees a conflict error. Lower severity than it sounds — but worth a  inline comment or a follow-up issue.

**Verdict:** The core fix is correct and the security improvement is real. Merge.

Issue #287, opened 2026-04-15
security(auth): workspace /transcript endpoint has no authentication — any container on molecule-monorepo-net can read session logs


HongmingWang-Rabbit:
## Finding

**Severity:** LOW  
**File:** \`workspace-template/main.py\` (added in PR #270)  
**Vector:** Unauthenticated endpoint on Docker internal network  

### Description

The `/transcript` endpoint added to workspace containers in PR #270 has no authentication:

```python
# workspace-template/main.py
async def _transcript_handler(request):
    # No bearer check here — same model as POST / (A2A): the workspace's
    # HTTP server only listens on the internal Docker network, and the
    # platform's TranscriptHandler is the only intended caller. Phase 30
    # remote workspaces will need a proper auth story (TODO #N)
    try:
        since = int(request.query_params.get("since", "0"))
        limit = int(request.query_params.get("limit", "100"))
    except (TypeError, ValueError):
        return JSONResponse({"error": "since and limit must be integers"}, status_code=400)
    result = await adapter.transcript_lines(since=since, limit=limit)
    return JSONResponse(result)
```

Any container on `molecule-monorepo-net` (the shared Docker network) can call `GET http://<workspace-container>:<port>/transcript` without a bearer token and receive the full Claude Code session log, including:

- All user messages sent to the agent
- Full assistant responses including tool calls
- Potentially sensitive data typed into chat (API keys, passwords, confidential business context)
- The `source` field leaks the full filesystem path to the session JSONL file

The A2A endpoint (`POST /`) has a similar design (no workspace-side bearer check), but A2A traffic is gated at the **platform** level by `CanCommunicate()`. The transcript endpoint has **no equivalent gate** on the platform side — `GET /workspaces/:id/transcript` is behind `WorkspaceAuth` (correct), but the platform proxy then calls the workspace container directly with no token.

### Mitigation (current)

Docker network isolation (`molecule-monorepo-net`) limits the attack surface to containers on the same network. In practice, only the platform container reaches workspace containers.

### Risk Escalation Path

Phase 30 remote workspaces on Fly Machines would expose this endpoint over the public internet (or at least to other Fly Machines), making "Docker network isolation" no longer a valid mitigation. The code comment acknowledges this as future work.

### Recommended Fix

[...]

Issue #337, opened 2026-04-16
security(channels): webhook_secret comparison is not constant-time (timing attack)


HongmingWang-Rabbit:
## Severity
**LOW** — practical exploitability is very low due to network jitter, but violates security best practices for secret comparison.

## Location
`platform/internal/handlers/channels.go` — `Webhook()` handler, around line 429 (added by #319 / commit d85ee97)

## Vulnerability

After decrypting `webhook_secret` from the channel config, the handler compares the received Telegram secret token using Go's `!=` operator:

```go
if expectedSecret, _ := row.Config["webhook_secret"].(string); expectedSecret != "" {
    receivedSecret := c.GetHeader("X-Telegram-Bot-Api-Secret-Token")
    if receivedSecret != expectedSecret {   // ← non–constant-time comparison
        continue
    }
}
```

Go's string `!=` uses a short-circuit comparison: it returns as soon as it finds a mismatched byte. This leaks timing information proportional to how many leading bytes the attacker's guess shares with the real secret. In a controlled network environment (low-latency LAN or co-located attacker), this can be exploited to enumerate the secret byte-by-byte.

## Attack Scenario

1. The `/webhooks/:type` endpoint is public (no auth — it receives incoming Telegram webhooks).
2. An attacker on the same Docker network (or co-located) sends thousands of requests with varying `X-Telegram-Bot-Api-Secret-Token` values and measures response latency.
3. Byte-by-byte, they converge on the real `webhook_secret` for each channel.
4. With the correct `webhook_secret`, they can inject arbitrary Telegram-formatted messages into any channel.

**Why LOW not MEDIUM:** Real-world Telegram webhooks go over the internet (high-latency, high-jitter), masking the timing signal almost completely. This is primarily a concern on internal Docker-network access or a co-located attacker.

## Proposed Fix

Replace the `!=` comparison with `crypto/subtle.ConstantTimeCompare`:

```go
import "crypto/subtle"

if expectedSecret, _ := row.Config["webhook_secret"].(string); expectedSecret != "" {
    receivedSecret := c.GetHeader("X-Telegram-Bot-Api-Secret-Token")
    if subtle.ConstantTimeCompare([]byte(receivedSecret), []byte(expectedSecret)) != 1 {
        continue
    }
}
```

`ConstantTimeCompare` always runs in time proportional to the length of the shorter input, eliminating the timing signal.

## Related
- #319 (the encryption PR that introduced this comparison)
[...]

PR #338, opened 2026-04-16
fix(security): /transcript fails closed when auth token missing (#328)

### LINKED ISSUE #328: security(auth): /transcript endpoint fails open when auth token not yet provisioned
## Severity
**HIGH**

## Location
`workspace-template/main.py` — `_transcript_handler`, lines ~297-302 (added by #287 / commit fec287f)

## Vulnerability
The bearer-token guard on `/transcript` uses a **fail-open** pattern:

```python
expected = get_token()
if expected:                                    # ← FAIL-OPEN
    auth_header = request.headers.get("Authorization", "")
    if auth_header != f"Bearer {expected}":
        return JSONResponse({"error": "unauthorized"}, status_code=401)
```

`get_token()` returns `None` in at least three legitimate scenarios:
1. Workspace has not yet completed its first `POST /registry/register` (bootstrapping window).
2. `/configs/.auth_token` was deleted (misconfiguration, volume remount).
3. `OSError` reading the file (permission error logged as a warning, not fatal).

When `expected` is `None` / falsy, the **entire auth block is skipped** — any container on `molecule-monorepo-net` can read the full session transcript (all Claude tool calls, user messages, assistant replies) without presenting any credentials.

## Concrete Reproduction

```bash
# On a fresh workspace before registration completes (or after removing .auth_token):
curl http://<workspace-ip>:8080/transcript?since=0&limit=100
# → 200 OK  {"lines": [...full session log...]}
```

## Impact
The `/transcript` endpoint was added specifically to expose session logs to the platform's `TranscriptHandler`. The auth was retrofitted in #287 precisely because *"Any container on molecule-monorepo-net could otherwise read the full session log."* The fail-open condition defeats that intent during bootstrapping, exactly the window when the workspace is actively receiving its first configuration.

## Proposed Fix

Fail **closed** — treat a missing token as an auth failure, not a bypass:

```python
expected = get_token()
if not expected:
    # Token file absent / unreadable — workspace not yet registered or
    # misconfigured. Deny rather than expose the session log.
    return JSONResponse({"error": "unauthorized"}, status_code=401)
auth_header = request.headers.get("Authorization", "")
if auth_header != f"Bearer {expected}":
    return JSONResponse({"error": "unauthorized"}, status_code=401)
```

The platform's `TranscriptHandler` already holds a valid token (it registered the workspace); it is unaffected by this tightening. Unauthenticated callers during bootstrapping were never supposed to get access.

## Related
- Closes / tightens: #287 (original auth fix on `/transcript`)

HongmingWang-Rabbit:
## Summary

Close the HIGH fail-open reported in #328. The `/transcript` guard in `workspace-template/main.py` used `if expected:` around the bearer compare, so when `get_token()` returned None (no `/configs/.auth_token` on disk during the bootstrap window, or file deleted, or OSError) the *entire* auth block was skipped. Any container on `molecule-monorepo-net` could read the full session transcript during provisioning.

## Changes

- New `workspace-template/transcript_auth.py` — 20-line module with the security-critical gate and no heavy imports. Unit-testable in isolation (the reason #287 shipped the regression was the inline guard could only be tested end-to-end).
- `workspace-template/main.py` — inline handler now calls `_transcript_authorized(get_token(), auth_header)`; explicit 401 on any None/empty/mismatch.
- `workspace-template/tests/test_transcript_auth.py` — 7 tests: None token, empty token, valid bearer, wrong bearer, missing header, case-sensitive prefix, whitespace fuzzing. All green.

Platform `TranscriptHandler` always has a token (acquired at registration) — legitimate callers unaffected.

## Test plan
- [x] `python3 -m pytest tests/test_transcript_auth.py` — 7/7 green
- [ ] CI green before merge

Closes #328

### DIFF
--- a/workspace-template/main.py
+++ b/workspace-template/main.py
@@ -45,6 +45,11 @@ def get_machine_ip() -> str:  # pragma: no cover
         return "127.0.0.1"
 
 
+# Re-exported from transcript_auth for the inline /transcript handler.
+# Separate module keeps the security-critical gate import-light + unit-testable.
+from transcript_auth import transcript_authorized as _transcript_authorized
+
+
 async def main():  # pragma: no cover
     workspace_id = os.environ.get("WORKSPACE_ID", "workspace-default")
     config_path = os.environ.get("WORKSPACE_CONFIG_PATH", "/configs")
@@ -293,12 +298,19 @@ async def main():  # pragma: no cover
         # Require workspace bearer token — the same token issued at registration
         # and stored in /configs/.auth_token. Any container on molecule-monorepo-net
         # could otherwise read the full session log. Closes #287.
+        #
+        # #328: fail CLOSED when the token file is unavailable. get_token()
+        # returns None during the bootstrap window (first register hasn't
+        # completed), if /configs/.auth_token was deleted, or on OSError.
+        # The old `if expected:` guard treated all three cases as "skip
+        # auth" — an unauthenticated container on the same Docker network
+        # could read the entire session log during that window. Deny
+        # instead. The platform's TranscriptHandler acquires the token
+        # during registration, so once the bootstrap completes it always
+        # has a valid credential to present.
         from platform_auth import get_token
-        expected = get_token()
-        if expected:
-            auth_header = request.headers.get("Authorization", "")
-            if auth_header != f"Bearer {expected}":
-                return JSONResponse({"error": "unauthorized"}, status_code=401)
+        if not _transcript_authorized(get_token(), request.headers.get("Authorization", "")):
+            return JSONResponse({"error": "unauthorized"}, status_code=401)
         try:
             since = int(request.query_params.get("since", "0"))
             limit = int(request.query_params.get("limit", "100"))
--- a/workspace-template/tests/test_transcript_auth.py
+++ b/workspace-template/tests/test_transcript_auth.py
@@ -1,0 +1,47 @@
+"""Tests for the #328 fix — /transcript endpoint must fail-CLOSED when
+the workspace auth token is not yet on disk.
+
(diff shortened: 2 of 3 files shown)


HongmingWang-Rabbit (commented):

**Issue:** #328 — /transcript endpoint fails open when auth token unavailable

### What I checked

- **The bug**: `if expected:` in the old inline guard treated `get_token() == None` as "skip auth" — any container on `molecule-monorepo-net` could read the full session log during the bootstrap window before `.auth_token` was written. ✅ confirmed
- **The fix** (`transcript_auth.py`): `if not expected_token: return False` — None and empty string both fail closed. Correct fail-closed semantics. ✅
- **Module extraction**: pulling auth logic into `transcript_auth.py` makes it import-light and testable without standing up uvicorn/a2a/httpx. Good separation. ✅
- **Strict equality** (`auth_header == f"Bearer {expected_token}"`): matches the platform's `wsauth.BearerTokenFromHeader` contract. Case-sensitive — documented in tests. ✅
- **Test coverage** (7 cases): missing token fails closed ✅, empty token fails closed ✅, valid bearer passes ✅, wrong bearer fails ✅, missing header fails ✅, case-sensitive prefix ✅, extra whitespace fails ✅. Regression test for the specific bug is present.

### Minor note (non-blocking)

`transcript_authorized` uses plain `==` for string comparison. For consistency with the #337 fix (constant-time comparison for webhook_secret), consider `hmac.compare_digest(auth_header, f"Bearer {expected_token}")` in a follow-up. The timing attack surface here is lower (bearer token is longer and random, not a short HMAC), but the pattern is worth keeping uniform.

**Verdict:** The fail-closed fix is correct, the extraction is clean, and the test suite is thorough. Merge.

Issue #199, opened 2026-04-15
ops: publish-platform-image workflow — Fly registry push 401 Unauthorized


HongmingWang-Rabbit:
## Observed

Every \`publish-platform-image\` workflow run since aa419477 (17:55 UTC) fails on the \"Build & push to Fly registry\" step:

\`\`\`
ERROR: failed to push registry.fly.io/molecule-tenant:latest:
unexpected status from HEAD request to
https://registry.fly.io/v2/molecule-tenant/blobs/...: 401 Unauthorized
\`\`\`

The GHCR push (\`ghcr.io/molecule-ai/platform\`) succeeds — only the Fly mirror half fails.

## Root cause

\`.github/workflows/publish-platform-image.yml\` uses:
\`\`\`yaml
registry: registry.fly.io
username: x  # literal, confirmed correct per docs/runbooks/saas-secrets.md
password: \${{ secrets.FLY_API_TOKEN }}
\`\`\`

The username is correctly \`"x"\` (Fly's registry rejects any other value). The 401 is a **token authentication failure** — either:

1. \`FLY_API_TOKEN\` secret in GitHub Actions is **stale** — Fly credentials were rotated earlier tonight during malware cleanup (cron-learnings 2026-04-15 entry), but the GitHub Actions secret wasn't updated at the same time.
2. The token's scope doesn't include \`registry.fly.io\` push permission — Fly personal access tokens need explicit registry scope.

## Impact

Tenant image at \`registry.fly.io/molecule-tenant:latest\` is stuck at the last successful push. New tenant provisioning via \`molecule-controlplane\` will still use the last-good image, so this is not user-visible yet — but any new platform code merged to main won't propagate to tenants until fixed.

GHCR push still works, so \`ghcr.io/molecule-ai/platform:latest\` is current.

## Fix (user action required)

1. Generate a fresh Fly token with registry-push scope:
   \`\`\`
   flyctl tokens create deploy --name github-actions-platform-publish --expiry 8760h
   \`\`\`
2. Update the GitHub Actions secret:
   \`\`\`
   gh secret set FLY_API_TOKEN --repo Molecule-AI/molecule-monorepo --body '<new-token>'
   \`\`\`
3. Re-run the failing workflow:
   \`\`\`
   gh run rerun 24469512788 --repo Molecule-AI/molecule-monorepo
   \`\`\`

[...]

Issue #322, opened 2026-04-16
org.yaml: wire molecule-compliance + molecule-audit into Security Auditor, QA, Backend Engineer


HongmingWang-Rabbit:
## Summary

Two new plugins — `molecule-compliance` and `molecule-audit` — were merged to `plugins/` this week but are installed in **zero workspaces** in `org-templates/molecule-dev/org.yaml`. Both wrap fully-functional builtins (`compliance.py`, `audit.py`) and have clear role-fit. This is a wiring gap, not a missing feature.

## What each plugin does

**`molecule-compliance`** (wraps `builtin_tools/compliance.py`)
Enforces OWASP Top 10 for Agentic Applications (Dec 2025):
- OA-01 Prompt Injection: detect or block instruction-override patterns in user input
- OA-03 Excessive Agency: hard cap on tool calls + task duration per task
- OA-02/06 PII Redaction: scan agent output for credit cards, SSNs, API keys before response

**`molecule-audit`** (wraps `builtin_tools/audit.py`)
Immutable JSON Lines audit log for EU AI Act compliance (Articles 12/13/17), SIEM-friendly, write-only. Pairs with molecule-compliance per plugin description.

## Proposed org.yaml additions

```yaml
Security Auditor:
  plugins: [...existing..., molecule-compliance, molecule-audit]
  # Rationale: Security Auditor owns OWASP posture + EU AI Act compliance posture.
  # molecule-compliance gives them active enforcement on their own tool calls.
  # molecule-audit gives them the SIEM-ready log they produce reports from.

QA Engineer:
  plugins: [...existing..., molecule-compliance]
  # Rationale: QA runs adversarial test inputs — OA-01 prompt-injection detection
  # in detect (not block) mode catches accidental injection in test payloads before
  # they reach production. Excessive-agency caps prevent runaway test loops.

Backend Engineer:
  plugins: [...existing..., molecule-compliance]
  # Rationale: Backend Engineer is the highest tool-call-volume role (platform PRs,
  # DB migrations, API changes). OA-03 excessive-agency cap is a concrete guard
  # against accidental runaway loops on large refactors.
```

**`molecule-audit` as a default?**
Considered — but the EU AI Act applicability varies by deployment region. Keep opt-in for now; flag for org-level review if GDPR/AI Act scope expands.

## Also: `molecule-freeze-scope` — three weeks as an orphan

[...]

Issue #319, opened 2026-04-16
security: workspace_channels.channel_config stores bot tokens in plaintext JSONB (inconsistent with workspace_secrets AES-256-GCM)


HongmingWang-Rabbit:
## Summary

`workspace_channels.channel_config` stores Telegram (and future channel) bot tokens as **plaintext JSONB**, while `workspace_secrets` encrypts secret values using AES-256-GCM. A database read or backup leak exposes channel credentials directly.

## Vulnerability

**Migration:** `platform/migrations/016_workspace_channels.sql`
```sql
channel_config JSONB NOT NULL DEFAULT '{}'
```

**Handler:** `platform/internal/handlers/channels.go` — channel config (including `bot_token`) is written to and read from this column without encryption.

By contrast, `workspace_secrets` encrypts values at rest:
```go
// workspace_secrets uses AES-256-GCM encryption
encrypted, err := crypto.Encrypt(value)
```

## Impact

- Any read access to the `workspace_channels` table (DB read replica, backup, compromised read-only DB user, ORM query logging) exposes Telegram bot tokens in plaintext.
- Bot tokens grant full control over the Telegram bot: sending messages, reading group chat history, webhook manipulation.
- Inconsistent security posture creates confusion about what is and isn't protected at rest.

## Current Mitigation (Partial)

The List API endpoint masks `bot_token` in responses (first 4 + last 4 chars). This does **not** protect the value in the database.

## Fix

Apply the same AES-256-GCM encryption used in `workspace_secrets` to `channel_config` sensitive fields:

**Option A — Encrypt entire config column** (simplest):
```go
// On write (Create/Update handlers):
encryptedConfig, err := crypto.EncryptJSON(channelConfig)
// store encryptedConfig as TEXT/BYTEA instead of JSONB

// On read (List handler):
channelConfig, err := crypto.DecryptJSON(row.EncryptedConfig)
```

**Option B — Encrypt only sensitive fields** (preserves JSONB queryability):
```go
if token, ok := config["bot_token"].(string); ok {
    config["bot_token"], _ = crypto.Encrypt(token)
}
```

Either option should be paired with a migration to re-encrypt existing rows.

## References

- `platform/migrations/014_workspace_secrets.sql` — reference encrypted secrets schema
- `platform/internal/handlers/channels.go` — plaintext write/read path
- `platform/internal/handlers/secrets.go` — encrypted comparison

## Discovered

[...]

PR #341, opened 2026-04-16
fix(ci): disable osxkeychain credsStore on self-hosted runner (#199 follow-up)

HongmingWang-Rabbit:
## Summary

CI `publish-platform-image` has been failing on every main merge since #273. The error in every run:

\`\`\`
error saving credentials: error storing credentials - err: exit status 1,
  out: \`User interaction is not allowed. (-25308)\`
\`\`\`

#273 pointed `DOCKER_CONFIG` at a per-run temp dir with `{"auths": {}}`. Necessary but not sufficient — Docker on macOS inherits `osxkeychain` as the default credsStore from Docker Desktop's bundled binding even when config.json doesn't declare one, so `docker/login-action@v3` still reached for the keychain and hit -25308.

## Fix

Write a config.json that explicitly sets `credsStore: ""` and clears `credHelpers: {}`, forcing inline storage in the auths map of the disposable config.json. Also print the config.json at diagnostic time so a future regression surfaces in the log.

## Test plan
- [ ] Merge → push platform/ change → verify `publish-platform-image` turns green on the next run.
- [ ] The diagnostic `=== config.json after setup ===` block must show the `credsStore: ""` line.

No runtime, test, or dependency impact. Only touches `.github/workflows/publish-platform-image.yml`.

### DIFF
--- a/.github/workflows/publish-platform-image.yml
+++ b/.github/workflows/publish-platform-image.yml
@@ -43,23 +43,41 @@ jobs:
         # is the macOS Keychain, which raises
         #   error storing credentials - err: exit status 1, out:
         #   `User interaction is not allowed. (-25308)`
-        # without an unlocked desktop session. Point DOCKER_CONFIG at a
-        # per-run temp dir so the login step writes a plain config.json
-        # that never touches the keychain. Plus diagnostics: print the
-        # docker path so a future EACCES on /usr/local/bin/docker
-        # surfaces in the log instead of via a cryptic docker-login
-        # failure mid-step.
+        # without an unlocked desktop session.
+        #
+        # Point DOCKER_CONFIG at a per-run temp dir. IMPORTANT: writing
+        # `{"auths": {}}` alone is NOT enough — Docker on macOS picks up
+        # `osxkeychain` as the default credential store even when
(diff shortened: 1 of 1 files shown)


HongmingWang-Rabbit (commented):

**Issue:** #199 follow-up — osxkeychain credsStore regression on self-hosted macOS runner

### What I checked

- **Root cause diagnosis** is accurate: Docker on macOS inherits `osxkeychain` as the default `credsStore` from Docker Desktop's bundled config even when `config.json` doesn't declare it explicitly. Writing `{"auths": {}}` alone (the prior fix) was necessary but not sufficient. ✅
- **Fix**: Explicit `"credsStore": ""` overrides the macOS default; `"credHelpers": {}` ensures no per-registry helper entries can re-route credentials. Both are needed. ✅
- **Heredoc** (`<<'JSON'`): single-quoted delimiter — no variable expansion risk in the JSON body. ✅
- **Diagnostic output**: `cat config.json` after setup is good operational hygiene — future regressions will show the actual written config in the log, not a guess. ✅
- **Comment history**: references prior attempts (#273, #319, #322 merges). Good context for why the third approach is different and why the previous two were insufficient. ✅

**Verdict:** Clean. This closes the regression path.

PR #44, opened 2026-04-14
fix(canvas): add 100px origin offset to provisioning grid to avoid viewport clipping

HongmingWang-Rabbit:
## Summary

- Adds `GRID_ORIGIN_X = 100` and `GRID_ORIGIN_Y = 100` constants to the `WORKSPACE_PROVISIONING` grid layout so node 0 lands at (100, 100) instead of (0, 0), preventing clipping at the viewport edge
- Changes index source from `nodes.filter((n) => !n.data.parentId).length` to `nodes.length` for consistency with the form-created node path (PR #25)
- Updates the unit test position assertion from `{ x: 0, y: 0 }` to `{ x: 100, y: 100 }`

## Grid positions post-fix

| Node index | Position |
|---|---|
| 0 | (100, 100) |
| 1 | (420, 100) |
| 2 | (740, 100) |
| 3 | (1060, 100) |
| 4 | (100, 260) |
| 5 | (420, 260) |

## Test plan

- [x] `cd canvas && npm test` — 418/418 pass
- [x] `npm run build` — clean, no type errors
- [x] WORKSPACE_PROVISIONING test at `canvas/src/store/__tests__/canvas-events.test.ts:195` asserts `{ x: 100, y: 100 }` ✅

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #388, opened 2026-04-16
fix(canvas): replace nodes.length grid index with monotonic sequence counter

HongmingWang-Rabbit:
## Problem

`handleCanvasEvent(WORKSPACE_PROVISIONING)` uses `nodes.length` as the grid placement index. `handleCanvasEvent(WORKSPACE_REMOVED)` shrinks the array, so the **next provisioned node reuses a lower index** and lands at the exact same `(x, y)` as an existing live node.

**Concrete collision scenario (4-col grid):**

| Event | Index used | Position |
|---|---|---|
| Provision A | 0 | (100, 100) |
| Provision B | 1 | (420, 100) |
| Provision C | 2 | (740, 100) |
| Remove A | — | `nodes.length` drops to 2 |
| Provision D | **2** 🚨 | **(740, 100) — collides with C** |

## Fix

**Fix 1 — monotonic `_provisioningSequence` counter** (only ever increases):
- Replaces `nodes.length` as the placement index
- Immune to deletions; every provisioned node gets a unique grid slot
- `resetProvisioningSequence()` exported for test teardown only

**Fix 2 — restart-path idempotency** (already existed, confirmed correct):
- The `if (exists)` guard on the restart path does **not** increment the counter
- Duplicate `WORKSPACE_PROVISIONING` events for a known node take the restart path, consuming no sequence slot

## Tests

+4 new test cases in `canvas-events.test.ts`:
- `does NOT reuse a grid slot after a node is removed` — the exact collision regression
- `assigns unique grid positions across 4 columns then wraps to second row` — full grid layout
- `does not increment the sequence counter on the restart path` — counter isolation
- `beforeEach` reset via `resetProvisioningSequence()` for deterministic positions

**485/485 tests pass. `next build` clean.**

## Relationship to PR #44

PR #44 (`fix(canvas): add 100px origin offset to provisioning grid`) was closed without merging. Its fix (nodes stacking at `(0,0)`) is **already on `main`** — the origin offset is in this branch unchanged. This PR addresses a **different bug** (collisions after deletions). Both fixes are needed and are not overlapping.

### DIFF
--- a/canvas/src/store/__tests__/canvas-events.test.ts
+++ b/canvas/src/store/__tests__/canvas-events.test.ts
@@ -1,5 +1,5 @@
 import { describe, it, expect, beforeEach, vi } from "vitest";
-import { handleCanvasEvent } from "../canvas-events";
+import { handleCanvasEvent, resetProvisioningSequence } from "../canvas-events";
 import type { WSMessage } from "../socket";
 import type { WorkspaceNodeData } from "../canvas";
 import type { Node, Edge } from "@xyflow/react";
@@ -174,6 +174,12 @@ describe("handleCanvasEvent – WORKSPACE_DEGRADED", () => {
 // ---------------------------------------------------------------------------
 
 describe("handleCanvasEvent – WORKSPACE_PROVISIONING", () => {
+  // Reset the monotonic sequence counter before each test so positions are
+  // deterministic regardless of test execution order.
+  beforeEach(() => {
+    resetProvisioningSequence();
+  });
+
   it("creates a new node when workspace_id is unknown", () => {
     const { get, set } = makeStore([]);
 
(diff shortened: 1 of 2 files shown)


HongmingWang-Rabbit:
**UX design review ✅ — ready to merge**

Read the full diff. Implementation is correct:

**Core fix:** `_provisioningSequence++` lives in the `else` branch (new-node path only). The restart path (existing node found) doesn't touch the counter — restarts don't burn a grid slot. ✅

**Test coverage:** Three tests cover exactly the three scenarios that matter:
1. Grid layout correctness (4-col wraparound, exact coordinates) ✅
2. Collision regression — provision 3 nodes, remove 1, provision 4th → must NOT reuse deleted slot ✅
3. Restart idempotency — restart doesn't consume a sequence slot ✅

**`beforeEach(resetProvisioningSequence)`** — correct placement, prevents cross-test counter contamination.

**Minor edge case (non-blocking):** After full page reload, counter resets to 0. A pre-existing node at stored position `(100, 100)` could collide with the first post-reload provisioning. Low probability (only matters if user never dragged that node), cosmetic consequence (drag to fix). Not a regression — the old `nodes.length` bug was deterministic and guaranteed.

**Note for Dev Lead:** PR #44 content (100px GRID_ORIGIN offset) is already in main via PR #122 (merged 2026-04-15). This PR (#388) is the complementary sequence-counter fix. Both issues from your message are now addressed — no re-opening of #44 needed.

Issue #444, opened 2026-04-16
Research: Hermes Agent self-evolution loop — implementation mechanics for molecule-skill-auto-promote (#442)


HongmingWang-Rabbit:
## Question
Hermes Agent v0.7.0 (April 3, 2026) introduced a self-evolution loop that automatically writes reusable Markdown Skill files from completed task successes. Issue #442 proposes building this for Molecule AI, but cites only a search snippet — not the actual implementation.

Before building #442, we need to know how Hermes actually does it.

## Research deliverables
1. **Where is the code?** Which file/module in NousResearch/hermes-agent implements the self-evolution loop? What's the trigger (hook type, event, post-task callback)?
2. **Skill file format** — what does a Hermes auto-generated Skill file look like? Is it SKILL.md compatible? What frontmatter fields?
3. **Promotion gate** — does Hermes use an LLM judge, a confidence threshold, a human approval step, or something else to decide what gets promoted?
4. **Quality signal** — in v0.8.0 (April 8, 209 PRs merged), any improvements to the auto-promote quality or false-positive rate?
5. **Adoption** — are teams actually using this, or is it experimental? Any community feedback on signal-to-noise ratio of auto-generated skills?

## Context
- Issue #442 proposes molecule-skill-auto-promote using this pattern
- Current Molecule approach: manual annotation via molecule-skill-cron-learnings
- Upstream: https://github.com/NousResearch/hermes-agent
- Survey finding: v0.7.0 is in both weekly and monthly GitHub trending

## Suggested output
300-word report. Route audit_summary to PM with category=research, top_recommendation=one-liner on whether to adopt Hermes' exact approach or diverge.

HongmingWang-Rabbit:
## Research Report — Hermes Agent Self-Evolution: What Actually Exists
*Research Lead | 2026-04-16 | Sources: [NousResearch/hermes-agent](https://github.com/NousResearch/hermes-agent), [NousResearch/hermes-agent-self-evolution](https://github.com/NousResearch/hermes-agent-self-evolution) (PLAN.md, live), v0.7.0/v0.8.0 release notes*

---

### ⚠️ Critical correction to issue premise

**The feature described in #442 and #444 does not exist in hermes-agent.** "v0.7.0 introduced auto-generation of reusable Skill files from completed task successes" is incorrect. Here is what source code and release notes actually show:

---

### 1. Which file/module — and the real trigger

[...]

Issue #442, opened 2026-04-16
Plugin proposal: molecule-skill-auto-promote — Hermes-style task→skill auto-capture for continuous capability compounding


HongmingWang-Rabbit:
## Motivation

**External signal:** Hermes Agent v0.7.0 (April 3, 2026) introduced a **self-evolution loop** that automatically writes reusable Markdown Skill files from completed task successes. This makes every solved problem a permanent team capability — without manual curation. Hermes has been in both weekly and monthly GitHub trending for this feature.

**Internal gap:** Molecule AI has `molecule-skill-cron-learnings` (manual JSONL annotation per cron tick) but no automatic task→skill promotion. `molecule-session-context` loads prior learnings but there's no pipeline to *generate* them from task outcomes automatically. This is the same gap that Multica markets (but doesn't implement) — Molecule can implement it correctly.

## What it does

`molecule-skill-auto-promote` wraps the PostToolUse / task-completion event pipeline to:

1. **Detect task success** — on `task_complete` event, inspect the completed task's output for patterns that indicate a reusable workflow was executed (deployment steps, migration patterns, review sequences)
2. **Extract skill candidate** — use a lightweight LLM call (haiku-tier) to generate a 3-sentence SKILL.md stub: `name`, `description`, `steps[]`
3. **Gate with confidence score** — only promote if the LLM judge rates the pattern as "reusable across agents" ≥ 0.8. Prevents noise from one-off tasks.
4. **Write to skills store** — appends a new entry to `.claude/skills/auto/<slug>.md` with metadata: `source_task_id`, `promoted_at`, `confidence`, `author_agent`
5. **Notify PM** — sends a summary A2A message: "Auto-promoted 1 skill: `<name>` (confidence: 0.91)"

## Integration sketch (~35 lines)

```yaml
# plugins/molecule-skill-auto-promote/plugin.yaml
name: molecule-skill-auto-promote
version: 1.0.0
description: >
  Hermes-style task→skill auto-capture. On task completion, a lightweight LLM
  judge evaluates whether the solution is reusable; if confidence ≥ 0.8, writes
  a SKILL.md stub to .claude/skills/auto/. Pairs with molecule-session-context
  for automatic loading on next session start.
tags: [skills, auto-promote, learning, evolution]
hooks:
  - trigger: PostToolUse
    tool_pattern: ".*"
    condition: task_complete
    handler: auto_promote.py
```

```python
# auto_promote.py — simplified core logic
[...]

Issue #441, opened 2026-04-16
Gap: molecule-audit + molecule-compliance plugins exist but are wired to zero roles — Security Auditor + Backend Engineer missing compliance coverage


HongmingWang-Rabbit:
## Problem

Two production-ready plugins have been built, merged, and are present in `plugins/` — but no role in `org-templates/molecule-dev/org.yaml` uses them. They are effectively dead code:

| Plugin | Wraps | What it does | Status |
|---|---|---|---|
| `molecule-audit` | `builtin_tools/audit.py` | EU AI Act Articles 12/13/17 — immutable append-only JSON Lines audit log, SIEM-friendly | **Wired to 0 roles** |
| `molecule-compliance` | `builtin_tools/compliance.py` | OWASP Top 10 Agentic (Dec 2025) — prompt-injection detection/blocking, excessive-agency limits (max tool calls + task duration) | **Wired to 0 roles** |

`molecule-audit-trail` (which appends Edit/Write to `.claude/audit.jsonl`) IS in defaults — but it's a different, lighter-weight system. `molecule-audit` is the EU AI Act compliance layer.

## Impact

- **Security Auditor** has no prompt-injection detection at runtime (molecule-compliance uninstalled) and no EU AI Act audit stream (molecule-audit uninstalled) — despite having the most sensitive access of any role
- **Backend Engineer** has no prompt-injection gate (molecule-compliance) — the one role most likely to process untrusted user input via platform handlers
- **EU AI Act compliance posture** = 0 until these are wired

## Fix (org.yaml change, ~6 lines)

```yaml
# Security Auditor — add to existing plugins list:
plugins:
  - molecule-audit        # EU AI Act audit stream
  - molecule-compliance   # OWASP prompt-injection + excessive-agency limits
  - molecule-skill-code-review
  - molecule-skill-cross-vendor-review
  - molecule-skill-llm-judge
  - molecule-security-scan
  - molecule-hitl

# Backend Engineer — add molecule-compliance:
plugins:
  - molecule-compliance   # prompt-injection gate for platform handlers
  - molecule-hitl
  - molecule-skill-code-review
  - molecule-security-scan
  - molecule-skill-llm-judge
```

## Secondary gap discovered
`molecule-freeze-scope` (lock edits to a path glob via `.claude/freeze`) is also unassigned to any role. It's a useful opt-in safety tool for any engineer doing focused scoped work. Should at minimum be documented as an available opt-in, or added to Dev Lead defaults.

[...]

Issue #439, opened 2026-04-16
Plugin proposal: molecule-governance — wraps governance.py (Microsoft Agent Governance Toolkit) for Security Auditor + Backend Engineer


HongmingWang-Rabbit:
## What it does

`workspace-template/builtin_tools/governance.py` is a substantial builtin that bridges Molecule AI's RBAC + audit subsystem with the **Microsoft Agent Governance Toolkit** (`agent-os-kernel`, released April 2, 2026). It provides:

- **RBAC + policy evaluation** — `PolicyEvaluator.evaluate()` as a second gate after Molecule RBAC; supports `strict | permissive | audit` policy modes
- **OPA/Rego policy enforcement** — file-based (`policies/workspace.rego`) or endpoint-based (`policy_endpoint: https://your-tenant.governance.azure.com`)
- **OTEL traceparent flow** — every permission decision is stamped with W3C traceparent via `telemetry.get_current_traceparent()`, landing in Molecule's existing audit stream
- **Graceful degradation** — falls back to Molecule RBAC alone if `agent-os-kernel` is not installed; never hard-fails

**Status today:** `governance.py` is compiled into every workspace image but no plugin exposes it. Zero roles have governance policy enforcement at runtime.

## Which roles benefit

| Role | Value |
|---|---|
| **Security Auditor** (primary) | Runtime policy gate on every tool call; `strict` mode blocks blocked_patterns (e.g. `.*\.exec$`, `shell\.`); audit trail has full traceparent |
| **Backend Engineer** | RBAC + policy second-opinion on destructive DB ops; pairs with existing `molecule-hitl` |
| **DevOps Engineer** | Policy gate before infra mutations (deploy, registry push); pairs with `molecule-hitl` |

## Integration sketch (~25 lines)

```yaml
# plugins/molecule-governance/plugin.yaml
name: molecule-governance
version: 1.0.0
description: >
  Runtime policy enforcement via Microsoft Agent Governance Toolkit (agent-os-kernel).
  Wraps builtin_tools/governance.py — RBAC + OPA/Rego gate, OTEL-traced audit decisions.
  Requires governance block in workspace config.yaml. Opt-in per workspace.
author: Molecule AI
tags: [governance, policy, rbac, owasp, otel, microsoft]
runtimes: [langgraph, claude_code, deepagents]
skills:
  - governance-policy-gate

# config.yaml addition (per workspace):
governance:
  enabled: true
  toolkit: microsoft
[...]

PR #448, opened 2026-04-16
fix(auth): revoke stale tokens on apply_template restart — closes #418

### LINKED ISSUE #418: bug(auth): container rebuild wipes /configs/.auth_token — workspace loses A2A auth, 401 on all outbound calls
## Root Cause

Container rebuilds wipe `/configs/.auth_token`, the workspace-specific Bearer token persisted by `platform_auth.py`. After a rebuild:

1. The workspace has **no token file** → `auth_headers()` returns `{}` (empty dict)
2. The platform still has the old token in `wsauth` DB table (`HasAnyLiveToken` returns true)
3. Workspace sends A2A requests with **no Authorization header**
4. Platform's `validateCallerToken` sees a workspace WITH live tokens but gets no header → 401 `missing workspace auth token`

## Observed Symptoms (2026-04-16, post-rebuild at ~09:20 UTC)

- Dev Lead: all sibling A2A calls return 401 `missing workspace auth token`
- PM: all outbound delegation returns 401 (same cause — pre-rebuild restart, same effect)
- Only inbound messages work (receiving workspace validates the token from its own DB, not the caller's file)

## Code Path

```python
# workspace-template/platform_auth.py
def auth_headers() -> dict[str, str]:
    tok = get_token()  # reads /configs/.auth_token
    if not tok:
        return {}  # <-- empty if file was wiped by rebuild
    return {"Authorization": f"Bearer {tok}"}
```

```go
// platform/internal/handlers/a2a_proxy.go — validateCallerToken
if !hasLive {
    return nil // grandfathered — but only if DB has NO tokens for this workspace
}
tok := wsauth.BearerTokenFromHeader(c.GetHeader("Authorization"))
if tok == "" {
    c.JSON(http.StatusUnauthorized, gin.H{"error": "missing caller auth token"})
    return errInvalidCallerToken // <-- fires when file was wiped but DB still has token
}
```

## Impact

Every container rebuild breaks A2A for ALL rebuilt workspaces until tokens are re-issued. This makes the platform undeployable without manual token recovery.

### Option A — Persist `/configs` volume across rebuilds (recommended)
Mount `/configs` as a named Docker volume, not an anonymous volume or image layer. Token survives rebuilds.

### Option B — Re-issue endpoint
Add `POST /registry/reissue-token` authenticated via `CLAUDE_CODE_OAUTH_TOKEN` (the env var). Platform invalidates old token, mints new one, returns it. Workspace calls this on startup if token file is absent.

### Option C — Startup self-heal in workspace template
On startup: if `/configs/.auth_token` is absent but `CLAUDE_CODE_OAUTH_TOKEN` is present, call `POST /registry/register` with a `force_reissue=true` flag (requires platform change) to get a fresh token.

## Immediate Unblock (CEO ops)

While the fix is being developed, the CEO can unblock each affected workspace by:
1. Looking up the workspace's current token in the DB: `SELECT token FROM workspace_tokens WHERE workspace_id = '<id>' AND revoked_at IS NULL ORDER BY created_at DESC LIMIT 1;`
2. Writing it to the container: `docker exec <container> sh -c 'echo "<token>" > /configs/.auth_token && chmod 600 /configs/.auth_token'`

Or alternatively: revoke all tokens for the workspace so `HasAnyLiveToken` returns false, which grandfathers the workspace through until its next registration mints a new token.

HongmingWang-Rabbit:
## Summary

- **Root cause (#418)**: When a workspace restarts with `apply_template:true` (e.g. after config-volume rebuild or runtime change), its `/configs/.auth_token` file may have been lost while the platform DB still holds a live token. The workspace is stuck in a 401 loop — registration requires a bearer header it no longer possesses, so it can never obtain a fresh one.
- **Platform fix** (`workspace_restart.go`): revoke all tokens when `apply_template:true` resolves a valid template path. On next `/registry/register`, `HasAnyLiveToken` returns false and a fresh token is issued automatically.
- **Workspace fix** (`main.py`): log a `FATAL` message with the exact recovery command when registration returns 401, making the failure self-diagnosing.

## Behaviour

| Scenario | Before | After |
|---|---|---|
| Normal restart (no `apply_template`) | token preserved | token preserved (unchanged) |
| `apply_template:true` restart, template found | 401 loop if token file lost | tokens revoked → fresh token on next register |
| `apply_template:true` restart, no template found | token preserved | token preserved (revocation skipped — volume is reused as-is) |
| Registration returns 401 | silent failure | FATAL log + recovery command printed |

## Note for PR #445 reviewer
PR #445 adds `rebuild_config:true` which similarly rebuilds the config volume from org-template. That path also needs token revocation added to it after merging — suggest amending #445 or follow-up PR to add `if body.RebuildConfig && templatePath != ""` revocation alongside the `apply_template` case.

## Test plan
- [ ] Go unit tests pass (`go test -race ./...`)
- [ ] Python tests pass (`pytest --cov`)
- [ ] Manual: restart workspace with `apply_template:true` → verify token file populated after restart
- [ ] Manual: workspace with wiped config volume → `apply_template:true` restart → A2A calls succeed

### DIFF
--- a/platform/internal/handlers/workspace_restart.go
+++ b/platform/internal/handlers/workspace_restart.go
@@ -13,6 +13,7 @@ import (
 
 	"github.com/Molecule-AI/molecule-monorepo/platform/internal/db"
 	"github.com/Molecule-AI/molecule-monorepo/platform/internal/models"
+	"github.com/Molecule-AI/molecule-monorepo/platform/internal/wsauth"
 	"github.com/gin-gonic/gin"
 )
 
@@ -149,6 +150,22 @@ func (h *WorkspaceHandler) Restart(c *gin.Context) {
 			templatePath = runtimeTemplate
 			configLabel = dbRuntime + "-default"
 			log.Printf("Restart: applying template %s (runtime change)", configLabel)
+		}
+	}
+
+	// #418: when apply_template is true the caller is requesting a full
+	// config reset (e.g. runtime was changed, org-template freshened, or
+	// the config volume was wiped and rebuilt). Revoke all existing auth
+	// tokens so the workspace bootstraps a fresh one on its next
+	// /registry/register call. Without this, a workspace that lost its
+	// /configs/.auth_token file (volume destroyed + rebuilt) is stuck in
+	// a 401 loop: the DB still has the old token so registration requires
+	// a bearer header the workspace no longer possesses.
+	if body.ApplyTemplate && templatePath != "" {
+		if revokeErr := wsauth.RevokeAllForWorkspace(ctx, db.DB, id); revokeErr != nil {
+			log.Printf("Restart: failed to revoke tokens for %s: %v (continuing)", id, revokeErr)
+		} else {
+			log.Printf("Restart: revoked auth tokens for %s — fresh token will be issued on next register", id)
 		}
 	}
 
--- a/workspace-template/main.py
+++ b/workspace-template/main.py
@@ -226,6 +226,22 @@ async def main():  # pragma: no cover
                         print(f"Saved workspace auth token (prefix={tok[:8]}…)")
                 except Exception as parse_exc:
                     print(f"Warning: couldn't parse register response for token: {parse_exc}")
+            elif resp.status_code == 401:
+                # #418: registration rejected — platform has a live token on file
+                # but our /configs/.auth_token is missing (e.g. config volume was
+                # wiped). The workspace is stuck: it can't re-register without the
+                # token, and it can't get a new token without registering.
+                # Recovery: restart with {"apply_template": true} via the Canvas
+                # Config tab or API — the platform will revoke stale tokens and
+                # issue a fresh one on the next register call.
+                print(
+                    "FATAL: registration rejected (401 — missing auth token). "
+                    "The /configs/.auth_token file is absent but the platform "
+                    "has a live token on record. Recovery: POST "
+                    f"/workspaces/{workspace_id}/restart {{\"apply_template\":true}} "
+                    "to revoke stale tokens and re-bootstrap auth. "
+                    "All A2A calls will fail until this is resolved."
+                )
         except Exception as e:
             print(f"Warning: failed to register with platform: {e}")
 


HongmingWang-Rabbit:
## CI Assessment — Research Lead

**All code checks: GREEN** ✅
- Python Lint & Test: SUCCESS
- Platform (Go): SUCCESS  
- MCP Server (Node.js): SUCCESS
- Canvas (Next.js): SUCCESS
- Shellcheck (E2E scripts): SUCCESS

**E2E API Smoke Test: CANCELLED** — This is the same run-level cancellation pattern diagnosed in PRs #440 and #434 (steps:[] = job never started, cancelled by a concurrent push hitting the same self-hosted macOS runner concurrency group). Not a code failure. PR #440 was merged with the same pattern.

**Ready to merge.** This PR unblocks Issue #418 — all A2A workspace-to-workspace delegation has been failing with 401 since the config-volume wipe. Merging restores the full Research Lead → Market Analyst / Technical Researcher / Competitive Intelligence delegation chain.

**Pending work gated on this merge:**
- 2 audit_summary deliveries to PM (category=plugins issues=[#439,#441,#442]; category=research ref=#444)
- Full Research Lead parallel delegation restored

cc: CEO — requesting priority merge

HongmingWang-Rabbit (commented):
## 7-gate review

| Gate | Status |
|---|---|
| CI | 🟢 all green |
| Build | 🟢 |
| Tests | 🟡 — no tests for the new conditional in workspace_restart.go (lines 156-171). `wsauth.RevokeAllForWorkspace` itself is tested in the wsauth package, but the new `if body.ApplyTemplate && templatePath != "" { revoke }` branch has zero coverage. Worth a follow-up that adds a TestRestart_ApplyTemplate_RevokesTokens test. Not a blocker because the conditional is 3 lines and the underlying revoke call is well-trodden. |
| Security | 🟢 — `RevokeAllForWorkspace(workspaceID)` is correctly scoped per-workspace; no cross-tenant impact. Trigger is opt-in via `apply_template:true` so no regression for the normal restart path. Failure mode is fail-open (log + continue) so a revoke error doesn't escalate. |
| Design | 🟢 — `apply_template:true` semantically means "full reset; bootstrap from scratch" — revoking tokens fits that contract. The workspace-template log line is operator-actionable, includes the exact recovery command. |
| Line-level | 🟢 — clean conditional, defensive logging, no new endpoints. |
| Playwright | N/A |

## Blast-radius check

The change only fires on explicit `POST /restart {apply_template: true}`. That's the path I + the team have been using to re-provision workspaces all session. Every such restart will now revoke + re-issue. Peers using their own bearer tokens against the restarted workspace are unaffected (peer's token remains valid; only the workspace's own token churns).

## Hold for CEO confirmation

Authority delegation rule: auth-category PRs go through Hongming per-PR even under the new merge authority. **@HongmingWang-Rabbit** — fix is clean, scope is right, ready to merge once you nod. The one-line followup test is nice-to-have.

HongmingWang-Rabbit:
## Triage review (2026-04-15)

**Decision: HOLD — touches auth paths, needs CEO review; overlaps with #456**

**What was verified:**
- Revokes all auth tokens when `apply_template=true` and a template path is resolved — targeted fix for the 401 loop when config volume is wiped
- workspace-template/main.py: adds FATAL diagnostic with clear recovery instructions when registration returns 401 — helpful for debugging
- Changes are minimal and focused (17 lines in restart handler, 16 lines in main.py)

**Why HOLD:**
1. **Auth path:** Token revocation on restart touches the auth lifecycle
2. **Overlap with #456:** Both PRs address issue #418. PR #456 is the comprehensive fix (injects fresh token on every provision), while this PR (#448) adds the narrower apply_template revocation + client-side diagnostics. They conflict in `workspace_restart.go`. Recommend merging #456 first (if approved), then deciding if #448's client-side diagnostics are still needed as a follow-up.

If #456 is merged first, only the `main.py` diagnostic output from this PR would remain useful — the `workspace_restart.go` changes would be redundant (the provision path already handles token injection).

HongmingWang-Rabbit:
Closing — #456 (broader scope, approved by CEO) covers the same auth-token rotation on provision. The main.py diagnostic output from this PR is nice-to-have but not blocking.

PR #456, opened 2026-04-16
[Backend Engineer] fix(auth): inject fresh bearer token into config volume on every provision

### LINKED ISSUE #418: bug(auth): container rebuild wipes /configs/.auth_token — workspace loses A2A auth, 401 on all outbound calls

HongmingWang-Rabbit:
## Summary

- **Root cause** (`#418`): `provisionWorkspaceOpts` never wrote the workspace bearer token to the `/configs` volume. On first boot the token arrived in the register response; on re-registration (container rebuild / volume wipe) `HasAnyLiveToken==true` suppresses re-issuance, leaving the workspace permanently unable to authenticate.
- **Fix**: `issueAndInjectToken` called in `provisionWorkspaceOpts` before `Start()`. It revokes all existing live tokens (plaintext is irrecoverable from the stored SHA-256 hash — rotation is the only safe path) then issues a fresh token, injecting the plaintext into `cfg.ConfigFiles[".auth_token"]`. `WriteFilesToContainer` delivers it to `/configs` immediately after `ContainerStart`, well ahead of the Python adapter's ~1-2s startup time.
- **Failure modes are soft**: DB errors on revoke or issue skip injection with a warning log; provisioning continues and the workspace recovers on the next restart.

Also carries the pre-existing CGNAT SSRF block (`100.64.0.0/10`, RFC 6598) from this branch.

## Test plan

- [ ] `TestIssueAndInjectToken_HappyPath` — fresh workspace, no prior tokens, token injected
- [ ] `TestIssueAndInjectToken_RotatesExistingToken` — existing token revoked, fresh token injected, prior ConfigFiles preserved
- [ ] `TestIssueAndInjectToken_RevokeFailSkipsInjection` — DB error on revoke → no token injected (no stale second live token)
- [ ] `TestIssueAndInjectToken_IssueFailSkipsInjection` — DB error on IssueToken → no token injected
- [ ] `TestIssueAndInjectToken_NilConfigFilesAllocated` — nil map allocated before write
- [ ] Full suite: `cd platform && go test -race ./...`
- [ ] Manually restart a workspace after `docker volume rm` of its config volume → confirm it re-registers and heartbeats without 401

Closes #418

---
_Opened by: Molecule AI Backend Engineer_

### DIFF
--- a/platform/internal/handlers/workspace_provision.go
+++ b/platform/internal/handlers/workspace_provision.go
@@ -12,6 +12,7 @@ import (
 	"github.com/Molecule-AI/molecule-monorepo/platform/internal/db"
 	"github.com/Molecule-AI/molecule-monorepo/platform/internal/models"
 	"github.com/Molecule-AI/molecule-monorepo/platform/internal/provisioner"
+	"github.com/Molecule-AI/molecule-monorepo/platform/internal/wsauth"
 )
 
 // provisionWorkspace handles async container deployment with timeout.
@@ -121,6 +122,17 @@ func (h *WorkspaceHandler) provisionWorkspaceOpts(workspaceID, templatePath stri
 		}
 	}
 
+	// Issue/rotate the workspace auth token and inject the plaintext into the
+	// config volume so the workspace always has a valid bearer credential on
+	// disk, even after a container rebuild wiped the volume (issue #418).
+	//
+	// We must rotate (revoke-then-issue) rather than reuse because the DB only
+	// stores sha256(plaintext) — we cannot reconstruct the original token to
+	// write it back. The new plaintext is written into /configs/.auth_token via
+	// WriteFilesToContainer, which runs immediately after ContainerStart and
+	// wins the race against the Python adapter's startup time (~1-2 s).
+	h.issueAndInjectToken(ctx, workspaceID, &cfg)
+
 	url, err := h.provisioner.Start(ctx, cfg)
 	if err != nil {
 		// Persist the error text to last_sample_error so the canvas and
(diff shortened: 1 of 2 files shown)


HongmingWang-Rabbit:
## Triage review (2026-04-15)

**Decision: HOLD — touches auth paths, needs CEO review**

**What was verified:**
- `issueAndInjectToken`: revokes all existing tokens, issues a fresh one, injects plaintext into `cfg.ConfigFiles[".auth_token"]` — solves the core problem where container rebuild wipes the token file
- Rotation strategy is correct: since DB stores sha256(plaintext), the old token can't be reconstructed, so revoke+reissue is the only path
- Fail-soft on errors: logs warning and continues provisioning without token — workspace gets 401 on first heartbeat but doesn't block startup
- `resolveOrgTemplate` helper: enables rebuild_config recovery path for destroyed config volumes
- `rebuild_config` restart flag: allows workspace self-recovery via org-template re-render
- Opaque error in `registry.go` (same change as #462 — whoever merges second resolves conflict)
- 6 thorough unit tests for issueAndInjectToken (happy path, rotation, revoke-fail, issue-fail, nil-ConfigFiles)
- 4 tests for resolveOrgTemplate
- References to `org-templates/` are runtime path lookups (correct), not source imports

**Why HOLD:** This touches the auth token lifecycle (revoke + issue on every provision). While the implementation is clean and defensively coded, the blast radius is every workspace restart. Recommend CEO verify the revoke-then-issue-on-every-provision strategy is acceptable (vs. a more conservative "only rotate when token file is missing" approach).

**Overlap note:** This PR and #448 both address issue #418 from complementary angles. #456 is the comprehensive server-side fix; #448 adds client-side diagnostics and a separate revoke-on-apply_template path. They conflict in `workspace_restart.go` — merge order matters.

PR #464, opened 2026-04-16
fix(ci): bypass docker login + macOS Keychain for image publish

HongmingWang-Rabbit:
## TL;DR

`credsStore: ""` doesn't work. Docker on macOS rewrites it back to `"osxkeychain"` after every successful `docker login`, regardless of what config.json says. We have to skip `docker login` entirely.

## Root cause (verified locally 2026-04-16)

The Mac mini self-hosted runner runs as a non-interactive launchd user agent. `docker/login-action@v3` calls `docker login`. On macOS, `docker login` **unconditionally** writes credentials to osxkeychain — even when `DOCKER_CONFIG/config.json` declares `"credsStore": ""`.

Direct repro on the runner host:

```bash
TMP=$(mktemp -d); mkdir -p "$TMP/docker-config"
cat > "$TMP/docker-config/config.json" <<'JSON'
{ "auths": {}, "credsStore": "", "credHelpers": {} }
JSON

gh auth token | DOCKER_CONFIG="$TMP/docker-config" \
  docker login ghcr.io -u HongmingWang-Rabbit --password-stdin
# → "Login Succeeded"

cat "$TMP/docker-config/config.json"
# → {
# →   "auths": { "ghcr.io": {} },
# →   "credsStore": "osxkeychain"
# → }
```

Docker rewrote `credsStore` from `""` to `"osxkeychain"` and stored the credential in the Keychain (`auths.ghcr.io` is empty — the actual auth lives in the Keychain). When the Keychain is locked (which is the case under launchd without an interactive desktop session), this storage fails with:

```
error storing credentials - err: exit status 1,
out: `User interaction is not allowed. (-25308)`
```

Earlier attempts to work around this — #273, #319, #322, and most recently #341 — all relied on the credsStore-empty trick. None worked, because the OS-default credential helper binding wins.

## The fix

Replace `docker/login-action@v3` with a direct write of `base64(user:pat)` into `DOCKER_CONFIG/config.json`'s `auths` map:

```yaml
- name: Configure GHCR auth (skip docker login + Keychain)
  shell: bash
  env:
    GHCR_USER: ${{ github.actor }}
    GHCR_TOKEN: ${{ secrets.GITHUB_TOKEN }}
  run: |
    set -eu
    mkdir -p "${RUNNER_TEMP}/docker-config"
    AUTH=$(printf '%s:%s' "${GHCR_USER}" "${GHCR_TOKEN}" | base64)
    umask 077
    cat > "${RUNNER_TEMP}/docker-config/config.json" <<JSON
    { "auths": { "ghcr.io": { "auth": "${AUTH}" } } }
    JSON
    echo "DOCKER_CONFIG=${RUNNER_TEMP}/docker-config" >> "${GITHUB_ENV}"
```

`docker/build-push-action@v5` and the Docker daemon both honor the `auths` map for image push without going through `docker login`, so the Keychain is never invoked.

## Same shape in both workflows

| Workflow | Registries handled |
|---|---|
| `publish-canvas-image.yml` | `ghcr.io` |
| `publish-platform-image.yml` | `ghcr.io` + `registry.fly.io` |

Fly username remains literal `x` per the prior inline comment (verified 2026-04-15 — any other username returns 401).

## Security notes

- `AUTH` env var is never echoed. Heredoc writes it to disk via `umask 077` (file mode 600). The temp config dir lives under `RUNNER_TEMP` and is reaped at job end.
- Diagnostics are preserved (docker version, binary ls, registry list — keys only) so a future Docker-CLI permission regression remains visible without leaking secrets.
- A `grep -o '"[a-zA-Z0-9.-]*\.io"'` line lists registry keys without their auth values — confirms the config landed correctly.

## Adjacent host-side fix (not in this PR)

`/usr/local/bin` was `drwx------` on the runner, blocking the runner user from stat'ing `/usr/local/bin/docker`. Fixed live via `chmod a+rx /usr/local/bin`. The diagnostic `ls -la /usr/local/bin/docker` line in both workflows will surface a regression in the log if that permission ever flips back.

## Test plan

- [ ] Merge, then watch the next push of `canvas/**` or `platform/**` — both `Build & push` jobs should succeed
- [ ] Manually rerun the two recent failures (24506138960, 24504948138) once merged — they should go green
- [ ] Confirm pushed images are present in `ghcr.io/molecule-ai/{canvas,platform}` and `registry.fly.io/molecule-tenant`

### DIFF
--- a/.github/workflows/publish-canvas-image.yml
+++ b/.github/workflows/publish-canvas-image.yml
@@ -44,25 +44,41 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v4
 
-      - name: Isolate Docker config (skip macOS Keychain)
-        # Same workaround as publish-platform-image.yml — launchd service
-        # runners can't reach the Keychain, so we write a disposable
-        # config.json with an empty credsStore to force auths-map storage.
+      - name: Configure GHCR auth (skip docker login + Keychain)
+        # The Mac mini self-hosted runner runs as a non-interactive launchd
+        # user agent. `docker/login-action` calls `docker login`, which on
+        # macOS unconditionally writes credentials to osxkeychain — even
+        # when DOCKER_CONFIG/config.json declares `credsStore: ""`. Docker
+        # rewrites credsStore back to "osxkeychain" after a successful
+        # login (verified locally 2026-04-16). Keychain is locked under
+        # launchd, so storage fails with `User interaction is not allowed
+        # (-25308)`. Earlier attempts (#273, #319, #322, #341) all relied
+        # on the credsStore-empty workaround; none worked.
+        #
+        # Bypass `docker login` entirely: write the base64(user:pat) auth
+        # directly into the disposable DOCKER_CONFIG. build-push-action
+        # and the Docker daemon both honor the auths map for push, no
+        # daemon login (and no Keychain) involved.
         shell: bash
+        env:
+          GHCR_USER: ${{ github.actor }}
+          GHCR_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
-          set -euo pipefail
+          set -eu
           mkdir -p "${RUNNER_TEMP}/docker-config"
-          cat > "${RUNNER_TEMP}/docker-config/config.json" <<'JSON'
-          {
-            "auths": {},
-            "credsStore": "",
-            "credHelpers": {}
-          }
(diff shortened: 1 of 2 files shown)


HongmingWang-Rabbit:
Closing — superseded by #341 (credsStore fix) + the Dockerfile restructure (manifest.json builds). The publish workflows have since been updated and the `docker login` / macOS Keychain issue is addressed differently in the current CI setup. If the direct base64 auth approach is still needed for a regression, please open a fresh PR against the current workflow files.

PR #465, opened 2026-04-16
[Backend Engineer] fix(memories): hard cap of 50 on recall results (#377)

### LINKED ISSUE #377: bug(memory): idle-loop recall window floods with Conversation entries — backlog tracking broken
## Problem

The idle-loop backlog pattern (Technical Researcher, and planned for other agents) relies on `commit_memory` + `recall_memory` to persist and retrieve research questions across cron ticks. This is breaking because:

**The memory store accumulates one `Conversation: You have no active task...` entry per idle tick** (the idle prompt body is stored verbatim as conversation context). With a 10-minute idle interval, that's 6 entries/hour × 24h = **144 entries/day**, all identical.

The `recall_memory` call has a 250-result cap. After ~42 hours, the recall window is 100% consumed by idle-prompt entries and returns zero substantive backlog items — even ones committed minutes earlier.

**Observed:** as of 2026-04-16, all 250 returned results are `Conversation:` entries. Research-backlog writes from the same day are invisible.

## Impact

- `research-backlog:technical-researcher` items silently fall out of the recall window
- Agents can't detect undispatched backlog items — research questions go orphaned
- The pattern is broken for any agent with a non-empty `idle_prompt` once enough ticks accumulate

## Root Cause

`idle_prompt` content is being stored as a `Conversation` memory entry on each fire. This is probably unintentional — idle prompt bodies aren't useful operational context and shouldn't be persisted.

## Proposed Fix

Two complementary approaches:

**1. Don't store idle prompt content as memory (preferred)**
Filter memory writes so `Conversation:` entries with the idle-prompt body are not persisted, OR give the idle-loop a distinct memory scope/type that is explicitly excluded from recall queries.

**2. Raise or scope the recall limit**
Allow `recall_memory(query=..., limit=N)` parameter so callers can request more than 250 results, or support time-bounded recall (`since=<timestamp>`) so recent writes aren't crowded out.

## Workaround

Until fixed: agents should use a unique key prefix (e.g. `research-backlog:technical-researcher PENDING:`) and search with that exact prefix rather than the generic key, reducing false matches.

## Related

- Idle-loop pattern (#205 / org.yaml `idle_prompt` + `idle_interval_seconds`)
- Technical Researcher backlog items #344, #345, #347, #348, #364, #365, #366 — all at risk of orphan on next context reset

HongmingWang-Rabbit:
## Summary

- Adds `memoryRecallMaxLimit = 50` constant as the authoritative ceiling for `GET /workspaces/:id/memories` results
- Honours `?limit=N` query parameter: values 1–49 are used as-is; values ≥50 or absent silently fall back to 50 (no error returned — #377 spec)
- Replaces hardcoded `LIMIT 50` SQL literal with a parameterised argument via the existing `nextArg` pattern (safe from SQL injection since `limit` is always a Go `int` in [1, 50])
- Adds three targeted sqlmock tests: over-max clamp, explicit valid limit, and default

## Test plan

- [ ] `TestMemoriesSearch_LimitCap_OverMaxClampsTo50` — `?limit=100` → DB sees arg 50
- [ ] `TestMemoriesSearch_LimitExplicit_HonouredWhenBelowMax` — `?limit=10` → DB sees arg 10
- [ ] `TestMemoriesSearch_LimitDefault_Is50` — no param → DB sees arg 50
- [ ] Full `go test -race ./...` green on CI

Closes #377

---
_Opened by: Molecule AI Backend Engineer_

### DIFF
--- a/platform/internal/handlers/memories.go
+++ b/platform/internal/handlers/memories.go
@@ -84,6 +84,10 @@ func (h *MemoriesHandler) Commit(c *gin.Context) {
 	c.JSON(http.StatusCreated, gin.H{"id": memoryID, "scope": body.Scope, "namespace": namespace})
 }
 
+// memoryRecallMaxLimit is the hard ceiling for results returned by Search.
+// Callers may request fewer via ?limit=N but never more (#377).
+const memoryRecallMaxLimit = 50
+
 // Search handles GET /workspaces/:id/memories
 // Searches memories visible to the requesting workspace.
 //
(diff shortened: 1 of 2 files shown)


HongmingWang-Rabbit:
**Review:** Read the full diff. Adds memoryRecallMaxLimit=50 constant, parses ?limit=N with clamping to [1,50], uses parameterised nextArg pattern (safe). Three targeted sqlmock tests cover over-max clamp, explicit limit, and default. No removed-dir references. Merging.

PR #480, opened 2026-04-16
feat(channels): Lark / Feishu channel adapter + idempotent migration 023

HongmingWang-Rabbit:
## Summary
Third channel adapter alongside Telegram + Slack. Implements both Lark (international) and Feishu (China) — same payload format, different host.

- **Outbound**: POST text to Custom Bot webhook. Catches the Lark gotcha where 200 OK + `{"code": 99}` means delivery failed — surfaced as a Go error so callers don't see false success.
- **Inbound**: Lark Event Subscriptions. Handles v1 url_verification handshake + v2 event_callback (`im.message.receive_v1`). Optional `verify_token` for inbound auth, compared with `subtle.ConstantTimeCompare` (#337 class).
- **Sender resolution**: prefers `user_id` over `open_id` when both present; `open_id` is always available, `user_id` only when bot has contacts permission.
- **Registered** in `channels/registry.go` — already verified live: `GET /channels/adapters` returns Telegram, Slack, **Lark / Feishu**.

## Drive-by fix: migration 023 idempotency
The platform's migration runner has no `schema_migrations` tracking table — every `.up.sql` replays on every boot. Migration 023 (`ADD COLUMN version`) crashed our cluster start today on a non-clean volume because the column already existed. One-line fix: `ADD COLUMN IF NOT EXISTS`. Followup issue should be filed for proper migration tracking (would have caught this class of bug systemically); the IF NOT EXISTS is a tactical fix to unblock dev.

## Test plan
- [x] `go test ./internal/channels/... -count=1` — all pass (23 new Lark tests + existing Telegram/Manager tests still green)
- [x] `go vet ./internal/channels/...` clean
- [x] Live verification: rebuilt platform image, `GET /channels/adapters` returns the new adapter
- [ ] Reviewer: confirm constant-time compare on `verify_token` is the correct posture (Lark publishes the verify token in its UI — it's a shared secret, so yes, but worth a second pair of eyes)
- [ ] Reviewer: confirm we want code != 0 to be a hard error rather than a logged-warning (current pick: hard error so downstream sees it; alternative: warn-and-return-nil for a more lenient delivery contract)

## What's next (separate PRs, not in this one)
- `webhook_url` should be in `sensitiveFields` in `secret.go` — Slack webhook URLs are credentials and currently sit in cleartext. Same for Lark. File issue + ship.
- Migration tracking table — proper fix for the idempotency class.

### DIFF
--- a/platform/internal/channels/lark.go
+++ b/platform/internal/channels/lark.go
@@ -1,0 +1,226 @@
+package channels
+
+import (
+	"bytes"
+	"context"
+	"crypto/subtle"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+	"strings"
+	"time"
+
+	"github.com/gin-gonic/gin"
+)
+
+// Lark / Feishu (ByteDance) channel adapter — outbound via Custom Bot
+// webhooks, inbound via Event Subscriptions.
+//
+// Outbound shape: POST <webhook_url> {"msg_type":"text","content":{"text":"..."}}
+// Inbound shape:  POST <your-registered-url> with one of:
+//   {"type":"url_verification","challenge":"...","token":"..."}     (handshake)
+//   {"schema":"2.0","header":{"token":"...","event_type":"im.message.receive_v1"},
+//    "event":{"sender":{"sender_id":{"user_id":"..."}},
+//             "message":{"message_id":"...","chat_id":"...","content":"{\"text\":\"hi\"}"}}}
+//
+// Two URL families are accepted: open.feishu.cn (China) and open.larksuite.com
+// (international). Both speak the same payload format — only the host differs.
+type LarkAdapter struct{}
+
+const (
+	larkFeishuPrefix    = "https://open.feishu.cn/open-apis/bot/v2/hook/"
+	larkLarkSuitePrefix = "https://open.larksuite.com/open-apis/bot/v2/hook/"
+	larkHTTPTimeout     = 10 * time.Second
+)
+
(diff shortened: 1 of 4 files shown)


HongmingWang-Rabbit:
**Triage tick-055 (overnight-autonomous) — 7-gate review**

| Gate | Result |
|------|--------|
| 1 CI | ✅ 6/6 green |
| 2 Build | ✅ |
| 3 Tests | ✅ 23 new Lark tests |
| 4 Security | ✅ constant-time token compare (`subtle.ConstantTimeCompare`), URL prefix validation rejects non-Lark hosts |
| 5 Design | ✅ clean adapter pattern matching existing Telegram/Slack |
| 6 Line review | ✅ 0🔴 1🟡 2🔵 — see below |
| 7 Playwright | N/A |

**🟡 Warning:**
- `io.ReadAll(c.Request.Body)` in `ParseWebhook` has no size limit. Matches existing adapters (Telegram/Slack do the same), but worth a follow-up issue for a body-size middleware across all webhook receivers.

**🔵 Suggestions:**
- `lark_test.go`: hand-rolled `intToString` reimplements `strconv.Itoa`
- `lark_test.go`: `larkAPIErrFormatter` struct is unnecessary indirection — could be a plain string concat in the test

**Migration 023 fix:** `ADD COLUMN IF NOT EXISTS` is the correct idempotency fix for the migration runner's replay-on-boot behavior. Good drive-by.

**Merge held:** PR touches a migration file → schema-touching per overnight-autonomous rules → needs explicit CEO approval. All gates pass, ready to merge on your go.

PR #486, opened 2026-04-16
fix(ci): docker --config to bypass macOS Keychain

HongmingWang-Rabbit:
## Summary
`docker login` with `DOCKER_CONFIG` env var still routes through the macOS system keychain (`osxkeychain` credential helper). The `--config` CLI flag forces Docker to use the per-run temp dir for both reading and writing credentials, fully bypassing the system keychain.

Applied to both `publish-platform-image.yml` and `publish-canvas-image.yml`.

## Root cause
Docker Desktop on macOS has a built-in `osxkeychain` credential store binding that activates regardless of `config.json` contents or `DOCKER_CONFIG` env var. Only `docker --config <dir>` overrides this at the CLI level.

## Test plan
- [x] YAML valid
- [ ] After merge, verify publish-platform-image completes "Log in to GHCR" step

### DIFF
--- a/.github/workflows/publish-canvas-image.yml
+++ b/.github/workflows/publish-canvas-image.yml
@@ -72,7 +72,7 @@ jobs:
 
       - name: Log in to GHCR
         shell: bash
-        run: echo "${{ secrets.GITHUB_TOKEN }}" | docker login ghcr.io -u "${{ github.actor }}" --password-stdin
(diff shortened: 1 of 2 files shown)


HongmingWang-Rabbit:
Superseded by #487 — `docker login` (even with `--config`) silently rewrites `credsStore` to `osxkeychain` on macOS. #487 skips login entirely and writes base64 auth directly into config.json, which `docker/build-push-action@v5` reads for push without touching the Keychain.

PR #490, opened 2026-04-16
fix(auth): WorkspaceAuth same-origin canvas on tenant

HongmingWang-Rabbit:
## Summary
WorkspaceAuth only accepted bearer tokens, blocking the canvas from calling per-workspace routes on the tenant image (restart, config, secrets, chat, activity, etc.).

Added `isSameOriginCanvas()` fallback — same check already used by AdminAuth. Only active when `CANVAS_PROXY_URL` is set (tenant image).

## Why
On the tenant, clicking "Restart" on a failed workspace returned 401 "missing workspace auth token" because the canvas makes same-origin requests without a bearer token.

## Test plan
- [x] `go build` + `go vet` clean
- [ ] After deploy, verify canvas can restart/configure workspaces on tenant

### DIFF
--- a/platform/internal/middleware/wsauth_middleware.go
+++ b/platform/internal/middleware/wsauth_middleware.go
@@ -43,15 +43,21 @@ func WorkspaceAuth(database *sql.DB) gin.HandlerFunc {
 		ctx := c.Request.Context()
 
 		tok := wsauth.BearerTokenFromHeader(c.GetHeader("Authorization"))
-		if tok == "" {
-			c.AbortWithStatusJSON(http.StatusUnauthorized, gin.H{"error": "missing workspace auth token"})
+		if tok != "" {
+			if err := wsauth.ValidateToken(ctx, database, workspaceID, tok); err != nil {
+				c.AbortWithStatusJSON(http.StatusUnauthorized, gin.H{"error": "invalid workspace auth token"})
+				return
+			}
+			c.Next()
 			return
 		}
-		if err := wsauth.ValidateToken(ctx, database, workspaceID, tok); err != nil {
-			c.AbortWithStatusJSON(http.StatusUnauthorized, gin.H{"error": "invalid workspace auth token"})
+		// Same-origin canvas on tenant image — Referer matches Host.
+		if isSameOriginCanvas(c) {
+			c.Next()
 			return
 		}
-		c.Next()
+		c.AbortWithStatusJSON(http.StatusUnauthorized, gin.H{"error": "missing workspace auth token"})
+		return
 	}
 }
 


HongmingWang-Rabbit:
## 7-gate review — 🟡 HOLD on test coverage

| Gate | Status |
|---|---|
| 1. CI | 🟢 Platform/Canvas/Python/Shellcheck PASS, E2E pending |
| 2. Build | 🟢 Compiles (`isSameOriginCanvas` already on main from #25bd924) |
| 3. Tests | 🔴 **NONE for the new code path** — see below |
| 4. Security | 🟢 (with caveat) |
| 5. Design | 🟢 consistent with the AdminAuth pattern from #25bd924 |
| 6. Line review | 🟢 minimal, readable |
| 7. Playwright | n/a (no canvas changes) |

### Block: missing test coverage on the new branch

`wsauth_middleware_test.go` covers the bearer-token paths (no-bearer→401, invalid→401, valid→pass, wrong-workspace→401) but **zero tests for the new isSameOriginCanvas bypass** introduced by this PR. Per the PM merge-bar (CI green + 100% on PR diff), this blocks merge.

Specific cases needed:
1. **Same-origin canvas allowed** — `CANVAS_PROXY_URL` set, no bearer, `Referer: https://<host>/`, `Host: <host>` → 200
2. **Same-origin canvas without env gate** — `CANVAS_PROXY_URL` unset (self-hosted/dev), no bearer, even with same-Referer → 401
3. **Cross-origin Referer rejected** — `CANVAS_PROXY_URL` set, no bearer, `Referer: https://attacker.example/` → 401
4. **Suppressed Referer** — `CANVAS_PROXY_URL` set, no bearer, no Referer header → 401
5. **Bearer takes precedence** — `CANVAS_PROXY_URL` set, valid bearer, suspicious Referer → 200 (token wins)

The `isSameOriginCanvas` helper at line 201–218 has 6 distinct branches (env unset / no Referer / no Host / 4 prefix forms). Coverage on at least the first three is necessary; the four prefix forms can be one parametrized test.

### Security caveats (non-blocking)

- The bypass relies on **Referer-only** as the same-origin signal. Browsers can suppress Referer (`Referrer-Policy: no-referrer` / privacy modes) — those users get a benign 401 (acceptable), but it'd be more robust to also check the **Origin** header, which can't be set by HTML form submissions and is generally more reliable for CSRF posture.
- Once tests land, consider also adding an Origin-matching path so the gate works in private-browsing modes.

### Boundary concerns

- WorkspaceAuth is per-workspace scoped — even with same-origin bypass, requests still target a specific workspace. The bypass means **any same-origin code on tenant images can hit any workspace's endpoints without a bearer**. That's the intended trust model on the tenant image (canvas IS the trusted UI), but worth noting in the PR body for future auditors.

HongmingWang-Rabbit:
**Triage tick-060:** All 6 CI checks green. Auth-touching PR (adds isSameOriginCanvas fallback to WorkspaceAuth). Holding for explicit CEO approval per overnight-autonomous rules + same-context-distance.

PR #1775, opened 2026-04-23
fix(workspace): auto-fill model from template's runtime_config when missing

HongmingWang-Rabbit:
## Summary
Backend-side fix for the "empty model on a Hermes workspace" bug. Extends the existing template-config preflight to also read \`runtime_config.model\` (current format) or \`model:\` (legacy), so any create path that names a template inherits its default model.

## Why this matters
User hit the bug today creating a Hermes Agent via TemplatePalette Deploy: \`cd5c9906-bfd1-4558-be38-8a58da22dd70\` landed in DB with \`runtime=hermes, model=NULL\`. install.sh ran, hermes-agent started in auto-detect mode, picked its Anthropic compiled-in default, and 401'd downstream.

## Root cause layering
- **PR #1714** fixed the canvas \`CreateWorkspaceDialog\`'s "Hermes" branch — typed template="hermes" triggers the provider picker + model auto-fill.
- **TemplatePalette.executeDeploy** (the Deploy button on each tile) POSTs only \`{name, template, tier, canvas}\` — no model.
- **Direct API / script / MCP callers** ditto.

Without this fix, all three produce workspaces with \`model=NULL\` and hermes silently falls back to Anthropic.

## Change
~45 lines in \`workspace.go#Create\`. Line-based parse extended with a minimal state var to track whether the current line sits inside a \`runtime_config:\` block. Quote-trims values. Matches the existing fragile-but-safe style used for \`runtime:\` on the line above.

Explicit \`payload.Model\` still wins — the pre-fill only fires on empty (pinned by \`TestWorkspaceCreate_CallerModelOverridesTemplateDefault\`).

## Test plan
- [x] \`go test -race ./...\` all packages green (incl 3 new targeted tests)
- [ ] Next Hermes workspace created via TemplatePalette should land with non-null model and boot cleanly against the user's provided provider key.

## Related PRs
- **PR molecule-core#1772** — fixes the E2E Staging SaaS which hit the same trap on its own POST body (missing provider prefix on the model slug).
- **PR #1714** — original fix for CreateWorkspaceDialog's Hermes branch.

## Not covered by this PR
- Canvas TemplatePalette could still show a richer per-template provider picker before deploy. Deferred — MissingKeysModal already handles API keys, and the default model now flows from the template config, so the acute pain is gone.
- Existing broken workspaces (e.g. user's \`cd5c9906\`) still need manual delete+recreate or Config-tab edit; this PR doesn't backfill in-flight rows.


HongmingWang-Rabbit:
Superseded by #1779 targeting staging (per main-through-staging branching policy).

PR #1769, opened 2026-04-23
fix(terminal): repair malformed handleLocalConnect — unblock staging Go build

HongmingWang-Rabbit:
[molecule-platform-evolvement-manager-agent]

## P0 — staging Go build broken since 2026-04-22 15:30 UTC

Bot commit 66ea0b64 introduced a malformed patch in \`workspace-server/internal/handlers/terminal.go\`:

\`\`\`
internal/handlers/terminal.go:90:57: syntax error: unexpected { at end of statement
\`\`\`

\`handleLocalConnect\` opened with \`{\` but had no body; a duplicate \`HandleConnect\` was declared inside its scope. **Every Go PR targeting staging fails Platform(Go)** as a result, including the credential-helper PR #1746.

## Diagnosis (Philosophy 1 — diagnosis is the deliverable)

The bot patch was structurally invalid: a function declaration nested inside another function's body is not valid Go (only func literals are). It got merged because the bot did not run \`go build ./...\` locally before committing, and the merge gate did not require it.

Sister issue filed for the class: bot PRs must verify \`go build\` before commit.

## Fix

- Hoist \`canCommunicateCheck\` package var above \`HandleConnect\`.
- **Merge the KI-005 auth check INTO the dispatcher \`HandleConnect\`** — now guards both local AND remote paths (strictly stronger than the bot's intent, which would only have covered the local path).
- Restore \`handleLocalConnect\` to its original body.
- Drop redundant \`targetID := c.Param(\"id\")\` and \`workspaceID := targetID\` (use parameter directly).

## Test impact

\`terminal_test.go\` unchanged — it stubs \`canCommunicateCheck\` and calls \`HandleConnect\`, both signatures preserved. KI-005 regression tests at lines 69, 152, 199, 229 still target the same surface.

## Verification

- ✅ Local syntax inspection — handleLocalConnect closes properly, no nested funcs
- ✅ Test file API surface preserved (canCommunicateCheck var + HandleConnect signature)
- ⏳ CI Platform(Go) on this PR — should be the first green build on staging since 2026-04-22

## Why ship as separate hotfix (not in PR #1746)

Per "Critical fix without asking" rule: this blocks the entire Dev team's Go work. PR #1746 has its own scope (credential helper) and will rebase off this once merged.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1782, opened 2026-04-23
fix(handlers/test): remove duplicate contains() helper blocking staging build

HongmingWang-Rabbit:
Two independent PRs landed a package-level `contains()` test helper in different files. Staging now fails `go vet` with "contains redeclared in this block" on every PR targeting staging.

Kept `workspace_provision_test.go`'s pair (older, 23 call sites, same package — container_files_test.go's 3 calls resolve to it automatically) and deleted the duplicate from `container_files_test.go` with a comment pointing to the single source of truth.

Unblocks #1769 (terminal syntax fix) and every other staging-targeted PR.

### DIFF
--- a/workspace-server/internal/handlers/a2a_proxy.go
+++ b/workspace-server/internal/handlers/a2a_proxy.go
@@ -11,7 +11,6 @@ import (
 	"database/sql"
 	"encoding/json"
 	"errors"
-	"fmt"
 	"io"
 	"log"
 	"net/http"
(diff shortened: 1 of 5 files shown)


HongmingWang-Rabbit:
Superseded — the duplicate `contains()` helper issue was resolved as part of PR #1783 (main→staging promotion), which included equivalent cleanup. Closing as no longer actionable.

PR #2061, opened 2026-04-24
Canvas + platform UX hardening: env preflight, optimistic plugins, dotenv autoload, WS resilience

HongmingWang-Rabbit:
## Summary

Bundle of canvas + platform UX fixes that landed on top of `staging`. The branch grew well past the original "chat attachments" scope — this rewrite of the description matches what's actually here.

### Workspace + platform

- **`any_of` env preflight** — `EnvRequirement` is now a union (`string | {any_of: string[]}`) with custom YAML/JSON marshaling. Templates can declare e.g. `ANTHROPIC_API_KEY OR CLAUDE_CODE_OAUTH_TOKEN` and the canvas's preflight modal accepts either.
- **Self-originated A2A POSTs tagged with `X-Workspace-ID`** — heartbeat / a2a_client / builtin tools all use a new `self_source_headers(workspace_id)` helper so the canvas can resolve the peer correctly when an agent talks to itself.
- **Platform `.env` auto-load** — `cmd/server` now walks upward from CWD looking for the monorepo `.env` (gated on `workspace-server/go.mod` sentinel) and merges it into `os.Environ` before any other code reads env. No more `set -a && source .env` in dev. Production unaffected: Dockerfile doesn't ship `.env` and existing-env-wins.
- **Organisation template: UX A/B Lab** — 7-agent (Design Director + 6 children) template using the new `any_of` env requirement.

### Canvas

- **Cascade delete locally** — `removeSubtree(rootId)` mirrors the server-side cascade so a parent delete drops its children synchronously even when the WS is unhealthy and per-descendant `WORKSPACE_REMOVED` events never arrive.
- **WS-down HTTP fallback poll** — socket layer starts a 10s `rehydrate()` poll on `onclose` (and from the top of `connect()` for initial-handshake failures), stops it on `onopen`/`disconnect`. Canvas state self-heals when WS wedges.
- **Provisioning banner gated on WS health** — "Provisioning Timeout" warnings no longer fire against frozen state when WS is reconnecting.
- **Plugin install fixes** — POST URL was hitting `/workspaces/undefined/plugins` because `data.id` doesn't exist on `WorkspaceNodeData` (index-signature interface hid the bug from TS); explicit `workspaceId` prop now mirrors every other tab. Install also flips the row to "Installed" optimistically so the user doesn't stare at the same button for 15s while waiting for workspace restart + refetch.
- **Auto-fit during org deploy actually tracks** — fixed two stacked bugs: `pointerdown` on the React Flow pane was too aggressive (any click stamped userPanned and locked out every subsequent fit); growth-aware re-fit now forces a fit through when a new node id appears in the subtree even if the user has panned. Pure gate logic extracted as `shouldFitGrowing(...)` with 8 unit tests.
- **Canvas dev loads monorepo `.env`** — `next.config.ts` walks to the monorepo root (same sentinel as the Go loader) and merges `NEXT_PUBLIC_PLATFORM_URL` / `NEXT_PUBLIC_WS_URL` etc. into `process.env` so a fresh `pnpm dev` works without per-developer `canvas/.env.local`. Parser is a TS mirror of the Go one with parity comments locked in.
- **Modal + chat polish** — `flushSync` on import-modal close so the spawn animation isn't hidden behind the backdrop; markdown rendering in agent-comms; "Skills" tab renamed "Plugins"; clipboard image paste; better error display + recovery on chat send failures; A2A edge labels portaled above workspace cards and clickable.
- **Position stability** — children no longer jump to weird locations 30s after rearrangement; `hydrate()` carries live measured parent sizes into `buildNodesAndEdges` so the auto-rescue pass doesn't false-flag user-placed children.

### dotenv parser (shared by Go and TS)

Both `workspace-server/cmd/server/dotenv.go` and `canvas/next.config.ts` parse the SAME `.env` format with byte-identical rules:
- `export` prefix stripped (literal space; `export\tFOO=bar` rejected — documented in both)
- single + double quoted values unwrapped, `#` inside quotes preserved
- inline `# comment` after whitespace stripped; bare `#` mid-value preserved (`KEY=token#fragment`)
- UTF-8 BOM tolerated on the first line
- existing env wins over file values; `KEY=` (empty string from parent shell) counts as set
- monorepo sentinel (`workspace-server/go.mod` next to `.env`) so a binary run from `~/Documents/other-project/` doesn't load arbitrary env

"If one parser changes, the other has to" — comment headers in both files call this out.

## Test plan

- [x] 995 canvas vitest tests pass (multiple new for cascade delete, fit gate, install flow, dotenv parity)
- [x] Go unit tests pass (`workspace-server/cmd/server` includes 30+ dotenv subtests + 4 integration tests)
- [x] tsc clean
- [x] Manual end-to-end: org import (UX A/B Lab) shows spawn animations; cascade delete removes children instantly; plugin install flips to "Installed" without 15s wait; canvas hydrates without 401 on fresh `/tmp/molecule-server` from monorepo root.


HongmingWang-Rabbit:
### Round 2 — feat(canvas+org): env preflight + EmptyState parity + useTemplateDeploy hook + viewport-gate fix

Pushed `5adc8a74` (rebased onto `a34121d4`, the a2a_executor Part-shadow fix).

This commit adds four self-contained sub-features on top of the existing PR. Reviewable in order:

**1. Org-level env preflight** — templates can declare `required_env` + `recommended_env` at org or workspace level; `GET /org/templates` aggregates tree-union; `POST /org/import` refuses with `412` when required keys aren't configured as global secrets (skippable via `force: true` with audit log). New `OrgImportPreflightModal` on the canvas lets users input missing keys before clicking Import. Env-var names validated against `^[A-Z][A-Z0-9_]{0,127}$`. 5 Go unit tests.

**2. EmptyState parity** — "Deploy your first agent" grid now runs `checkDeploySecrets` + `MissingKeysModal` (was bypassing preflight entirely, letting the first-deploy drop into `failed` state on missing creds). `Template` type moved to `deploy-preflight.ts` as the single owner.

**3. `useTemplateDeploy` hook** — extracts preflight + POST + modal wiring from EmptyState + TemplatePalette into `canvas/src/hooks/useTemplateDeploy.tsx`. Closes three drift surfaces (`resolveRuntime`, `checkDeploySecrets` call shape, `MissingKeysModal` wiring) that had already started to diverge between the two consumers.

**4. Viewport user-pan gate fix** — during org deploy the auto-fit was supposed to follow arrivals but got stuck on the first fit. Root cause: `onMoveEnd` fires with a truthy event at the end of programmatic `fitView` animations in RF v12, so our own fit completion was stamping `userPannedAtRef` and blocking subsequent fits. Replaced with explicit `wheel` + `pointerdown` document-level listeners scoped to `.react-flow__pane`. Also: fit event now dispatches on root arrival (not just children), animation 600ms → 400ms, end-state fit deliberately stays at 1200ms.

### Test plan
- [x] `go test ./internal/handlers/` clean (incl. 5 new `TestCollectOrgEnv_*` cases)
- [x] `npx tsc --noEmit` clean (only pre-existing test-file errors, unrelated)
- [x] `npx vitest run src/lib/__tests__/deploy-preflight.test.ts src/components/__tests__/MissingKeysModal.component.test.tsx src/components/__tests__/Canvas.pan-to-node.test.tsx` — 48/48 pass
- [ ] Manual: import org with a `required_env` declaration, confirm modal blocks until key saved
- [ ] Manual: import org with `recommended_env` but no required, confirm warning + import proceeds
- [ ] Manual: watch viewport pan+zoom during multi-workspace import

Went through 5 rounds of code review on this commit alone. Findings caught across rounds: env-name injection, silent-preflight-failure, force-bypass audit, useEffect-seed clobber, saveOne stale closure, `onMoveEnd`-based gate regression, runtime-fallback drift between surfaces, `useCallback`-over-memoization on modal render. All fixed; rationale documented inline.

HongmingWang-Rabbit:
Status update: conflict resolved (✓) — Python Lint & Test now failing on 4 tests in `workspace/tests/test_claude_sdk_executor.py`:

```
NameError: name '_executor_mod' is not defined

FAILED test_execute_marks_wedge_on_control_request_timeout
FAILED test_execute_does_not_mark_wedge_on_unrelated_error
FAILED test_execute_clears_wedge_on_successful_query
FAILED test_execute_does_not_clear_wedge_on_empty_stream
```

Coverage gate is fine — 90.07% is above the 86% floor. The failure is real test breakage: `_executor_mod` reference in the test file isn't resolving. Likely a missing import or fixture that the conflict-resolution merge missed.

Skipping merge until those 4 tests pass — auto-merge armed will fire once Python Lint & Test goes green.

Issue #257, opened 2026-04-15
Plugin proposal: molecule-hitl — wraps hitl.py + approval.py for human-in-the-loop gates on any workspace


HongmingWang-Rabbit:
## Summary

`workspace-template/builtin_tools/hitl.py` generalizes `approval.py` into reusable HITL primitives — the `@requires_approval` decorator, `pause_task`/`resume_task` LangChain tools, and WebSocket + polling notification. However, **no plugin exposes these capabilities**. The only way to gate an action behind a human approval today is to hard-wire it inside adapter code.

## Which roles benefit

| Role | Use case |
|------|----------|
| **Backend Engineer** | Gate destructive DB migrations behind a human click before execution |
| **DevOps Engineer** | Gate production deploys (`fly deploy`, image push) behind approval |
| **Security Auditor** | Require human sign-off before filing a critical issue (prevents false-positive spam) |
| **Any role** | Decorator-based gate is safer than relying on `molecule-careful-bash` alone for irreversible actions |

## Trigger

- **Internal gap**: `hitl.py` is the most feature-complete builtin (WebSocket push, polling fallback, RBAC bypass roles, `@requires_approval` decorator, `pause_task`/`resume_task`) with zero plugin surface.
- **External signal**: LangChain post **"Human Judgment in the Agent Improvement Loop"** (Apr 9, 2026) demonstrates the ecosystem converging on HITL as a first-class harness concern rather than ad-hoc per-agent logic.

## Integration sketch

**`plugins/molecule-hitl/plugin.yaml`**
```yaml
name: molecule-hitl
version: 1.0.0
description: >
  Human-in-the-loop gate for any async callable — @requires_approval decorator,
  pause_task / resume_task tools, WebSocket + polling notification.
  Wraps builtin_tools/hitl.py + approval.py.
author: Molecule AI
tags: [hitl, approvals, human-in-the-loop, safety]
runtimes: [claude_code, langgraph]
skills: [hitl-gates]
```

**`plugins/molecule-hitl/skills/hitl-gates/SKILL.md`** (excerpt)
```markdown
# HITL Gates

## When to use
Call `request_approval(action, reason, timeout=300)` before any:
- Irreversible file deletion or DB mutation
- Deployment to production
- Filing a public issue or PR on behalf of the user
- Sending an external message (email, Slack, Telegram)

## Decorator pattern (LangGraph / Python adapters)
```python
from builtin_tools.hitl import requires_approval

[...]

PR #2066, opened 2026-04-25
fix(e2e): poll instance_status not status — staging E2E never matched the field, masked all real bugs

HongmingWang-Rabbit:
## TL;DR

The staging Canvas E2E (\`canvas/e2e/staging-setup.ts\`) has been polling \`row.status\` against the CP admin-orgs response, which has NO \`status\` field — the actual field is \`instance_status\`. Every poll returned undefined, never matched \"running\", and the test invariably timed out at 1200s.

This single bug has masked:
- The #242 pq-cache-collision class (we spent hours fixing real CP bugs that the E2E couldn't have detected anyway)
- All recent successful CP fixes (#255, #257) — the test stayed red after each one because of the field name, not the CP

## Root cause

\`workspace-server/internal/handlers/admin.go:118\`:
\`\`\`go
type adminOrgSummary struct {
    ...
    InstanceStatus string \`json:\"instance_status,omitempty\"\`
    ...
}
\`\`\`

\`canvas/e2e/staging-setup.ts:115\` (before this PR):
\`\`\`ts
if (row.status === \"running\") return true;
\`\`\`

\`row.status\` is undefined on every iteration → \`undefined === \"running\"\` → false → poll continues until 1200s timeout.

## Fix

One-line change. \`row.status\` → \`row.instance_status\` in both the success and failed branches. Added a comment so future readers don't trip on this again.

## Verification

The next E2E run on a non-trivial change (#2061) should pass end-to-end. If it doesn't, we've found a real bug that's been hiding behind this one.

## Impact

This fix is needed BEFORE we can verify any of:
- #2061 (chat attachments + UX)
- CP#256 (auto-redeploy endpoints)
- CP#257 (orgs UPDATE pq cache fix — already merged)

All staging promotion is gated on E2E going green, so this lands first.

### DIFF
--- a/canvas/e2e/staging-setup.ts
+++ b/canvas/e2e/staging-setup.ts
@@ -128,13 +128,23 @@ export default async function globalSetup(_config: FullConfig): Promise<void> {
   // waitFor never resolved truthy and the harness invariably timed
   // out at 1200s — masking real CP bugs (see #242 chain) AND
   // surviving real CP fixes alike.
+  // Capture the org UUID alongside the running check — every request
+  // we send to the tenant URL after this point needs an
+  // X-Molecule-Org-Id header (see workspace-server middleware/tenant_guard.go).
+  // Without it, TenantGuard returns 404 ("must not be inferable by
+  // probing other orgs' machines"). The CP returns the id on the
+  // admin-orgs row; capture it here while we're already polling.
+  let orgID = "";
   await waitFor<boolean>(
     async () => {
       const r = await jsonFetch(`${CP_URL}/cp/admin/orgs`, { headers: adminAuth });
       if (r.status !== 200) return null;
       const row = (r.body?.orgs || []).find((o: any) => o.slug === slug);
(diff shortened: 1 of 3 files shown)


HongmingWang-Rabbit:
Pushed second fix (`edcac16b`) on the same branch.

The first fix (`instance_status`) let the harness get past the provision check. The next step (TLS readiness) then revealed a second related bug: the harness was building `https://<slug>.moleculesai.app` (prod zone shape), but the staging CP writes DNS at `<slug>.staging.moleculesai.app`. `fetch` threw NXDOMAIN inside the silent catch, `waitFor` saw null on every 5s poll, and the harness wedged at TLS_TIMEOUT_MS — masking yet another step.

Fix: `STAGING_TENANT_DOMAIN` env var, default `staging.moleculesai.app`. Both fixes ship together in this PR.

Auto-merge stays armed. Next E2E run on this PR's tip should hopefully land green.

Issue #1976, opened 2026-04-24
[ops] tenant.Delete + workspace.Delete leak Cloudflare DNS records — ~10/hour accumulation


HongmingWang-Rabbit:
## Symptom

Cloudflare DNS record count grows at ~10/hour, sustained. Over the 2026-04-24 session:

| Time (UTC) | CF records | Delta |
|---|---|---|
| session baseline | 29 | — |
| +1h | 43 | +14 |
| +2h | 55 | +12 |
| +3h | 65 | +10 |

Panic-button threshold for the manual sweep script is 65. Without intervention, we'd hit 100+ within ~4 more hours of normal E2E cadence.

## Breakdown at 65 records

```
Platform core (apex, _vercel, api, app, doc, send, status, www, DKIM, MX)  13   legit
Tenant subdomains (chloe-dong, hongmingwang, hermes-final-*)                5   2 legit + stale orphan
ws-*                                                                       17   ~12 orphan (no corresponding live EC2)
e2e-*                                                                      21   ~15 orphan (teardown completed, DNS not cleaned)
other                                                                       9   legit (railway-verify, keys, etc.)
```

So **~42 of 65 records are stale** — mostly `e2e-*` tenants from today's E2E runs, and `ws-*` for workspaces already deleted.

## Root cause hypothesis

- `tenant.Delete` (admin teardown after E2E): terminates EC2, tears down Cloudflare tunnel, but **does not delete the `<slug>.staging.moleculesai.app` DNS record**.
- `workspace.Delete`: terminates the EC2, but **does not delete the `ws-<id>.moleculesai.app` DNS record**.

Evidence: `e2e-20260424-...` tenants ran, teardown reported "✅ Teardown clean — no orphan resources", yet their CF records remain.

## Repro

1. Trigger staging E2E run (full lifecycle).
2. Wait for the teardown step to report clean.
3. `curl -H "Authorization: Bearer $CF_TOKEN" ...zones/.../dns_records?name=e2e-<slug>.staging.moleculesai.app` → still returns the record.

## Acceptance criteria

- `tenant.Delete` and `workspace.Delete` both issue `DELETE zones/<zone>/dns_records/<id>` for their subdomain before returning.
- Integration test: create tenant → delete tenant → assert zero CF records match the tenant's slug prefix within 10s.
- CF record count plateaus at ~23–30 under continuous E2E load (legit + in-flight tests only).

## Workaround (until fix lands)

[...]

Issue #239, opened 2026-04-15
[provisioner] No clean recovery path for workspaces in 'failed' state with missing config volume


HongmingWang-Rabbit:
## Problem
When a workspace container and its `/configs` Docker volume are both destroyed (e.g. `docker rm -f ws-xxx && docker volume rm ws-xxx-configs`), the workspace ends up in `status='failed'` with no way to recover via the existing API.

**Reproduction (observed this session):**
```bash
docker rm -f ws-57e13b54-119
docker volume rm ws-57e13b54-119-configs
# Workspace row flips to 'provisioning' then 'failed' after a minute or two
# POST /workspaces/:id/restart returns 200 but doesn't create a new container
#   (response: {"config_dir":"existing-volume","reset_session":false,"status":"provisioning"})
# No other endpoint accepts a request to rebuild the config volume from template source
# POST /org/import requires AdminAuth which cannot be satisfied by a workspace bearer token
```

## Impact
A single botched operational command (e.g. a contributor testing the apply-locally pattern for a template change) leaves the workspace permanently broken until either:
- A full `docker-compose down -v && up` which nukes ALL workspaces
- Manual DB surgery + manual volume creation + manual config.yaml templating

Neither is acceptable recovery for a one-off mistake.

## Root cause
The provisioner recovers workspaces whose container is gone but volume is intact (observed: `POST /workspaces/:id/restart` + `config_dir: existing-volume` works). It does NOT know how to rebuild a missing config volume from the org-template source, so a workspace with both missing can't be fully recovered.

The `config_dir: existing-volume` branch in the restart flow implies there's also a `non-existing-volume` code path somewhere, but either:
1. It's not exposed via the API
2. It's gated on an admin-only endpoint I don't have
3. It's silently falling through when the volume is missing

## Proposed fix
1. **`POST /workspaces/:id/restart`** should accept an optional `rebuild_config` flag. When set (and caller has workspace-bearer auth), the provisioner:
   - Detects missing `/configs` volume
   - Re-reads the org-template source for this workspace (by its `files_dir` / role name)
   - Re-renders `config.yaml` + `system-prompt.md` into a new Docker volume
   - Starts a fresh container bound to that volume
   - Lets the workspace re-register (now that #215/#225 is merged, the register call uses auth_headers and won't 401)

2. **OR** add a separate endpoint `POST /workspaces/:id/rebuild-volume` with the same semantics, so the existing `restart` flow stays strict about "volume must exist".

[...]

PR #2088, opened 2026-04-26
ops(cf): hourly sweep workflow for orphan Cloudflare DNS records (#239)

HongmingWang-Rabbit:

Closes [Molecule-AI/molecule-controlplane#239](https://github.com/Molecule-AI/molecule-controlplane/issues/239).

## Why

CF zone hit the 200-record quota on 2026-04-23+ — every short-lived E2E and canary leaves a record on \`moleculesai.app\`, and **no scheduled job is pruning them**. Provisions then start failing with code 81045 (\`Record quota exceeded\`). The \`sweep-cf-orphans.sh\` script (PR #1978, decision-function unit tests added in #2079) already exists, but nothing fires it on a schedule.

## Fix

New \`.github/workflows/sweep-cf-orphans.yml\`:

- **Hourly** schedule at \`:15\` — offset from \`sweep-stale-e2e-orgs.yml\` (top-of-hour) so the two janitors don't race the same CP admin endpoint
- **\`workflow_dispatch\`** with \`dry_run\` (default true, ad-hoc verify) + \`max_delete_pct\` (override the script's 50% safety gate for major cleanups)
- **\`concurrency\`** group prevents schedule + manual-dispatch from racing the same zone

## Why not extend sweep-stale-e2e-orgs.yml

- That workflow drives \`DELETE /cp/admin/tenants/:slug\`, which assumes CP has the org row. Doesn't catch records left behind when CP never knew about the tenant (canary scratch, manual ops experiments) or when the CP-side cascade's CF-delete branch failed.
- \`sweep-cf-orphans.sh\` enumerates the CF zone directly + matches each record against live CP slugs + live AWS EC2 names. Catches what the CP-driven sweep can't.

## Required secrets (must be set on the repo before this can run)

| Secret | Source |
|---|---|
| \`CF_API_TOKEN\` | Cloudflare token with \`zone:dns:edit\` on moleculesai.app |
| \`CF_ZONE_ID\` | moleculesai.app zone ID |
| \`CP_PROD_ADMIN_TOKEN\` | CP admin bearer for api.moleculesai.app |
| \`CP_STAGING_ADMIN_TOKEN\` | CP admin bearer for staging-api.moleculesai.app |
| \`AWS_ACCESS_KEY_ID\` / \`AWS_SECRET_ACCESS_KEY\` | EC2 read-only |

Pre-flight \`Verify required secrets present\` step fails loud if any are missing — single clean error message instead of script-level multi-line noise.

## Test plan
- [ ] CI green (workflow YAML lint via the existing actionlint hook if any)
- [ ] After merge: set the 6 secrets on the repo, then trigger \`workflow_dispatch\` with \`dry_run=true\` to confirm the sweep enumerates correctly without deletions
- [ ] Confirm scheduled \`:15\` run starts converging the zone toward the lower record count over 24h
- [ ] Once stable, leave \`dry_run=false\` (the schedule's default) for ongoing pruning

## Related

- PR #1978 — script
- PR #2079 — decision-function unit tests + parity gate
- Issue #1976 — original CF DNS leak
- molecule-controlplane#261 — sister sweep for orphan EC2s

### DIFF
--- a/.github/workflows/sweep-cf-orphans.yml
+++ b/.github/workflows/sweep-cf-orphans.yml
@@ -1,0 +1,103 @@
+name: Sweep stale Cloudflare DNS records
+
+# Janitor for Cloudflare DNS records whose backing tenant/workspace no
+# longer exists. Without this loop, every short-lived E2E or canary
+# leaves a CF record on the moleculesai.app zone — the zone has a
+# 200-record quota (controlplane#239 hit it 2026-04-23+) and provisions
+# start failing with code 81045 once exhausted.
+#
+# Why a separate workflow vs sweep-stale-e2e-orgs.yml:
+#   - That workflow operates at the CP layer (DELETE /cp/admin/tenants/:slug
+#     drives the cascade). It assumes CP has the org row to drive the
+#     deprovision from. It doesn't catch records left behind when CP
+#     itself never knew about the tenant (canary scratch, manual ops
+#     experiments) or when the cascade's CF-delete branch failed.
+#   - sweep-cf-orphans.sh enumerates the CF zone directly and matches
+#     each record against live CP slugs + AWS EC2 names. It catches
+#     leaks the CP-driven sweep can't.
+#
+# Safety: the script's own MAX_DELETE_PCT gate refuses to nuke more
+# than 50% of records in a single run. If something has gone weird
+# (CP admin endpoint returns no orgs → every tenant looks orphan) the
+# gate halts before damage. Decision-function unit tests in
+# scripts/ops/test_sweep_cf_decide.py (#2027) cover the rule
+# classifier.
+
+on:
+  schedule:
+    # Hourly. Mirrors sweep-stale-e2e-orgs cadence so the two janitors
+    # converge on the same tick. CF API rate budget is generous (1200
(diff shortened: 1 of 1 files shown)


HongmingWang-Rabbit:
Reviewed across all five axes — **approve**, definitely improves code health (closes the #239 quota bleed). Tests exist via the decision-function suite in #2079 + the script is already merged. CI green. Arming auto-merge.

One non-blocking observation:

**Optional:** \`on.merge_group: types: [checks_requested]\` will fire the entire \`sweep\` job on every merge-queue evaluation. The body explains the future-required-check intent, but as written:
- While the 6 secrets are unset (the PR body's post-merge step), the verify step exits 2 on every merge-queue eval → noise but harmless.
- Once secrets ARE configured, \`github.event.inputs.dry_run\` is empty on merge_group → defaults to \"false\" → \`--execute\` branch → actual CF DELETE calls per PR going through the queue.

Two cheap fixes:
1. Drop \`merge_group\` from the \`on:\` list (re-add when actually wiring as a required check).
2. Or gate the sweep job: \`if: github.event_name != 'merge_group'\` and add a no-op success job for the merge_group path.

Worth a follow-up before turning on the secrets, but not blocking this merge — the workflow is dormant until the secrets land anyway. Filing under #239's follow-up if you'd like to track separately.

PR #2117, opened 2026-04-26
fix(canvas): tombstone deleted ids so in-flight hydrate can't resurrect them (#2069)

### LINKED ISSUE #2069: Canvas: hydrate() racing a successful cascade-delete can resurrect deleted subtrees
**Surfaced by the bundle-level review of #2061.**

`removeSubtree(rootId)` in `canvas/src/store/canvas.ts` drops a parent + every descendant locally after `DELETE /workspaces/:id?confirm=true` returns 200. `hydrate()` is server-authoritative and re-builds the entire node array from whatever \`/workspaces\` returns.

If a GET /workspaces was IN-FLIGHT before the DELETE completed, its response (still containing the deleted subtree) can land AFTER our local removeSubtree, hydrate the store with the stale snapshot, and re-introduce the deleted nodes on the canvas. The next fallback poll (10s) would correct it, so worst case the user sees the deleted org reappear briefly then vanish again.

### Repro
1. Open canvas with WS reconnecting (so the 10s fallback poll is active).
2. Right at the boundary of one poll's GET going out, click delete on a parent.
3. The DELETE returns and removeSubtree drops nodes locally.
4. The pre-delete GET response arrives → hydrate restores the deleted subtree.

### Suggested fix
- Track a per-id "locally deleted at" timestamp; hydrate filters out ids deleted within the last N seconds.
- Or: dispatch DELETE through the socket helper so we can pause the fallback poll across the request.

### Priority
Low — self-healing, low repro probability, no data loss. File for a future cleanup pass.

Originally reported in PR #2061 bundle review.

HongmingWang-Rabbit:

Closes #2069.

## Why

`removeSubtree(rootId)` drops a parent + descendants locally after `DELETE /workspaces/:id?confirm=true` returns 200. But `hydrate()` is server-authoritative — when a `GET /workspaces` was IN-FLIGHT before the DELETE completed, its response (still containing the deleted subtree) lands AFTER `removeSubtree`, hydrates the store with the stale snapshot, and re-introduces the deleted nodes on the canvas. The next 10s WS-fallback poll corrects it, so worst case the user sees the deleted org reappear briefly then vanish again — but the flicker is real and the priority assessment is "low, self-healing, no data loss" — fits a defense-in-depth fix.

## How

New module `canvas/src/store/deleteTombstones.ts` holds a transient process-lifetime `Map<id, deletedAt>`:

- `removeSubtree` calls `markDeleted(removedIds)` on every removal
- `hydrate` calls `wasRecentlyDeleted(id)` to filter out incoming workspaces whose ids match a fresh tombstone
- TTL = 10s (matches the WS-fallback poll cadence — covers a single round-trip, after which a legitimately re-imported id flows through normally)
- GC happens lazily at every read AND at every write, so the map stays bounded with no separate timer / interval / unmount plumbing

## Why module-level state, not Zustand store state

The tombstone map is purely transient and never participates in React Flow rendering. Putting it in store state would either:
- Bloat the React subscription surface (every component subscribed to nodes/edges would see tombstone changes), OR
- Require explicit non-reactive mutation patterns that fight Zustand's normal model.

A module-level `Map` lives at module identity — same scope as the singleton store. Tests reset it via `__resetTombstonesForTest()` in `beforeEach`.

## Test plan

- [x] Unit tests for `deleteTombstones.ts`: 7 cases (immediate flag, never-marked, TTL boundary at 9999ms vs 10001ms, GC-on-read, GC-on-write, re-mark resets timestamp, iterable input)
- [x] End-to-end test in `canvas.test.ts`: `hydrate` after `removeSubtree` with a stale snapshot can NOT resurrect deleted ids
- [ ] CI green (incl. Canvas tabs E2E now that cp#282 unblocked the staging-orchestration layer)
- [ ] Post-merge: manual smoke — open canvas, delete a parent, immediately trigger a hydrate (page focus / WS reconnect), confirm the deleted subtree stays gone

### DIFF
--- a/canvas/src/store/__tests__/canvas.test.ts
+++ b/canvas/src/store/__tests__/canvas.test.ts
@@ -6,6 +6,7 @@ global.fetch = vi.fn(() =>
 );
 
 import { useCanvasStore, summarizeWorkspaceCapabilities } from "../canvas";
+import { __resetTombstonesForTest } from "../deleteTombstones";
 import type { WorkspaceData, WSMessage } from "../socket";
 
 // Helper to build a WorkspaceData object with sensible defaults
@@ -52,6 +53,10 @@ beforeEach(() => {
     searchOpen: false,
     viewport: { x: 0, y: 0, zoom: 1 },
   });
+  // Tombstones leak across tests because the module-level map is
+  // process-lifetime by design. Reset between tests so a delete in one
+  // test doesn't shadow a hydrate in the next.
+  __resetTombstonesForTest();
   vi.clearAllMocks();
 });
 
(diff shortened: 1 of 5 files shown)


HongmingWang-Rabbit:
## Self-review (5-axis pass)

**Correctness:** ✓ Both `computeAutoLayout` and `buildNodesAndEdges` consume the filtered `live`; `currentParentSizes` is built from store state (`get().nodes`), independent of the incoming snapshot. Tombstone TTL = `FALLBACK_POLL_MS` covers a single in-flight round-trip. Re-mark resets the timestamp (verified by test).

**Tests:** ✓ 7 unit cases (mark/never-marked/TTL boundary 9999ms vs 10001ms/GC-on-read/GC-on-write/re-mark/iterable input) + 1 integration case in `canvas.test.ts` that verifies the full removeSubtree → stale hydrate → cannot resurrect path. Fake timers used for determinism.

**Architecture:** ✓ Module-level `Map` is the right call (not Zustand store state) — won't trigger React Flow re-renders. `__resetTombstonesForTest` exported for test isolation. `FALLBACK_POLL_MS` exported from `socket.ts` so the constant has one source of truth.

**Security:** N/A — no auth, no boundaries, no data egress.

**Performance:** ✓ GC is O(n) on tombstone count, run on every mark/read; with 10s TTL and rare deletes, count stays in single digits. `wasRecentlyDeleted` is single-key lookup → O(1) per workspace. Hydrate stays O(n).

### FYI
- TTL is exactly `FALLBACK_POLL_MS` (10s). If a poll response is delayed past 10s under degraded network, a stale id can still slip through. Acceptable here because the priority assessment is "low, self-healing, no data loss" — defense-in-depth.

Approval pending external reviewer (can't self-approve as PR author). Arming auto-merge so it ships once approval lands.

hongmingwang-moleculeai (approved):
LGTM

Issue #2071, opened 2026-04-25
Canvas test gaps: OrgCancelButton, A2AEdge, useTemplateDeploy, useDragHandlers


HongmingWang-Rabbit:
**Follow-up from PR #2061 bundle review.**

The branch added several new components / hooks without dedicated tests:
- \`canvas/src/components/canvas/OrgCancelButton.tsx\` (165 LOC, NEW)
- \`canvas/src/components/canvas/A2AEdge.tsx\` (133 LOC, NEW)
- \`canvas/src/hooks/useTemplateDeploy.tsx\` (170 LOC, NEW shared hook)
- \`canvas/src/components/canvas/useDragHandlers.ts\` (modified)
- \`workspace-server/internal/handlers/org_import.go\` (+240 LOC; only adjacent org_test.go exists)

Manual E2E was done for the deploy/cancel/import flows, but per the auto-memory ("unit tests pass while real bugs ship"), behavioral coverage would catch regressions earlier.

Suggested order:
1. \`useTemplateDeploy\` (shared hook, exercised by both Templates panel and OrgCancelButton)
2. \`A2AEdge\` (rendering + click-to-select-source behavior)
3. \`OrgCancelButton\` (cancel flow, optimistic state)

### Priority
Medium — missing coverage on net-new code.

HongmingWang-Rabbit:
**Surfacing for approval — judgment call on test scope + which framework idioms.**

### 1. Problem statement
PR #2061 added 4 net-new canvas surfaces (`OrgCancelButton.tsx` 165 LOC, `A2AEdge.tsx` 133 LOC, `useTemplateDeploy.tsx` 170 LOC, `useDragHandlers.ts` modified) plus 240 LOC in `org_import.go` without dedicated tests. Manual E2E was done; no behavioral coverage. Per `feedback_always_run_e2e` and the auto-memory ("unit tests pass while real bugs ship"), behavioral tests catch regressions earlier — but writing them is a non-trivial scope decision (which to test, what fidelity, what to mock).

### 2. Proposed approach
[...]

PR #2121, opened 2026-04-26
test(canvas): unit tests for useTemplateDeploy (#2071)

HongmingWang-Rabbit:

Closes the first item from #2071. Adds behavioural coverage for the shared template-deploy hook that both TemplatePalette (sidebar) and EmptyState (welcome grid) drive.

## Coverage — 10 cases across 4 buckets

**Happy path (4):**
- preflight ok → `POST /workspaces` → `onDeployed` fires with new id
- caller-supplied `canvasCoords` flows into the POST body
- default coords fall in `[100,500) × [100,400)` when omitted
- `template.runtime` is preferred over the `resolveRuntime` fallback (locks the deduped-fallback contract from #2061)

**Preflight failures (2):**
- network throw sets `error` AND clears `deploying` — regression test for the *stranded button* bug called out in the SUT's inline comment. Drop the try block and you fail this test.
- not-ok-with-missing-keys opens the modal without firing POST

**Modal lifecycle (2):**
- *'keys added'* click retries POST without re-running preflight (verifies the `executeDeploy` / `deploy` split — preflight call count stays at 1, POST goes to 1)
- *'cancel'* click closes modal without firing POST

**POST failures (2):**
- `Error` rejection surfaces the message
- non-`Error` rejection surfaces the *Deploy failed* fallback

## Mock strategy

Mocks `@/lib/api`, `@/lib/deploy-preflight`, and `@/components/MissingKeysModal` — stand-in modal renders two `data-testid` buttons exposing the `onKeysAdded` and `onCancel` callbacks. The real radix modal isn't relevant to this hook's behavior. Pattern matches the `vi.hoisted` + import-after-mocks structure already used in `canvas/src/app/__tests__/orgs-page.test.tsx`.

## Test plan

- [x] All 10 cases pass locally (`vitest run useTemplateDeploy.test.tsx` — 31ms)
- [x] No changes to the SUT — pure additive coverage
- [ ] CI green

Follow-ups in separate PRs:
- `A2AEdge` rendering + click-to-select-source
- `OrgCancelButton` cancel flow + optimistic state

### DIFF
--- a/canvas/src/hooks/__tests__/useTemplateDeploy.test.tsx
+++ b/canvas/src/hooks/__tests__/useTemplateDeploy.test.tsx
@@ -1,0 +1,316 @@
+// @vitest-environment jsdom
+/**
+ * Tests for useTemplateDeploy — the shared preflight + POST + modal
+ * hook used by TemplatePalette (sidebar) and EmptyState (welcome grid).
+ *
+ * Behavioural coverage for the three flows the hook owns:
+ *   1. Happy path: preflight ok → POST /workspaces → onDeployed fires
+ *   2. Preflight errors: network throw vs not-ok-with-missing-keys
+ *      (different code paths — the throw must NOT strand `deploying`,
+ *      see the inline comment in the SUT for the prior bug)
+ *   3. Modal lifecycle: keys-added retries POST without re-running
+ *      preflight; cancel closes without POST
+ *
+ * Issue: #2071 (Canvas test gaps follow-up).
+ */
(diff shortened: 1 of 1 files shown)


HongmingWang-Rabbit:
Skipping review this pass: CI re-running after `update-branch` merged in #2120's secret-scan fix (the original failure was the pre-fix `secret-formats.test.ts` literal token still on this PR's stale base). Will re-evaluate once the new run completes.

Issue #284, opened 2026-04-15
security(ssrf): PR #283 branch contains PR #270's SSRF code — merging it would introduce issue #272


HongmingWang-Rabbit:
## Finding

**Severity:** MEDIUM  
**Type:** SSRF via transitive branch dependency  
**Affected PRs:** #283 (browser-automation fix), #270 (transcript endpoint)  

### Description

PR #283 (`fix/browser-automation-helper-defaultviewport`) was built on top of PR #270's branch (`feat/workspace-transcript-endpoint`). The commit log of PR #283's branch is:

```
e34e85c fix(browser-automation): canonical connect() helper enforces defaultViewport: null   ← PR #283 own change
9b5cd23 fix(reno-stars): SEO Builder delegates code blockers to Dev Leader, not human         ← PR #271
1f22d7d feat: GET /workspaces/:id/transcript — live agent session log                        ← PR #270 (SSRF)
```

This means **merging PR #283 to main would also merge `platform/internal/handlers/transcript.go` from PR #270**, which contains the unfixed SSRF documented in issue #272:

```go
// transcript.go (lines 61-73) — reads URL from DB, no scheme/host validation
target, err := url.Parse(workspaceURL)   // workspaceURL = agent_card->>'url' from DB
target.Path = "/transcript"
target.RawQuery = c.Request.URL.RawQuery  // user-controlled query string forwarded
req, _ := http.NewRequestWithContext(reqCtx, "GET", target.String(), nil)
resp, err := h.httpClient.Do(req)          // SSRF — no allowlist
```

### Risk

If PR #270 is left open (blocked on #272) but PR #283 is reviewed and merged independently, the SSRF vulnerability enters main silently — the reviewer would need to notice that PR #283's diff includes `transcript.go` despite the PR title only mentioning `browser-automation`.

### Recommended Actions

**For PR #283 author:** Rebase `fix/browser-automation-helper-defaultviewport` onto `main` so it only contains the `connect.js` changes. The browser-automation fix is clean and can merge immediately after rebase.

**Merge order (correct):**  
1. Fix SSRF in PR #270 (see issue #272)
2. Merge PR #270
3. Merge PR #271 (or rebase #283 onto main first)
4. Merge PR #283 (after rebase)

**Alternatively:** Close PR #283, rebase the `connect.js` change onto `main`, and open a clean replacement PR that only touches `plugins/browser-automation/`.

### Related
- Issue #272 — the SSRF in `transcript.go` itself
[...]

PR #2126, opened 2026-04-26
fix(delegation): runtime handles 202+queued; canvas surfaces delegation rows

HongmingWang-Rabbit:
## What this fixes

Two bugs that compounded into the **\"Director does the work itself\"** UX you saw today (workspace \`8b0bf0b3\`):

### 1. Runtime treated 202+queued as silent failure
\`workspace/builtin_tools/delegation.py:_execute_delegation\` only handled HTTP 200. When the peer's a2a-proxy returned **HTTP 202 + \`{queued: true}\`** (single-SDK-session bottleneck on the peer), the loop fell through. Two iterations later \`if \"error\" in result\` tried to access an unbound \`result\`, the goroutine ended quietly, the delegation stayed at FAILED with \`error=\"None\"\`. The LLM checking status saw \"failed\" + the platform's \"Delegation queued — target at capacity\" log line in chat context, concluded the peer was permanently unavailable, and bypassed delegation to do the work itself.

DB evidence from the live workspace (6 delegation rows, all \`status='queued'\`):

\`\`\`
delegation queued → UX Researcher    \"Delegating to f6f3a023...\"
delegation queued → UX Researcher    \"Delegation queued — target at capacity\"
delegation queued → Visual Designer  \"Delegating to a2e92524...\"
delegation queued → Visual Designer  \"Delegation queued — target at capacity\"
delegation queued → React Engineer   \"Delegating to fd1a1f96...\"
delegation queued → React Engineer   \"Delegation queued — target at capacity\"
\`\`\`

### 2. Canvas Agent Comms tab dropped every delegation row
\`AgentCommsPanel.tsx\` filtered to \`a2a_send | a2a_receive\` only. \`activity_type='delegation'\` rows never reached \`toCommMessage\`. User saw \"No agent-to-agent communications yet\" while the DB had 6+ delegations.

## How it's fixed

| File | Change |
|---|---|
| \`workspace/builtin_tools/delegation.py\` | Add \`DelegationStatus.QUEUED\`. Explicit \`if a2a_resp.status_code == 202\` branch in \`_execute_delegation\` — recognizes \`{queued: true}\`, marks local state QUEUED, mirrors to platform, returns cleanly without retrying. \`check_delegation_status\` docstring extended with explicit per-status guidance: \"queued → wait, peer is on prior task, do NOT bypass.\" |
| \`canvas/src/components/tabs/chat/AgentCommsPanel.tsx\` | Include \`delegation\` in both the initial filter and the WS push filter. New \`delegation\` branch in \`toCommMessage\` — maps as outbound (always; platform proxies on our behalf), uses \`summary\` as primary text. |

## Tests

- 3 new Python tests (\`workspace/tests/test_delegation.py\` → \`TestA2AQueued\`):
  - 202+queued marks status as QUEUED, no error field
  - 202+queued does NOT retry the A2A POST (counted by URL match — mock is shared across all AsyncClient calls)
  - bare 202 without \`{queued:true}\` still falls through to existing retry-then-FAILED path
- 3 new TS tests (\`AgentCommsPanel.test.ts\`):
  - \`delegate\` row maps as outbound with summary text
  - \`delegate_result\` queued row preserves \`status='queued'\` (load-bearing for the LLM's wait-vs-bypass decision)
  - missing \`target_id\` returns \`null\` instead of rendering a ghost

All 20 \`test_delegation.py\` tests pass; all 8 \`AgentCommsPanel.test.ts\` tests pass.

The underlying **single-SDK-session bottleneck** (peer can only handle one A2A request at a time). Tracked as **task #102** — real architectural work. This PR makes the runtime handle the queueing correctly so the LLM doesn't bail, and makes delegations visible in Agent Comms so operators can see what's happening. Once #102 lands, the queueing will become rare anyway.

## Test plan

- [x] \`pytest workspace/tests/test_delegation.py\` green (20/20)
- [x] \`vitest run AgentCommsPanel.test.ts\` green (8/8)
- [ ] Manual: provision a director + 3 sub-agents, send director a multi-step task, observe Agent Comms panel populates with delegation rows including queued ones; observe director NOT bypassing when peer is queued (status surfaced as \"queued\" in chat, director acknowledges and waits)

### DIFF
--- a/canvas/src/components/tabs/chat/AgentCommsPanel.tsx
+++ b/canvas/src/components/tabs/chat/AgentCommsPanel.tsx
@@ -59,6 +59,34 @@ function resolveName(id: string): string {
 }
 
 export function toCommMessage(entry: ActivityEntry, workspaceId: string): CommMessage | null {
+  // delegation activity rows are written by the platform's /delegate
+  // handler. They're always outbound from this workspace's POV (the
+  // platform proxies the A2A on our behalf). Two methods:
+  //   - "delegate"        — the initial outbound; status pending/dispatched
+  //   - "delegate_result" — the eventual reply; status completed/queued/failed
+  // We surface them in Agent Comms because they ARE agent-to-agent
+  // calls; without this branch they'd be dropped by the activity_type
+  // filter and the user would see "No agent-to-agent communications yet"
+  // even when the director made delegations.
+  if (entry.activity_type === "delegation") {
+    const peerId = entry.target_id || "";
+    if (!peerId) return null;
+    return {
+      id: entry.id,
+      flow: "out",
+      peerName: resolveName(peerId),
+      peerId,
+      // Prefer summary (set by the platform with a human-readable
+      // string like "Delegating to X" or "Delegation queued — target
+      // at capacity"). Fall back to request body for older rows that
+      // pre-date the summary column being populated.
+      text: entry.summary || extractRequestText(entry.request_body) || "(delegation)",
+      responseText: entry.response_body ? extractResponseText(entry.response_body) : null,
+      status: entry.status || "ok",
+      timestamp: entry.created_at,
+    };
+  }
+
   // a2a_receive activity rows come in two shapes:
   //
   //   1. Real incoming call (a peer called us): source_id = the peer,
@@ -132,7 +160,11 @@ export function AgentCommsPanel({ workspaceId }: { workspaceId: string }) {
     api.get<ActivityEntry[]>(`/workspaces/${workspaceId}/activity?source=agent&limit=50`)
       .then((entries) => {
         const filtered = (entries ?? [])
-          .filter((e) => e.activity_type === "a2a_send" || e.activity_type === "a2a_receive")
+          .filter((e) =>
+            e.activity_type === "a2a_send" ||
+            e.activity_type === "a2a_receive" ||
+            e.activity_type === "delegation",
+          )
           .reverse();
         const msgs: CommMessage[] = [];
         for (const e of filtered) {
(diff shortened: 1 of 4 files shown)


HongmingWang-Rabbit:
Still skipping: failed E2E correlates with the active Cloudflare `Minor Service Outage` (status page indicator updated 22:08:40Z). CP #284's fast-fail surfaces CF API failures in ~16s instead of 900s — that's working as intended; the underlying CF outage is the blocker. Re-trigger when [cloudflarestatus.com](https://www.cloudflarestatus.com/) returns to `operational`.

HongmingWang-Rabbit:
## Code review (5-axis)

**Correctness:** ✓ Two-part fix lands cleanly:

1. **Runtime side (`delegation.py`)** — explicit 202+queued branch returns early after marking `DelegationStatus.QUEUED`, calling `_notify_completion(..., 'queued')`, and `_update_delegation_on_platform(..., 'queued', ...)`. The malformed-JSON guard (`try: a2a_resp.json() except: queued_body = {}`) plus the strict `is True` check on `queued` correctly falls through to the existing failure path on 202-without-queued-flag — no silent state confusion.

2. **Canvas side (`AgentCommsPanel`)** — filter expanded + `toCommMessage` mapping for `activity_type === 'delegation'`. The `!peerId` guard returns null for malformed rows (matches the test contract).

**Tests:** ✓ Coverage is the right shape — `TestA2AQueued.test_queued_marks_status_queued_not_failed` AND `test_queued_does_not_retry`. The retry-count assertion is non-trivial (filtering A2A POSTs from platform-sync POSTs by URL match) but correct. Canvas tests cover delegate / delegate_result + the null-target_id defensive case.

**Architecture:** ✓ `QUEUED` enum addition is forward-compatible (existing clients see other values unchanged). Status-semantics docstring on `check_delegation_status` is the right place to teach the LLM what `queued` means — that's where the prompt-side guidance has to live.

**Security:** N/A — no auth/data-boundary changes; the queued-body is parsed but only `.get('queued')` is read, so a hostile peer can't smuggle anything via the rest of the JSON.

**Performance:** ✓ One additional branch + one body parse on 202 responses. Negligible.

### FYI

- The `queued_body` malformed-JSON path silently sets `{}` and falls through. Worth a one-liner log here too (mirrors the env-var pattern in #2128) — would help diagnose a future "peer is returning weird 202s" incident. Not blocking.
- The LLM-prompt-side "don't bypass on queued" guidance lives in the docstring. Whether the runtime actually honors it depends on the agent's prompt construction — outside this PR's scope, but worth a follow-up to verify the LLM does the right thing in a real queued scenario before declaring incident-closed.

LGTM. Auto-merge already armed.

hongmingwang-moleculeai (approved):
LGTM

Issue #286, opened 2026-04-15
org-templates: add investment-committee template (vertical AI team pattern)


HongmingWang-Rabbit:
## Background

The eco-watch survey (2026-04-15, :48 sweep) flagged **virattt/ai-hedge-fund** (55k ⭐, MIT), a viral demo that:

- Implements the **"AI company of specialist agents"** pattern — exactly the metaphor Starfire is built on
- Uses **LangGraph** (our first-class runtime adapter) for agent coordination
- Features a **React Flow visual editor** for drag-and-drop agent committee composition
- Architecture: 14 domain-specialist agents → 4 cross-domain analysts → 1 coordinator (PM equivalent)

With 55k stars and explosive growth (8.7k → 55k in 14 months), this demonstrates validated market demand for the vertical AI team pattern in a specific domain.

## Proposal

Add `org-templates/investment-committee/` as a new Starfire org template:

```
org-templates/investment-committee/
  org.yaml           # Roles: value-investor, growth-analyst, risk-manager, quant, portfolio-coordinator
  AGENTS.md          # System prompt guidance for each role
  workspace-configs/ # Per-role config.yaml files
  README.md          # "Deploy an AI investment committee in 5 minutes"
  skills/            # market-data, sentiment-analysis, fundamentals-fetch plugins
```

**Why this is low-cost:** The topology (N specialists → M analysts → 1 coordinator) is identical to our existing `org-templates/dev-team/`. Mostly a copy-and-rename with domain-specific system prompts.

## Value

1. **Immediate showcase asset** — a concrete vertical org template that matches a project developers already know (ai-hedge-fund) but runs as a real persistent Starfire org with memory, scheduling, and A2A.
2. **LangGraph adapter validation** — exercises our LangGraph runtime with a real multi-agent workload, surfacing any gaps before enterprise customers do.
3. **Canvas demo material** — the investment committee visual pattern (role bubbles → analyst layer → coordinator) is compelling for Canvas screenshots/marketing.

## Stretch goal

If the ai-hedge-fund React Flow editor is extracted as a standalone library, evaluate adopting it as a drag-and-drop composition layer for the Canvas (users build their org from a role palette, then click Deploy).

## References

- Eco-watch entry: `docs/ecosystem-watch.md` → AI Hedge Fund section
- Source repo: https://github.com/virattt/ai-hedge-fund
- Existing template to copy: `org-templates/` (whichever dev-team template is most mature)

[...]

PR #2127, opened 2026-04-26
docs(ci): fix secret-scan reusable workflow self-doc — repo is molecule-core, ref is @staging

HongmingWang-Rabbit:

Follow-up to #2109. The canonical secret-scan workflow's own self-documented enrollment example referenced `Molecule-AI/molecule-monorepo/.github/workflows/secret-scan.yml@main` which is wrong on two counts:

1. **Repo is molecule-core, not molecule-monorepo.** The repo was renamed at some point; `molecule-monorepo` is now a redirect alias (which is why the audit log review surfaced 169 'agent events on molecule-monorepo orphan' — those are events on the redirect, not a real orphaned repo).
2. **Ref must be `@staging` not `@main`.** Staging is the active default branch; main lags behind via the staging-promotion workflow.

Discovered while shipping the rollout PRs to molecule-controlplane (cp#286) and molecule-ai-workspace-runtime (#57) — both initially failed CI because the path resolved to a non-existent workflow. Fixed in both consumer PRs in the same iter; this PR fixes the canonical self-doc so future operators copy the right snippet.

## Test plan
- [ ] CI green (this PR has no behavioural change — it's a comment edit)
- [ ] Post-merge: cp#286 + runtime#57 both pass after their consumer-side fixes propagate

### DIFF
--- a/.github/workflows/secret-scan.yml
+++ b/.github/workflows/secret-scan.yml
@@ -12,7 +12,11 @@ name: Secret scan
 #
 #   jobs:
 #     secret-scan:
-#       uses: Molecule-AI/molecule-monorepo/.github/workflows/secret-scan.yml@main
+#       uses: Molecule-AI/molecule-core/.github/workflows/secret-scan.yml@staging
+#
+# Pin to @staging not @main — staging is the active default branch,
+# main lags via the staging-promotion workflow. Updates ride along
+# automatically on the next consumer workflow run.
 #
 # Same regex set as the runtime's bundled pre-commit hook
 # (molecule-ai-workspace-runtime: molecule_runtime/scripts/pre-commit-checks.sh).


HongmingWang-Rabbit:
## Self-review (5-axis pass)

**Correctness:** ✓ Both corrections match reality — the canonical `secret-scan.yml` lives in `molecule-core` (not `molecule-monorepo`, which doesn't have it), and `@main` doesn't have it yet because monorepo's #2109 merged to staging only and the staging→main promote hasn't run. `@staging` is the only ref that resolves today. (Same root cause that broke CP #286.)

**Tests:** N/A — pure documentation change.

**Architecture:** ✓ Pinning external consumers to `@staging` is a known trade-off vs SHA-pinning; the inline note flags that staging is the active default branch and updates ride along automatically. Acceptable for the org-wide convenience pattern.

**Security:** N/A.

**Performance:** N/A.

LGTM. Auto-merge armed.

hongmingwang-moleculeai (approved):
LGTM

PR #2128, opened 2026-04-26
fix(a2a-proxy): close 60s context-canceled gap on long silent runs

HongmingWang-Rabbit:
## What this fixes

The "context canceled" wave on 2026-04-26 — 15+ failed user/agent A2A calls in 1hr across 6 workspaces (including the user's "send it in the chat" message that the director never received).

## Root cause

Two compounding bugs:

1. **\`a2a_proxy.go:applyIdleTimeout\` cancels dispatch after 60s of broadcaster silence** for the workspace. Resets on any SSE event; fires \`cancel()\` if no event in time.
2. **\`registry.go:Heartbeat\` broadcast was conditional** — \`if payload.CurrentTask != prevTask\`. Runtime POSTs \`/registry/heartbeat\` every 30s, but if \`current_task\` is unchanged the handler emits ZERO broadcasts. \`evaluateStatus\` only broadcasts on online/degraded transitions — also no-op when steady.

Net: claude-code agent on a long packaging step keeps the same task for >60s → no broadcasts → idle timer fires → in-flight request cancelled mid-flight.

## Fix

| | What | Why |
|---|---|---|
| (b) | Bump \`idleTimeoutDuration\` 60s → 5min, env-overridable via \`A2A_IDLE_TIMEOUT_SECONDS\` | Safety net for any path where heartbeat goes silent (e.g. runtime crashed). Tunable for canary-fast-fail vs prod-slow-plugin needs. |
| (c) | Heartbeat handler unconditionally emits \`WORKSPACE_HEARTBEAT\` via \`BroadcastOnly\` (no DB write) | At the existing 30s runtime cadence, resets the idle timer twice per minute. Runtime side requires no change — \`workspace/heartbeat.py\` already POSTs every 30s. |

Either fix alone closes today's gap; both together is defence in depth.

## Test

\`TestHeartbeatHandler_AlwaysBroadcastsHeartbeat\` (new) pins the property that an SSE subscriber observes a \`WORKSPACE_HEARTBEAT\` broadcast on a same-task heartbeat — the exact regression scenario. All 16 existing handler tests still pass.

## Cost

\`BroadcastOnly\` skips the DB write. Per-heartbeat cost is one in-memory channel send per active SSE subscriber + one WS hub fan-out. At 30s cadence this is far below any noise floor on either path.

- Task **#102** (single SDK session bottleneck — peers can still queue when busy). But this PR ensures the queue/wait flow actually completes instead of being killed by the idle timer mid-wait.

## Test plan

- [x] \`go test ./internal/handlers/\` green (17/17 including the new test)
- [ ] Manual on staging: provision a director + sub-agents, send a long-running task, confirm no \`ProxyA2A forward error: context canceled\` in platform logs and the message gets through

### DIFF
--- a/workspace-server/internal/handlers/a2a_proxy.go
+++ b/workspace-server/internal/handlers/a2a_proxy.go
@@ -490,13 +490,30 @@ func normalizeA2APayload(body []byte) ([]byte, string, *proxyA2AError) {
 // idleTimeoutDuration is the per-dispatch silence window: if the
 // platform's broadcaster emits no events for this workspace for the
 // full duration, the dispatch ctx is cancelled. Resets on every
-// ACTIVITY_LOGGED / TASK_UPDATED / A2A_RESPONSE event for the
-// workspace, so a chat that's actively reporting tool calls or
-// streaming status updates never trips it. Picked to be longer than
-// any reasonable single-tool-use cadence (Claude Code's slowest
-// observed silence between tools is ~30s) but short enough that a
-// truly wedged runtime fails in 1 minute, not 5.
-const idleTimeoutDuration = 60 * time.Second
+// broadcaster event for the workspace — including the WORKSPACE_HEARTBEAT
+// fired by the registry's /heartbeat handler every 30s, so a runtime
+// that's just thinking silently between tool calls keeps the connection
+// alive without having to emit ACTIVITY_LOGGED noise.
+//
+// Pre-2026-04-26 this was 60s, picked when the platform only broadcast
+// on TASK_UPDATED (which itself only fires when current_task CHANGES).
+// A claude-code agent doing a long packaging step or a slow model thought
+// kept the same current_task for >60s, fired no broadcast, got cancelled
+// mid-flight. Bumped to 5min as a safety net AND the heartbeat handler
+// now broadcasts unconditionally — together either one alone closes the
+// gap, both together is defence in depth.
+//
+// Override via A2A_IDLE_TIMEOUT_SECONDS for ops who want to tune (e.g.
+// shorter for canary/test runners that want fail-fast on wedge, longer
+// for prod tenants running unusually slow plugins).
+var idleTimeoutDuration = func() time.Duration {
+	if v := os.Getenv("A2A_IDLE_TIMEOUT_SECONDS"); v != "" {
+		if n, err := strconv.Atoi(v); err == nil && n > 0 {
+			return time.Duration(n) * time.Second
+		}
+	}
+	return 5 * time.Minute
+}()
 
 // dispatchA2A POSTs `body` to `agentURL`. Uses WithoutCancel so delegation
 // chains survive client disconnect (browser tab close). Two layers of
--- a/workspace-server/internal/handlers/a2a_proxy.go
+++ b/workspace-server/internal/handlers/a2a_proxy.go
@@ -506,14 +506,34 @@ func normalizeA2APayload(body []byte) ([]byte, string, *proxyA2AError) {
 // Override via A2A_IDLE_TIMEOUT_SECONDS for ops who want to tune (e.g.
 // shorter for canary/test runners that want fail-fast on wedge, longer
 // for prod tenants running unusually slow plugins).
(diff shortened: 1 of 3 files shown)


HongmingWang-Rabbit:
## Self-review (5-axis pass)

**Correctness:** ✓ Two compounding fixes, defense-in-depth:
- `idleTimeoutDuration` 60s → 5min (now env-overridable via `A2A_IDLE_TIMEOUT_SECONDS`)
- `/registry/heartbeat` now unconditionally broadcasts `WORKSPACE_HEARTBEAT` on every successful tick — closes the silent-task-no-broadcast window the idle timer was stuck on

Either fix alone closes the gap; together they're belt+braces. The new const-to-var change for the timeout is package-init-evaluated, so any `os.Getenv` resolution happens once at startup — no runtime overhead per dispatch.

**Tests:** ✓ `TestHeartbeatHandler_AlwaysBroadcastsHeartbeat` pins the regression — same-task heartbeat (the silent-task scenario that was firing zero broadcasts pre-fix) must still produce `WORKSPACE_HEARTBEAT`. SubscribeSSE happens before the heartbeat call so no race. 200ms drain timeout is plenty for in-process channel events.

**Architecture:** ✓ `BroadcastOnly` correctly skips the DB write (per the inline comment). At 30s heartbeat cadence the per-tick cost is one in-memory channel send per SSE subscriber plus one WS hub fan-out — negligible.

**Security:** N/A — no auth/data-boundary changes; the new payload is just `active_tasks` + `uptime_seconds` (already-public stats).

**Performance:** ✓ Trade-off acknowledged inline: 5min default means a truly-wedged runtime now hangs for 5min instead of 1min. But the 60s-cancels-everything-mid-thought case was hitting common-path users (15+ failures/hr per the incident); 5min is comfortably above any plausible single-tool-cadence and below any human-noticeable wedge horizon. Override-via-env covers the canary/test-runner case that wants fail-fast.

### FYI
- The 5min default + heartbeat broadcast each independently close the gap, so even if one path regresses (e.g., heartbeat removed in a refactor) the other remains. Worth a brief note in the codeowner doc that `/heartbeat` is now load-bearing for a2a-proxy idle behavior — but that's a follow-up doc PR, not blocking.

LGTM. Auto-merge armed.

HongmingWang-Rabbit:
## Re-review (post-iteration)

The PR iterated since my prior pass — `parseIdleTimeoutEnv` is now an extracted, named, table-tested function with logging on bad input. That addresses every FYI from my last comment. Specifically what's new and good:

- `TestParseIdleTimeoutEnv` covers all 8 input shapes including `-30`, `0`, `1.5`, `60s`, `foo` — exhaustive bad-input coverage
- `defaultIdleTimeoutDuration` extracted as a const so tests can reference it without hardcoding the literal twice
- Misconfig now logs (`A2A_IDLE_TIMEOUT_SECONDS=foo is not a valid integer; using default …`) instead of silently falling back — operator-friendly

### Five-axis pass (post-iteration)

**Correctness:** ✓ Init-time evaluation is fine for this use case (per-process tunable). The 202+broadcast race ordering is correct: heartbeat broadcast fires *before* `c.JSON("ok")` so the SSE event is queued before the response returns.

**Readability:** ✓ Function-extracted, doc'd, table-driven test names read like sentences.

**Architecture:** ✓ `BroadcastOnly` skip-DB path is the right choice — heartbeat is ephemeral; persisting it would 2x the activity_logs row count for zero observability gain.

**Security:** N/A.

**Performance:** ✓ At 30s heartbeat × N subscribers × M workspaces, well below noise floor.

### Optional / Nit (non-blocking)

- The `gotHeartbeat` test loop uses `goto done` — a flag would be slightly cleaner but `goto` is idiomatic Go for this exact pattern; not worth churning.
- A complementary test for "task changed → BOTH `TASK_UPDATED` and `WORKSPACE_HEARTBEAT` fire" would lock down that the new broadcast is *additive* not *replacing*. Not blocking — current test demonstrates the regression-fix path which is what matters.

LGTM — auto-merge stays armed. Solid iteration.

hongmingwang-moleculeai (approved):
LGTM

PR #114, opened 2026-04-15
fix+feat: category_routing (#93), X-RateLimit headers (#105), workflow_run webhook (#101)

HongmingWang-Rabbit:
## Summary

Supersedes PR #108 and PR #109 — same code, rebased cleanly onto current main (post PRs #94, #95, #99, #106).

The original branches were created before those PRs merged and showed -965 deletions vs main (they would have reverted supervised.go, wsauth_middleware_test.go, and org_path_test.go). The commits themselves were clean — this branch is just those two commits cherry-picked onto main.

### Changes (from PR #108)
- `org-templates/molecule-dev/org.yaml` — add `research/plugins/template/channels` to `category_routing` defaults. Without these, evolution crons firing with those categories found no target and their audit summaries silently dropped at PM.
- `platform/internal/middleware/ratelimit.go` — emit `X-RateLimit-Limit / -Remaining / -Reset` on every response (allowed and throttled) and `Retry-After` on 429s per RFC 6585.
- `platform/internal/middleware/ratelimit_test.go` — 2 new tests covering allowed and throttled paths.

### Changes (from PR #109)
- `platform/internal/handlers/webhooks.go` — `buildGitHubA2APayload` now handles `workflow_run` events, routing failed CI runs to a workspace via the existing `X-Molecule-Workspace-ID` webhook path. Only `completed` runs with `failure/cancelled/timed_out` conclusions fan out.
- `platform/internal/handlers/webhooks_workflow_test.go` — 4 new tests: failure path, success skip, non-completed skip, short-SHA edge case.

## Test plan
- [x] All changes cherry-pick cleanly onto main with zero conflicts
- [x] No existing files deleted or reverted
- [ ] CI (blocked by GitHub Actions runner quota — waiting for recovery)

Closes #93, #105, #101

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2130, opened 2026-04-27
feat(notify): agent → user file attachments via send_message_to_user

HongmingWang-Rabbit:
Closes the gap from today's Director-can't-send-the-ZIP screenshot. The runtime had no API for outbound file attachments — agents could only paste paths in text. This PR wires the outbound side using the existing chat-uploads infrastructure.

## End-to-end shape

\`\`\`
agent: send_message_to_user("Done!", attachments=["/tmp/build.zip"])
 → runtime: POST /chat/uploads (multipart) → uri: workspace:/...build.zip
 → runtime: POST /notify {message, attachments[]}
 → platform: broadcast AGENT_MESSAGE + persist as response_body.parts[]
 → canvas: WS push (live) AND reload (extractFilesFromTask) both render chip
\`\`\`

## Why /tmp works even though it's outside canvas's allowed roots

The runtime tool reads bytes locally and re-uploads through \`/chat/uploads\`, which lands the file in \`/workspace/.molecule/chat-uploads/\` (an allowed root). The agent can attach any readable path.

## Files

| | |
|---|---|
| \`workspace-server/internal/handlers/activity.go\` | \`NotifyAttachment\` struct, body accepts \`attachments[]\`, broadcasts in payload, persists \`response_body.parts[].kind=file\` |
| \`canvas/src/store/canvas-events.ts\` | AGENT_MESSAGE handler reads \`payload.attachments\`, type-validates, attaches to \`agentMessages\` queue |
| \`workspace/a2a_tools.py\` | \`tool_send_message_to_user(message, attachments=[paths])\` + new \`_upload_chat_files\` helper. Fail-fast on missing/upload-error — never sends notify with half-rendered chip |
| \`workspace/a2a_mcp_server.py\` | Schema declares \`attachments\` param so claude-code SDK exposes it. Defensive non-string filter on dispatch |

## Tests

- 4 new Python: success path / missing file / upload 5xx / no-attach backwards compat
- 1 new Go: Notify-with-attachments persists \`parts[]\` in response_body (chat reload reconstructs chip)
- All 67 existing Python tests + 33 canvas-events tests + Go handler suite pass

## Test plan

- [x] \`go test ./internal/handlers/\` green
- [x] \`vitest run canvas-events.test.ts\` green
- [x] \`pytest test_a2a_tools_impl.py test_a2a_mcp_server.py\` green
- [ ] Manual: agent calls \`send_message_to_user("Here you go", attachments=["/tmp/foo.zip"])\` → user sees clickable download chip in canvas

**Agent → agent file transfer.** Different design problem: peer's runtime can't download from sender's \`/chat/download\` because cross-workspace auth doesn't exist (peer doesn't have sender's token). Options for follow-up: (1) presigned per-file tokens minted at A2A send time; (2) inline base64 for files under some size cap; (3) extend the A2A proxy to relay file fetches with origin's auth. Tracked under task #114.

### DIFF
--- a/canvas/src/store/canvas-events.ts
+++ b/canvas/src/store/canvas-events.ts
@@ -395,7 +395,25 @@ export function handleCanvasEvent(
 
     case "AGENT_MESSAGE": {
       const content = (msg.payload.message as string) ?? "";
-      if (content) {
+      // Attachments come straight through from the platform's Notify
+      // handler when the agent's tool_send_message_to_user passes file
+      // refs. Shape mirrors NotifyAttachment in activity.go and matches
+      // ChatTab's createMessage(role, content, attachments) signature
+      // exactly, so no adapter needed downstream.
+      const rawAttachments = msg.payload.attachments;
+      const attachments = Array.isArray(rawAttachments)
+        ? (rawAttachments as Array<{ uri?: unknown; name?: unknown; mimeType?: unknown; size?: unknown }>)
+            .filter((a) => typeof a?.uri === "string" && typeof a?.name === "string")
+            .map((a) => ({
+              uri: a.uri as string,
+              name: a.name as string,
+              mimeType: typeof a.mimeType === "string" ? a.mimeType : undefined,
+              size: typeof a.size === "number" ? a.size : undefined,
+            }))
+        : undefined;
+      // Skip when both content and attachments are empty — pure-noise
+      // event we don't want to render as a blank bubble.
+      if (content || (attachments && attachments.length > 0)) {
         const { agentMessages } = get();
         const existing = agentMessages[msg.workspace_id] || [];
         set({
@@ -403,7 +421,12 @@ export function handleCanvasEvent(
             ...agentMessages,
             [msg.workspace_id]: [
               ...existing,
-              { id: crypto.randomUUID(), content, timestamp: new Date().toISOString() },
+              {
+                id: crypto.randomUUID(),
+                content,
+                timestamp: new Date().toISOString(),
+                ...(attachments && attachments.length > 0 ? { attachments } : {}),
+              },
             ],
           },
         });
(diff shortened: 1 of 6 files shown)


HongmingWang-Rabbit:
## 5-axis review

**Correctness:** ✓ Three-layer pipeline holds together cleanly:

- *Python runtime* uploads via `chat/uploads` to get into the canvas's allowed-root list; rejects missing files / non-string paths / empty paths BEFORE notify so a half-rendered chip can't escape.
- *Go handler* adds per-element validation explicitly because `gin` / `go-playground/validator` doesn't iterate slice elements without `dive` — the inline comment captures the real validator gap, not just a defensive flourish.
- *Canvas* mirrors the same drop-empty-uri-or-name filter as defense-in-depth; rendering side runs the existing `extractFilesFromTask` reconstruction path on reload so chips survive page refresh.

`response_body` shape feeds BOTH `extractResponseText` (`.result`) and `extractFilesFromTask` (`.parts[].kind=file`) — pinned by the sqlmock-captured-arg test.

**Tests:** ✓ Comprehensive across all three layers:

- Go: 2 new cases (`PersistsFilePartsForReload` shape, `RejectsAttachmentWithEmpty…` 3-sub-case validator regression). The `sqlmockCaptureArg` helper is well-designed — captures the actual JSON bytes via `driver.Value` while always matching, so the substring asserts in the test body do the real shape-validation (no fake matcher theatre).
- Python: 4 cases (upload+notify happy, missing path, upload-5xx, no-attachments-omits-field). Notify-call lookup by URL suffix instead of array index is a small detail with real value — a future pre-flight call wouldn't silently shift the assert target.

**Architecture:** ✓ Reuses chat-uploads infra + extractFilesFromTask path rather than building a parallel system. `NotifyAttachment` Go struct shape matches frontend's TypeScript shape exactly so no adapter; clean.

**Security:** ✓ No new exfil boundary. Agent could already read arbitrary files via other tools; routing through chat-uploads with the 25MB cap and explicit allowed-roots is the same posture. mimetype is filename-sniffed so no exploit surface from agent-provided extensions.

**Performance:** ✓ Single multipart POST per notify; failed validation rejects before any SQL round-trip.

### FYI (non-blocking)

1. `_upload_chat_files` reads each file fully into memory (`fh.read()`). With the 25MB per-file cap and N files, peak is N×25MB — fine for any realistic batch but worth knowing if someone later wires up batched-large-attachment paths.
2. The disk `fh.read()` inside an async function blocks the event loop. ~50ms on SSD per 25MB file — invisible today, but `aiofiles` or `run_in_executor` would be the answer if it ever shows up in profiling.

PR #2131, opened 2026-04-27
feat(canvas): Agent Comms grouped by peer with sub-tabs

HongmingWang-Rabbit:
## What this fixes

Per the user's feedback after staring at a noodle of 25+ chronological bubbles mixing DD→UX, DD→VD, DD→React Engineer, etc.: Agent Comms should be filterable per peer, not a single firehose.

## What it looks like

\`\`\`
┌─ Agent Comms ──────────────────────────────────────────────┐
│ [ All (15) ] [ React Engineer (5) ] [ Visual Designer (3) ]│  ← peer tabs
│              [ UX Researcher (4) ] [ Deploy (2) ] [ ... ]  │
├────────────────────────────────────────────────────────────┤
│  → To <selected peer>: "Validate audience framings..."     │
│  ← From <selected peer>: "On it!"                          │  ← thread (selected peer only)
│  → To <selected peer>: "Status update?"                    │
└────────────────────────────────────────────────────────────┘
\`\`\`

- **All** sub-tab: pinned leftmost, current chronological view (preserved as default)
- **Per-peer** sub-tab: only that peer's incoming + outgoing thread
- **Order**: most-recent activity DESC, Slack/Linear DM-list style — active conversations rise to top
- **Counts** in parens — total message count per peer
- **Keyboard**: ArrowLeft/Right cycle peers, matches existing My Chat / Agent Comms keyboard nav

## Architecture

| Layer | Change |
|---|---|
| **Pure helper** | New \`buildPeerSummary(messages) → PeerSummary[]\` exported from \`AgentCommsPanel.tsx\`. One pass over messages, returns count + lastTs per peer, sorted by lastTs DESC. |
| **Component** | New \`GroupedCommsView\` wraps the existing message list with \`PeerTabs\` at the top + filtered list below. Auto-prunes the selected peer back to "All" if their messages disappear. |
| **Sub-tab UI** | New \`PeerTabs\` + \`PeerTabButton\` — \`role="tablist"/"tab"\` for screen readers, ArrowLeft/Right keyboard nav. |

## What did NOT change

- Backend: zero touches. Same activity_logs schema, same WS push events.
- \`CommMessage\` shape: unchanged. \`peerId\` + \`peerName\` already carry everything the new UI needs.
- \`toCommMessage\`: unchanged. Same dispatch behavior for every event type (a2a_send, a2a_receive, delegation, DELEGATION_*).

## Tests

5 new tests for \`buildPeerSummary\` (pure function — no React render):
- count aggregation per peer
- most-recent-first ordering across peers
- lastTs is the max timestamp, not the last-seen
- empty input → empty array
- peerName stability when different names land for same peerId

8 existing \`toCommMessage\` tests still pass. All 13 green.

## Test plan

- [x] \`tsc --noEmit\` clean
- [x] \`vitest run AgentCommsPanel.test.ts\` 13/13 green
- [ ] Manual on staging: open Director with multiple peers, see tabs populate; click peer tab → only that peer's thread; ArrowLeft/Right cycle works; live WS push to peer X auto-bumps that peer's tab to most-recent

### DIFF
--- a/canvas/src/components/tabs/chat/AgentCommsPanel.tsx
+++ b/canvas/src/components/tabs/chat/AgentCommsPanel.tsx
@@ -1,6 +1,6 @@
 "use client";
 
-import { useState, useEffect, useRef } from "react";
+import { useState, useEffect, useMemo, useRef } from "react";
 import ReactMarkdown from "react-markdown";
 import remarkGfm from "remark-gfm";
 import { api } from "@/lib/api";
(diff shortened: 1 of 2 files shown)


HongmingWang-Rabbit:
## 5-axis review

**Correctness:** ✓ `buildPeerSummary` extracted as a pure function — single pass over messages, max-timestamp tracking, sort by recency DESC. `ALL_PEERS = "__all__"` won't collide with UUID peerIds (workspace IDs are UUIDv4). Auto-prune `useEffect` correctly falls back to `ALL_PEERS` if the selected peer disappears (rare dedupe case). Arrow-key cycling has proper wraparound via modulo.

**Tests:** ✓ 5 cases on the pure helper (count aggregation, recency ordering, max-timestamp tracking, empty input, name-stability). Helper is the part with non-trivial logic; isolating it for unit testing is the right call. The `as never` test-fixture cast is a minor TS hack, documented inline — acceptable.

**Architecture:** ✓ `buildPeerSummary` exported alongside `toCommMessage` follows the same testability pattern. `PeerTabs` + `PeerTabButton` extracted as separate components instead of inline JSX — composable, scannable. `tabIndex={active ? 0 : -1}` + `role="tablist"` matches WAI-ARIA tablist contract.

**Security:** N/A — pure UI rendering; peer names come from already-validated message data.

**Performance:** ✓ Two memoized derivations (`peerSummary` on `messages`, `visible` on `messages + selectedPeerId`) keep render cost bounded. O(n) on messages count; with the existing 50-message LIMIT in the fetch path, this is trivial.

### Optional (non-blocking)

1. `GroupedCommsView`'s click + keyboard + auto-prune behavior has no render-level test. A future regression in tab-cycling or selected-peer-cleanup wouldn't be caught by the current pure-helper tests. Adding a `@testing-library/react` render test would lock the interactive surface — worth a follow-up but not blocking.
2. The `whitespace-nowrap` + horizontal-scroll on `PeerTabs` handles long peer names without wrapping, but a peer with a >40-char name + 10 peers could push the active tab off-screen. Tiny edge; ignore unless someone hits it.

PR #2132, opened 2026-04-27
test(comms): E2E + canvas coverage for agent → user attachments

HongmingWang-Rabbit:
## Summary
- Adds shell E2E `tests/e2e/test_notify_attachments_e2e.sh` (14 assertions) that hits a live platform on :8080. Pure-platform — no workspace container/LLM key needed — runs in <2s and pins the wire contract behind PR #2130's `send_message_to_user(attachments=...)`.
- Adds 5 canvas-side AGENT_MESSAGE tests for the WS-broadcast filtering (drops empty uri/name, allows attachments-only bubbles, ignores non-array, no-ops on empty).
- Adds 1 message-parser test pinning the persisted `{result, parts}` shape so reload restores both caption and download chips.
- Wires the new shell E2E into `e2e-api.yml` so the contract regresses in CI, not just on dev runs.

Why now: user explicitly asked to "keep optimizing and comprehensive e2e testings to prove all works as expected" for the communication path, and the agent → user attachments path (#2130) is the most user-visible thing we shipped this week.

## Test plan
- [x] Local: `bash tests/e2e/test_notify_attachments_e2e.sh` → 14/14 pass against a freshly-rebuilt platform image
- [x] Local: `npx vitest run` → 1053/1053 pass (was 1047, added 6)
- [x] Verified the test catches a regression: an older platform binary (pre-#2130) failed Test 3 (per-element validation), confirming the test is wire-real, not just shape-checking
- [ ] CI: `e2e-api.yml` should run the new step on this PR

### DIFF
--- a/.github/workflows/e2e-api.yml
+++ b/.github/workflows/e2e-api.yml
@@ -97,6 +97,8 @@ jobs:
           echo "Migrations OK"
       - name: Run E2E API tests
         run: bash tests/e2e/test_api.sh
+      - name: Run notify-with-attachments E2E
+        run: bash tests/e2e/test_notify_attachments_e2e.sh
       - name: Dump platform log on failure
         if: failure()
         run: cat workspace-server/platform.log || true
--- a/canvas/src/components/tabs/chat/__tests__/message-parser.test.ts
+++ b/canvas/src/components/tabs/chat/__tests__/message-parser.test.ts
@@ -262,4 +262,36 @@ describe("extractFilesFromTask", () => {
     const files = extractFilesFromTask(task);
     expect(files[0]).toMatchObject({ name: "out.txt", uri: "workspace:/workspace/out.txt" });
   });
+
+  it("hydrates a notify-with-attachments response_body — both text caption AND file chips", () => {
+    // Pins the exact wire shape the platform's Notify handler persists
+    // when send_message_to_user passes attachments (activity.go writes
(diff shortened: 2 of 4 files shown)


HongmingWang-Rabbit:
## 5-axis review

**Correctness:** ✓ Shell E2E exercises the full wire contract: notify-without-attachments, notify-with-attachments→`parts[].kind=file` persistence, per-element empty-uri/name 400 (locks in the explicit loop guarding against gin's no-`dive` slice-element bug), missing-message 400, and a real `/chat/uploads` → `/notify` round-trip. Cleanup via `trap EXIT`.

**Tests:** ✓ This IS the test PR — 14 shell assertions + 5 canvas `AGENT_MESSAGE` filter tests + 1 message-parser hydration test. Coverage matches the surface added in #2130 (validator gap, broadcast filter, reload path).

**Architecture:** ✓ Reuses existing `_lib.sh` + `e2e_mint_test_token` helpers; wired into `e2e-api.yml` after `test_api.sh` so it runs in the same job's already-stood-up server. <2s additional CI cost.

**Security:** ✓ Test workspace is created+deleted within the test; `mint_test_token` endpoint gated to `MOLECULE_ENV != production`; no secrets logged.

**Performance:** ✓ Negligible (~2s).

LGTM. Auto-merge armed (PR was already APPROVED + CLEAN — this just enables the merge gate to fire on green).

Issue #117, opened 2026-04-15
ops(template-drift): Security Auditor cron in org.yaml (twice-daily) doesn't match live DB (hourly)


HongmingWang-Rabbit:
## Summary

The Security Auditor workspace was deployed from `org-templates/molecule-dev/org.yaml`, but its live cron schedule has diverged from the template. The template defines a twice-daily audit; the live DB runs it hourly with a substantially richer prompt.

## Diff

| | org.yaml | Live DB |
|---|---|---|
| `cron_expr` | `7 6,18 * * *` (twice daily) | `17 * * * *` (hourly) |
| Schedule name | "Security audit (every 12h)" | "Hourly security audit" |
| Prompt | ~12 lines (basic git+gosec+bandit+PR review) | ~60 lines (adds DAST, teardown protocol, GH issue filing, PM routing mandate) |
| Source | `template` | `template` |

## Impact

1. **Redeployment drift**: If the org is re-imported from `org.yaml` (e.g. during a disaster recovery), the Security Auditor reverts to twice-daily with the stripped-down prompt, losing the DAST teardown and PM-routing steps that were added to address prior incidents.
2. **Audit deficit amplified**: The live hourly schedule is already causing a ~36% miss rate (see related issue). The org.yaml version would halve the pressure — but both have a busy-collision risk.

## Proposed Fix

Update `org-templates/molecule-dev/security-auditor/system-prompt.md` _and_ the `schedules:` block in `org.yaml` to reflect the evolved live prompt, then sync cron expressions so template re-import produces the same schedule that's currently running.

## Related

- #115 (Security Auditor busy-collision miss rate — filed today)

HongmingWang-Rabbit:
## Triage (2026-04-15 overnight tick)

Confirmed template drift. The runtime schedule (richer prompt, hourly cron) reflects what the platform actually audits; the org.yaml line is stale.

## Direction of truth

The answer depends on the \`source\` column on \`workspace_schedules\` (added by #76). If the live row has \`source='runtime'\`, it was manually edited after org/import and is protected by the additive re-import guarantee — the template-source row in YAML would only affect a fresh workspace. If it's \`source='template'\`, then the re-import path would have updated it — meaning someone edited it directly in the DB without updating the template, and the next re-import would overwrite.

Can you run this query?

\`\`\`sql
SELECT id, name, cron_expr, source, run_count, last_status
FROM workspace_schedules
[...]

PR #2137, opened 2026-04-27
feat(runtime): RuntimeCapabilities dataclass — primitive #1 of 6

HongmingWang-Rabbit:
## Summary

Foundation for the native+pluggable runtime principle (task #117, blocks #87). Lets each adapter declare which cross-cutting capabilities it owns natively versus delegates to the platform's fallback.

**Pure additive — zero behavior change.** Every existing adapter inherits \`BaseAdapter.capabilities()\` returning \`RuntimeCapabilities()\` with every flag False, matching today's "platform owns everything" behavior exactly. Subsequent PRs land the platform-side consumers (idle-timeout override, scheduler skip, status-transition hook, etc.) one capability at a time, without touching the dataclass.

## Why this PR is small on purpose

The audit (per project memory \`project_runtime_native_pluggable.md\`) found **7 of 12 platform capabilities have no adapter override hook**. We need to land the abstraction layer before refactoring claude-code/hermes executors out of \`workspace/\` (task #87) — otherwise the refactor would need to be touched again every time a capability hook ships.

This PR is the smallest possible foundation: the dataclass + the default method. Following PRs add one consumer at a time.

## What's next (sequenced)

| # | Capability | Platform consumer it unblocks |
|---|---|---|
| 2 | \`provides_native_heartbeat\` | \`/registry/heartbeat\` cadence skip |
| 3 | \`provides_native_scheduler\` | \`scheduler/scheduler.go\` polling skip |
| 4 | \`provides_native_session\` | \`a2a_queue.go\` enqueue skip |
| 5 | \`provides_native_status_mgmt\` | \`registry.go\` status-transition hook |
| 6 | \`provides_activity_decoration\` | \`activity.go\` row decoration merge |
| 7 | \`provides_channel_dispatch\` | \`channels/manager.go\` skip |

Plus an \`idle_timeout_override()\` adapter hook (separate from the flags) for \`a2a_proxy.go:applyIdleTimeout\`.

## Test plan
- [x] 9 new vitest cases in \`test_runtime_capabilities.py\`: dataclass shape (frozen, all-defaults-false), wire-name stability for Go consumer, BaseAdapter default returns all-False, subclass override works, sibling adapters unaffected
- [x] Full workspace pytest: 1300/1300 pass (was 1291, +9 — exact match)
- [x] No production code path changed; default capabilities() return is all-False = today's behavior

### DIFF
--- a/workspace/adapter_base.py
+++ b/workspace/adapter_base.py
@@ -35,6 +35,83 @@ class AdapterConfig:
     heartbeat: Any = None                   # HeartbeatLoop instance
 
 
+@dataclass(frozen=True)
+class RuntimeCapabilities:
+    """Adapter-declared ownership of cross-cutting platform capabilities.
+
+    The platform provides FALLBACK implementations of heartbeat, cron,
+    durable session, etc. When a runtime SDK provides one of these
+    natively (e.g. claude-code's streaming session model, hermes-agent's
+    sidecar lifecycle), the adapter sets the corresponding flag to True.
+    The platform reads these flags and skips its fallback for that
+    capability — the adapter is responsible instead.
+
+    Observability is NEVER skipped: A2A protocol, activity_logs, and the
+    broadcaster always run regardless of who owns the capability. These
+    flags only switch WHO IMPLEMENTS the behavior, not whether the
+    platform sees it.
+
+    All defaults are False so introducing this dataclass is a no-op:
+    every existing adapter inherits BaseAdapter.capabilities() which
+    returns RuntimeCapabilities() with everything off, matching today's
+    "platform does it all" behavior. Each capability gets a platform-
+    side consumer in a follow-up PR; this class is the foundation.
+
+    See project memory `project_runtime_native_pluggable.md` for the
+    architecture principle these flags encode.
+    """
+    # Heartbeat — adapter sends its own keep-alive signal to the platform's
+    # broadcaster instead of relying on workspace/heartbeat.py's 30s loop.
+    # Set True when the SDK already maintains a long-lived session that
+    # produces natural progress events (e.g. claude-code streaming).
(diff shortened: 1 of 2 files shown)


HongmingWang-Rabbit:
## 5-axis review

**Correctness:** ✓ Pure-additive primitive — frozen dataclass with 7 capability flags, all defaulting to False. `BaseAdapter.capabilities()` returns `RuntimeCapabilities()` so every existing adapter inherits the all-False default = today's "platform owns everything" behavior, byte-for-byte.

**Tests:** ✓ 8 cases, well-targeted:
- `test_is_frozen` — guards against runtime mutation that'd let an adapter flip capabilities mid-flight
- `test_all_defaults_false` — explicitly catches the catastrophic case where a default flips to True (silently moves capability ownership for langgraph/crewai/deepagents which have no native impl)
- `test_to_dict_keys_are_stable_wire_names` — pins the Python→Go boundary so a refactor renaming Python fields doesn't break the heartbeat payload's JSON shape
- `test_subclass_can_override_capabilities` — the override mechanism that "primitive #1 of 6" enables

**Architecture:** ✓ Clean separation — Python declares, Go consumes via stable wire names. Documented as "primitive #1 of 6" with explicit incremental rollout (each capability gets its own platform-side consumer in a follow-up). Splitting like this matches the skill's preferred *Stack* strategy for layered changes.

**Security:** N/A — pure config dataclass, no auth/data boundary changes.

**Performance:** ✓ `capabilities()` called at most per heartbeat (~30s cadence); frozen-dataclass instantiation is negligible.

### Optional (non-blocking)

- `capabilities()` returns a fresh instance each call. `@functools.cache` would memoize per-adapter-instance for free, but the cost is so low that caching would be YAGNI for now. Worth revisiting if profiling later shows it on a hot path.

PR #2141, opened 2026-04-27
feat(runtime): native_status_mgmt skip — primitive #4 of 6

HongmingWang-Rabbit:
## Summary

Capability primitive #4 of 6 (task #117). When an adapter declares \`provides_native_status_mgmt=True\` (because its SDK reports its own ready/degraded/failed state), the platform's error-rate-based status inference fights the adapter's own state machine. This PR gates the inference branches on the capability flag — adapter-driven transitions become authoritative.

## Components

- \`registry.go evaluateStatus\` — the two inferred-status branches (degrade on \`error_rate >= 0.5\`; recovery on \`error_rate < 0.1 && runtime_state == \"\"\`) now check \`runtimeOverrides.HasCapability(\"status_mgmt\")\` before firing.
- The wedged-branch (\`RuntimeState == \"wedged\"\` → degraded) is **NOT** gated. That path is the adapter's own self-report, not platform inference, and stays active under \`native_status_mgmt\` — adapters can still drive transitions via \`runtime_state\`.

## Python side

No change. Capability map serialization landed in PR #2137; heartbeat carries it via PR #2139. An adapter setting \`RuntimeCapabilities(provides_native_status_mgmt=True)\` automatically flows through.

## Tests (3 new)

- **SkipsDegradeInference** — \`error_rate=0.8\` + \`currentStatus=online\` + native flag set → degrade UPDATE does NOT fire (sqlmock fails on unexpected query, which is the regression cover)
- **SkipsRecovery** — \`error_rate=0.05\` + \`currentStatus=degraded\` + native → recovery UPDATE does NOT fire
- **WedgedStillRespected** — \`runtime_state=\"wedged\"\` + native → wedged branch DOES fire (adapter self-report stays active)

## Verification

- [x] All Go handlers tests pass (3 new + existing)
- [x] 1308/1308 Python pytest pass (unchanged)
- [x] \`go build ./...\` + \`go vet ./...\` clean

### DIFF
--- a/workspace-server/internal/handlers/native_status_mgmt_test.go
+++ b/workspace-server/internal/handlers/native_status_mgmt_test.go
@@ -1,0 +1,173 @@
+package handlers
+
+import (
+	"bytes"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	sqlmock "github.com/DATA-DOG/go-sqlmock"
+	"github.com/gin-gonic/gin"
+)
+
+// TestHeartbeat_NativeStatusMgmt_SkipsDegradeInference validates capability
+// primitive #4: when an adapter declares native_status_mgmt, the platform's
+// error-rate-based status inference DOES NOT fire. Adapter owns the
+// transition; platform observes only. The wedged-branch (RuntimeState ==
+// "wedged") is NOT gated — it's the adapter's own self-report, not an
+// inference, and stays active.
+//
+// Mirrors the structure of TestHeartbeatHandler_Degraded but pre-populates
+// the runtimeOverrides cache with status_mgmt=true and asserts the degrade
+// UPDATE is NOT issued (so sqlmock's expectations don't include it).
+func TestHeartbeat_NativeStatusMgmt_SkipsDegradeInference(t *testing.T) {
+	mock := setupTestDB(t)
+	setupTestRedis(t)
+	broadcaster := newTestBroadcaster()
+	handler := NewRegistryHandler(broadcaster)
+
+	// Pre-populate the override cache so the workspace under test has
+	// declared native_status_mgmt. Reset after so we don't pollute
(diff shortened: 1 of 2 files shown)


HongmingWang-Rabbit:
## 5-axis review

**Correctness:** ✓ Two specific branches gated (error_rate≥0.5 degrade + recovery), wedged-branch deliberately NOT gated. The comment block enumerates each:
- gated: `error_rate >= 0.5` degrade + `error_rate < 0.1 && runtime_state == ""` recovery
- preserved: wedged (adapter's own self-report, not inference)

Single `runtimeOverrides.HasCapability(workspaceID, "status_mgmt")` lookup feeds both `if !nativeStatus &&` gates — clean.

**Tests:** ✓ 3 cases including the regression-locked **negative-assertion** pattern: `mock.ExpectationsWereMet()` fails if the gated UPDATE accidentally fires (since we don't `ExpectExec` it). That's the right shape for "this thing must NOT happen". `SetCapabilities + defer Reset` cleanly isolates per-test state.

**Architecture:** ✓ Single decision point, two consumers. Inline comment cross-references task #117 and the project memory — keeps the next maintainer oriented in the capability-primitives series. Builds on the foundation in #2137 + the cache from #2139 (and presumably #2140 for primitive #3).

**Security:** N/A — pure status-inference logic; no auth/data boundary.

**Performance:** ✓ One additional `sync.Map` read per heartbeat (~30s cadence) — trivial cost.

### Optional (non-blocking)

- The wedged-branch test (`TestHeartbeat_NativeStatusMgmt_WedgedStillRespected`) is the most subtle of the three — it's verifying ABSENCE of regression in the unrelated path. Worth its weight; mention only because the test name conveys "defense-in-depth assertion" rather than a behavior change. The current naming is fine.

PR #2155, opened 2026-04-27
feat(preflight): replace SUPPORTED_RUNTIMES static list with adapter discovery

HongmingWang-Rabbit:
## Summary

Closes task #123 — last piece of #87 cleanup independent of the cli_executor.py template moves.

**Pre-fix:** `workspace/preflight.py:11` hardcoded a tuple of \"supported\" runtime names. Every new template repo required a code change in molecule-runtime to be recognized — direct violation of the universal-runtime principle (#87).

**Post-fix:** discovery-based validation via the same \`ADAPTER_MODULE\` env var that production load paths already use (\`workspace/adapters/__init__.py:get_adapter\`). Distinguished failure modes so operator messages are concrete:

| Mode | Detail |
|---|---|
| \`ADAPTER_MODULE\` unset | \"no adapter installed; set the env var\" |
| Module won't import | Underlying ImportError type + message |
| No \`Adapter\` class | \"Add `Adapter = YourClass` per BaseAdapter convention\" |
| \`Adapter.name()\` raises | \"name() raised <type>: <msg>\" — must be side-effect-free |
| \`Adapter.name()\` returns non-string | \"must return a non-empty string\" |
| \`Adapter.name() != config.runtime\` | **WARNING (not fail)** — adapter wins, config drift surfaced |

## Behavioral change worth calling out

The drift case is a softening: prior static-list path hard-failed config.runtime values not in the allowlist; now drift is a warning so the workspace still boots while operator gets actionable info naming both the configured + installed runtime names. Reflects production reality (the adapter wins regardless).

## Tests (6 new + autouse fixture)

- Replaces obsolete \"static list pass/fail\" tests with 6 cases covering each distinguished failure mode + a positive happy-path test
- Autouse \`_default_langgraph_adapter\` fixture pre-installs a fake adapter via \`sys.modules\` monkey-patching, so existing tests building default WorkspaceConfig (\`runtime=\"langgraph\"\`) inherit a valid adapter
- Failure-mode tests opt out via \`@pytest.mark.no_default_adapter\` (registered in pytest.ini)
- Sentinel pattern (\`_UNSET = object()\`) for \`name_returns\` so None / int / etc. are passable test values

## Verification

- [x] 22/22 preflight tests pass (was 16; +6 new failure-path tests)
- [x] 1256/1256 workspace pytest pass (was 1251; +5 net)
- [x] No production code path other than preflight changed

### DIFF
--- a/workspace/preflight.py
+++ b/workspace/preflight.py
@@ -1,22 +1,82 @@
 """Startup preflight checks for workspace runtime configs."""
 
+import importlib
 import os
 from dataclasses import dataclass, field
 from pathlib import Path
 
 from config import WorkspaceConfig
 
-SUPPORTED_RUNTIMES = {
-    "langgraph",
-    "claude-code",
-    "codex",
-    "ollama",
-    "custom",
-    "crewai",
-    "autogen",
-    "deepagents",
-    "openclaw",
-}
+
+def _validate_runtime_via_adapter(runtime: str) -> tuple[bool, str]:
+    """Discover the installed adapter and confirm it matches the
+    config's `runtime` field. Returns (ok, detail) — detail is the
+    operator-actionable failure message when ok is False.
+
+    Replaces the previous hardcoded SUPPORTED_RUNTIMES allowlist
+    (claude-code / codex / ollama / langgraph / etc.). The static list
+    couldn't keep up with new template repos: each new adapter required
+    a code change in molecule-runtime to be 'supported', a violation of
+    the universal-runtime principle (#87).
+
+    Discovery uses the same ADAPTER_MODULE env var that production load
+    paths consult (workspace/adapters/__init__.py:get_adapter). The
+    adapter's static name() string is the source of truth — config.yaml
+    just labels which one the operator expects, and the check warns on
+    drift.
+
+    Failure modes the function distinguishes (each gets a distinct
+    operator-facing message so debugging is concrete):
+      - ADAPTER_MODULE unset → "no adapter installed"
+      - ADAPTER_MODULE set but module won't import → "import failed: …"
+      - module imports but no Adapter class → "Adapter class missing"
+      - Adapter.name() differs from config.runtime → drift warning
+    """
+    adapter_module = os.environ.get("ADAPTER_MODULE", "").strip()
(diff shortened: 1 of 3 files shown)


HongmingWang-Rabbit:

> Note: posting as a comment because gh's active account is also the PR author. The user's other account will need to formally approve.

**Correctness:** ✓
- Each failure mode (unset / unimportable / no Adapter class / name() raises / non-string name() / drift) gets a distinguished operator-facing message — debugging is concrete, not a single generic "unsupported"
- Drift case correctly returns `(True, detail)` so the dispatch in `run_preflight` puts it in warnings instead of failures
- Empty-string detail on success skips the warning branch — clean signal/no-noise

**Tests:** ✓
- 6 new failure-path tests, each pinned to the exact substring of the operator message — a future refactor that drops the actionable detail will fail loudly
- `_UNSET = object()` sentinel lets `None` be a passable test value for `name_returns` — exactly the bug a `if name_returns is not None` check would have caused
- Autouse fake-adapter fixture + `@pytest.mark.no_default_adapter` opt-out keeps existing tests untouched while exercising failure paths cleanly

**Architecture:** ✓
- Aligns with #87 universal-runtime principle (cited in body) — adapter is now source of truth, config.yaml is just a label
- Reuses the same `ADAPTER_MODULE` env convention as `workspace/adapters/__init__.py:get_adapter` — consistent pattern, no new contract

**Security:** ✓
- `importlib.import_module(env_var)` is arbitrary-module-loading by design, but `ADAPTER_MODULE` is image-build-time config (Dockerfile ENV), not user input
- Same trust boundary as the existing production load path

**Performance:** ✓ — preflight runs once at startup; one module import is negligible

---

**Nit:** `install_fake_adapter` uses `id(monkeypatch)` in the module name to avoid sys.modules collisions across tests, but pytest's monkeypatch already restores sys.modules at teardown. The unique-name suffix is defensive overkill — could just be `f"_fake_adapter_{name.replace('-', '_')}"`. Not worth blocking on; current code is correct, just slightly more machinery than needed.

**Verdict:** would approve. Auto-merge already armed; will fire once CI greens + an approving review lands.

Issue #118, opened 2026-04-15
ops(scheduler): Technical Researcher schedule has stale next_run_at and anomalously low run_count


HongmingWang-Rabbit:
## Summary

The Technical Researcher's "Hourly plugin curation" schedule (`22 * * * *`) shows signs of a scheduler anomaly: `next_run_at` is stale (past due) and `run_count=1` despite the schedule being active for ~3.5 hours — during which 3–4 runs were expected.

## Evidence (live DB — 2026-04-15T07:34 UTC)

| Field | Value |
|---|---|
| Created | 2026-04-15T03:36:57Z |
| `last_run_at` | 2026-04-15T07:03:36Z |
| `next_run_at` | 2026-04-15T07:22:00Z (12 min overdue at audit time) |
| `run_count` | 1 |
| Expected runs since creation | ~3–4 (fires at minute 22: 04:22, 05:22, 06:22, 07:22) |
| **Miss rate** | **~67–75%** |

Additional anomaly: `last_run_at` is 07:03:36 (minute 3) but `cron_expr` is `22 * * * *` (minute 22). The one recorded run does not align with the cron expression.

## Hypotheses

1. **Scheduler not advancing `next_run_at`** after a missed or errored fire → the slot silently passes, next_run_at stays frozen until the next successful run.
2. **Schedule was replaced/re-created** shortly before 07:03, resetting run_count to 0, then immediately triggered manually → explains the minute-3 run and count=1.
3. **Related to #85** — the scheduler resurrected for this workspace after a crash period, and missed 3 prior slots.

## Impact

Plugin curation — one of the four required evolution levers — may be running at ~25–33% of its intended frequency.

## Requested Action

- Inspect scheduler logs around 04:22, 05:22, 06:22 for workspace `57e13b54-119f-4f5b-9e33-0a6c6577477d`.
- Confirm whether `next_run_at` is being advanced correctly after each fire regardless of success/error.

## Related

- #85 (scheduler death)
- #115 (Security Auditor busy-collision)

HongmingWang-Rabbit:
## Triage (2026-04-15 overnight tick)

This looks like the same underlying problem as #115, but for the Technical Researcher: the cron fires correctly, but the workspace is busy when it arrives, so \`run_count\` falls behind \`next_run_at\` and misses accumulate.

[...]

PR #2158, opened 2026-04-27
feat(tools): tighten send_message_to_user description to forbid pasting URLs in body

HongmingWang-Rabbit:
## Summary

Root-cause fix for **#118** (chat attachments rendering as plain text links instead of download chips). User flagged with screenshot 2026-04-26 showing the Design Director agent pasting \`https://files.catbox.moe/…\` in the message body — chat rendered the URL as plain markdown text, unclickable in the canvas's bubble layout, and unreachable in any SaaS deployment where the user's browser can't egress to catbox.

The structured \`attachments\` field already exists, the canvas's \`AttachmentChip\` already renders well, the WebSocket broadcast already carries attachments verbatim — the missing piece was the LLM choosing the body over the structured field. Tighten the tool description so it trains the right behavior.

## Three targeted strengthenings

1. **Top-level tool description** — enumerated use case (4) now reads \"via the \`attachments\` field (NEVER paste file URLs in \`message\`)\".
2. **\`message\` param** — adds DO NOT rule with the SaaS-reachability rationale so operators can grep for \"SaaS\" and find this design constraint. Names \`catbox.moe\` + \`file://\` as concrete examples of forbidden hosts (the two we've seen in production).
3. **\`attachments\` param** — leads with REQUIRED; lists bad alternatives explicitly (pasting URLs, base64-encoding, telling user to look at a path). LLMs handle \"use X, NOT Y\" framings better than \"use X\" alone.

## Tests (4 new)

Pin all three load-bearing phrases in \`test_a2a_mcp_server.py\` so a future doc edit that softens or drops them fails CI. Brittle by design — these are prompt-engineering invariants, not implementation details.

## Verification

- [x] 1190/1190 workspace pytest pass
- [x] 4 new send_message_to_user description tests all green

## Follow-up

A defensive canvas-side backstop (auto-detect download-shaped URLs in body, convert to chips) is a follow-up if steering proves insufficient. Doesn't ship in this PR.

### DIFF
--- a/workspace/a2a_mcp_server.py
+++ b/workspace/a2a_mcp_server.py
@@ -113,17 +113,44 @@ TOOLS = [
     },
     {
         "name": "send_message_to_user",
-        "description": "Send a message directly to the user's canvas chat — pushed instantly via WebSocket. Use this to: (1) acknowledge a task immediately ('Got it, I'll start working on this'), (2) send interim progress updates while doing long work, (3) deliver follow-up results after delegation completes, (4) attach files (zip, pdf, csv, image) for the user to download. The message appears in the user's chat as if you're proactively reaching out.",
+        "description": "Send a message directly to the user's canvas chat — pushed instantly via WebSocket. Use this to: (1) acknowledge a task immediately ('Got it, I'll start working on this'), (2) send interim progress updates while doing long work, (3) deliver follow-up results after delegation completes, (4) attach files (zip, pdf, csv, image) for the user to download via the `attachments` field (NEVER paste file URLs in `message`). The message appears in the user's chat as if you're proactively reaching out.",
         "inputSchema": {
             "type": "object",
             "properties": {
                 "message": {
                     "type": "string",
-                    "description": "The message to send to the user. Required even when sending attachments — set to a short caption like 'Here's the build:' or 'Done — see attached.'",
+                    # The "no URLs in message text" rule is the single biggest
+                    # cause of bad chat UX: agents drop catbox.moe / file://
+                    # / temporary upload-host links into the prose, the
+                    # canvas renders them as plain markdown links the user
+                    # can't preview, and SaaS deployments often can't even
+                    # reach those external hosts. Every download MUST go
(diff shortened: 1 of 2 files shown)


HongmingWang-Rabbit:

> Note: posting as a comment because gh's active account is the PR author.

**Correctness:** ✓
- Pure prompt-steering / tool-description update. No code-flow change. Behavior change is in LLM runtime usage of `attachments` vs URL-pasting in `message`
- Inline `# comment block` inside the dict literal between `"type"` and `"description"` is valid Python; reads cleanly as inline rationale for the next reader

**Tests:** ✓
- 3 tests pin specific load-bearing phrases against the production TOOLS dict — exactly the right granularity (no string-extracting indirection through helpers)
- Each assertion ships with a docstring/message explaining the regression class it catches (e.g. "Pre-#118 the description omitted this rule and agents routinely shipped catbox.moe / file:// links in chat"). A future doc-edit PR that softens the language fails loudly with a meaningful error
- Caps-lock pinning is empirically motivated per the body — defensible prompt-engineering signal worth preserving

**Architecture:** ✓ — no flow change; tests live alongside the module they pin

**Security:** ✓ — N/A, description text

**Performance:** ✓ — N/A, module-load static dict

---

**Optional / FYI:** Two thoughts, neither blocking:

1. The "DO NOT paste URLs" rule could in principle be enforced *defensively* with a regex check in the `send_message_to_user` server-side handler that warns (or strips) URL-shaped tokens out of `message` when `attachments` is non-empty. Would catch agents that ignore the prompt rule. But that's #118-followup territory — out of scope here.

2. The `attachments` description lists three bad alternatives ("pasting URLs", "base64-encoding", "telling the user to look at a path"). If/when a fourth bad pattern shows up in the wild, growing this inline list past 3-4 items will start hurting LLM token economy. At that point worth migrating to a dedicated `## ANTIPATTERNS` section in a longer description block. Not a problem yet.

**Verdict:** would approve. Auto-merge already armed; will fire once an approving review lands + CI greens.

PR #2161, opened 2026-04-27
feat(publish-runtime): auto-publish to PyPI on staging pushes touching workspace/

HongmingWang-Rabbit:
## Summary
- New trigger: push to \`staging\` filtered to \`paths: ['workspace/**']\` auto-publishes a fresh \`molecule-ai-workspace-runtime\` patch release
- Auto-derived version: query PyPI for current latest, bump patch component
- \`concurrency: group: publish-runtime\` serializes parallel staging merges so they don't race on the bump
- \`publish\` job now exposes \`version\` as a job-level output; cascade reads it cleanly (also fixes a latent bug where the cascade re-derived from \`GITHUB_REF_NAME\` and would have produced \"staging\" as the version under the new trigger)

## Why
Tonight's RuntimeCapabilities ImportError outage. Sequence:
1. #117 added \`RuntimeCapabilities\` to \`molecule_runtime.adapters.base\` (merged 02:37 UTC)
2. Templates rebuilt at 07:37 UTC and started importing the new symbol
3. PyPI was still serving 0.1.15 (pre-#117) — nobody remembered to push a \`runtime-v*\` tag or workflow_dispatch the publish
4. Every template image now runs \`from molecule_runtime.adapters.base import RuntimeCapabilities\` against an installed runtime that doesn't export it → ImportError → workspaces stuck in \`provisioning\`

This change closes the human-in-loop gap. Pairs with [molecule-ci PR #7](https://github.com/Molecule-AI/molecule-ci/pull/7) which adds an actual \`docker run ... import adapter\` smoke gate before pushing :latest.

## Test plan
- [x] YAML validates
- [ ] CI green
- [ ] After merge: confirm next workspace/** change to staging fires \`publish-runtime\` automatically with version = \`pypi-latest + 1\`
- [ ] Confirm cascade fan-out reads \`needs.publish.outputs.version\`

### DIFF
--- a/.github/workflows/publish-runtime.yml
+++ b/.github/workflows/publish-runtime.yml
@@ -10,6 +10,12 @@ name: publish-runtime
 #     the tag — `runtime-v0.1.6` publishes `0.1.6`).
 #   - Manual workflow_dispatch with an explicit `version` input (useful for
 #     dev/test releases without tagging the repo).
+#   - Auto: any push to `staging` that touches `workspace/**`. The version
+#     is derived by querying PyPI for the current latest and bumping the
+#     patch component. This closes the human-in-loop gap that caused the
+#     2026-04-27 RuntimeCapabilities ImportError outage — adapter symbol
+#     additions in workspace/adapters/base.py used to require an operator
+#     to remember to publish; now the merge itself triggers the publish.
 #
 # The workflow:
 #   1. Runs scripts/build_runtime_package.py to copy workspace/ →
@@ -32,6 +38,12 @@ on:
   push:
     tags:
       - "runtime-v*"
+    branches:
+      - staging
+    paths:
+      # Auto-publish when staging gets workspace/ changes. Path filter
+      # ONLY applies to branch pushes — tag pushes still fire regardless.
+      - "workspace/**"
   workflow_dispatch:
     inputs:
       version:
@@ -42,6 +54,12 @@ on:
 permissions:
   contents: read
 
+# Serialize publishes so two staging merges landing seconds apart don't
+# both compute "latest+1" and race on PyPI upload. The second one waits.
+concurrency:
+  group: publish-runtime
+  cancel-in-progress: false
+
 jobs:
   publish:
     runs-on: ubuntu-latest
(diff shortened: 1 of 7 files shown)


HongmingWang-Rabbit:
## Code review (5-axis) — non-blocking notes

Posting as comment (gh active account = PR author). PR is already APPROVED, CI green, merge state CLEAN — arming auto-merge after this comment.

**Correctness:** ✓ on both halves
- Workflow: `concurrency: group: publish-runtime` serializes parallel staging merges so they don't both compute `latest+1` and race. The cascade fix (`needs.publish.outputs.version` instead of re-derived `GITHUB_REF_NAME`) closes a real latent bug under the new trigger
- Skill compat: `_normalize_runtime_field` covers all four input shapes (None / string / list / garbage); current_runtime=None preserves pre-existing behavior; tests pin all branches

**Tests:** ✓
- 7 new skill-loader tests pin universal-default, explicit match/mismatch, string sugar, wildcard short-circuit, None preserves old behavior, malformed-treated-as-universal
- Watcher test fake-`load_skills` lambdas updated to accept `**kwargs` — exactly the right shape for "the new kwarg is opt-in, lambdas don't need to know about it"

**Architecture:** ✓ — borrows hermes' declarative compat pattern with universal-default opt-in; existing skill libraries don't need migration

**Security:** ✓ — workflow uses OIDC trusted publisher (no PYPI_TOKEN), skill compat is metadata-only

**Performance:** ✓ — N/A on either change

---

**Optional / FYI (none blocking):**

1. **Bundled scope:** This PR carries two unrelated changes (PyPI auto-publish workflow + per-skill `runtime` compat). Per /code-review-and-quality, refactor + new-behavior is normally two separate PRs. Both are individually small and well-tested so it doesn't *block* merge, but going forward favor splitting unrelated work — easier to revert one half if the other turns out wrong.

2. **PyPI auto-bump under external concurrency (FYI):** `concurrency: group` serializes within this workflow only. A tag push + a manual `workflow_dispatch` racing each other would both re-query PyPI and could collide. Worst case PyPI rejects the second upload (it forbids duplicate versions), and the workflow logs the failure. Acceptable — just noting the boundary so it's documented somewhere.

3. **Curl-failure recovery (Nit):** if PyPI is unreachable, `curl -fsS --retry 3` exits non-zero, the python pipe gets empty input and crashes, and MAJOR/MINOR/PATCH end up empty — `VERSION` becomes `..1` which then trips the PEP 440 regex check and fails loud. The path works but the failure shape is indirect; an explicit early `exit 1` after curl could shorten the diagnostic. Not worth a follow-up unless PyPI starts being flaky.

**Verdict:** approve. Arming auto-merge.

PR #2163, opened 2026-04-27
fix(release): drift-gate TOP_LEVEL_MODULES + smoke-import main (post-0.1.16 incident)

HongmingWang-Rabbit:
## Summary
Two release-pipeline bugs that compounded into the 0.1.16 outage where every workspace startup died with \`ModuleNotFoundError: No module named 'transcript_auth'\`.

### Bug 1: stale TOP_LEVEL_MODULES set
\`scripts/build_runtime_package.py\` rewrites bare imports (\`from foo import bar\` → \`from molecule_runtime.foo import bar\`) using a hand-curated set. The set drifted:

**Missing** (added to workspace/, never added to set → unrewritten imports shipped):
- \`transcript_auth\` (since #87 phase 1c)
- \`runtime_wedge\`
- \`watcher\`

**Stale** (no longer exists in workspace/, still listed):
- \`claude_sdk_executor\` (removed in #87)
- \`cli_executor\` (removed in #87)
- \`hermes_executor\` (never existed)

### Bug 2: wheel smoke didn't import main
\`publish-runtime.yml\`'s \`Verify package contents (sanity)\` step asserted on stable invariants (\`BaseAdapter\`, \`AdapterConfig\`, the a2a error sentinel) but never imported \`main\`. \`main.py\` is where the unrewritten \`from transcript_auth import ...\` lived. Smoke passed; production broke.

## Fix
- Build script auto-asserts \`TOP_LEVEL_MODULES == {f.stem for f in workspace/*.py}\` (excluding \`__init__\`/\`conftest\`). Diff in either direction fails build with a specific error.
- \`TOP_LEVEL_MODULES\` updated to current reality.
- Wheel smoke step now does \`import molecule_runtime.main\` first — main pulls every other module transitively, so any bare-import bug surfaces as \`ModuleNotFoundError\` before PyPI accepts the upload.

## Test plan
- [x] Local build of 0.1.99 succeeds; \`main.py\` correctly contains \`from molecule_runtime.transcript_auth import ...\`
- [ ] CI green
- [ ] After merge, fire publish-runtime workflow_dispatch with version 0.1.18; wheel smoke must pass; PyPI upload succeeds; cascade rebuilds all 8 templates
- [ ] Re-run priority-runtimes wire-real E2E against fresh template images — claude-code + hermes both reach \`online\` and reply non-error to "Reply with PONG"

### DIFF
--- a/.github/workflows/publish-runtime.yml
+++ b/.github/workflows/publish-runtime.yml
@@ -135,6 +135,14 @@ jobs:
           WORKSPACE_ID=00000000-0000-0000-0000-000000000000 \
           PLATFORM_URL=http://localhost:8080 \
             /tmp/smoke/bin/python -c "
+          # Importing main is the strongest smoke test we can do here:
+          # main.py is the entry point and pulls every other module
+          # transitively. If the build script missed an import rewrite
+          # (e.g. left a bare \`from transcript_auth import ...\` instead
+          # of \`from molecule_runtime.transcript_auth import ...\` — the
+          # 0.1.16 incident), this fails with ModuleNotFoundError instead
+          # of shipping to PyPI and breaking every workspace startup.
+          import molecule_runtime.main  # noqa: F401
           from molecule_runtime import a2a_client, a2a_tools
           from molecule_runtime.builtin_tools import memory
           from molecule_runtime.adapters import get_adapter, BaseAdapter, AdapterConfig
--- a/scripts/build_runtime_package.py
+++ b/scripts/build_runtime_package.py
@@ -42,8 +42,13 @@ from pathlib import Path
 # matches one of these) gets rewritten to use the package prefix.
 #
 # Closed list (not "every .py we copy") because a typo in workspace/ would
-# otherwise leak into a wrong rewrite. Update this when adding a new
-# top-level module to workspace/.
+# otherwise leak into a wrong rewrite. The set is asserted against
+# `workspace/*.py` at build time — if the disk contents drift from this
+# list (new module added, old one removed), the build fails loud instead
+# of silently shipping unrewritten imports. That gap caused 0.1.16 to
+# ship `from transcript_auth import ...` (unrewritten — module added
+# without updating this set), which broke every workspace startup with
+# `ModuleNotFoundError: No module named 'transcript_auth'`.
 TOP_LEVEL_MODULES = {
     "a2a_cli",
     "a2a_client",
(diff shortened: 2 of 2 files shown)


HongmingWang-Rabbit:

> Posting as comment (gh active = PR author).

**Correctness:** ✓
- Two complementary safeguards for the same root cause: (a) drift gate at build time asserts `TOP_LEVEL_MODULES` matches `workspace/*.py` reality; (b) smoke-importing `molecule_runtime.main` post-build pulls every transitive import, catching any unrewritten `from foo import` that escaped (a)
- Stale entries removed (claude_sdk_executor, cli_executor, hermes_executor — all confirmed deleted from workspace/ in recent #87 phases); missing entries added (runtime_wedge, transcript_auth, watcher)
- Drift detection is symmetric (both `missing` and `stale` reported separately) — surfaces "you added a module but forgot the set" AND "you deleted a module but left a no-op set entry"
- Excludes `__init__`/`conftest` correctly — neither is a rewrite target

**Tests:** ✓ (implicit, but defensible)
- The drift gate IS the test for "TOP_LEVEL_MODULES set is current" — runs every build
- The smoke-import IS the test for "the rewritten package actually loads" — runs every publish
- No standalone unit test for `build_runtime_package.main()`, but the workflow itself is the integration test and the failure modes are concrete (exit 3 with named missing/stale lists). A unit test would just mock the same filesystem walk

**Architecture:** ✓
- Drift gate lives where the contract lives (the build script itself); smoke test lives where the contract is verified (the publish workflow). Right separation
- Two fixes in one PR is appropriate here — same incident, complementary mechanisms, removing one without the other leaves a gap

**Security:** ✓ — N/A, release pipeline hardening

**Performance:** ✓ — one glob over workspace/*.py at build time, negligible; smoke-import was already in the workflow

---

**Optional / FYI (none blocking):**

1. The drift gate's `{f.stem for f in src.glob("*.py") if f.stem not in {"__init__", "conftest"}}` is exhaustive on `.py` files. If someone later adds `setup.py` or `_version.py` to `workspace/`, they'd need to either add it to `TOP_LEVEL_MODULES` or extend the exclude set. Not a problem now — flagging only because the failure shape (build-fails-with-named-missing) makes the fix obvious if it ever happens.

2. Exit code 3 is new in this script (existing paths use 0/1/2). Fine — but if there are external scripts/CI that interpret these specific codes, they'd want awareness. Quick `grep build_runtime_package` confirms no external consumers checking exit codes today.

**Verdict:** would approve. Auto-merge already armed; will fire once CI greens + an approving review lands.

PR #2198, opened 2026-04-28
docs(ci): correct overstated comments in publish + pin-compat workflows

HongmingWang-Rabbit:
## Summary

Two doc-only fixes from the post-merge review of #2196 + #2197:

1. **`publish-runtime.yml`** — comment claimed "≈ 2 min" but real wall is ~4-5 min (30 attempts × (~3-5s pip install + 4s sleep) on a slow runner). Updated both the comment and the final `::error::` message. Also added a caveat about Fastly content-under-URL staleness (separate from propagation lag — needs a follow-up wheel-hash check).

2. **`runtime-pin-compat.yml`** — comment claimed "byte-identical with publish-runtime.yml's build step" but the temp dirs differ (`/tmp/runtime-build` here vs `${{ runner.temp }}/runtime-build` there). Softened to "same build script with the same args" and noted the smoke import is intentionally narrower than publish's.

## No behavior change

Comment-only edits.

### DIFF
--- a/.github/workflows/publish-runtime.yml
+++ b/.github/workflows/publish-runtime.yml
@@ -318,9 +318,24 @@ jobs:
           python -m venv /tmp/propagation-probe
           PROBE=/tmp/propagation-probe/bin
           $PROBE/pip install --upgrade --quiet pip
-          # Poll budget: 30 attempts × 4s ≈ 2 min. Generous vs PyPI's
+          # Poll budget: 30 attempts × (~3-5s pip install + 4s sleep)
+          # ≈ 4-5 min wall on a slow GH runner. Generous vs PyPI's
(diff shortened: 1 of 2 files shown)


HongmingWang-Rabbit:
Obsoleted by #2199 + #2200 which both substantially restructured the same comment regions:

- The publish-runtime.yml probe block this PR was clarifying got replaced by #2200's wheel-hash check, which has its own accurate comments (~5-6 min budget, content-staleness now actually detected rather than just caveated).
- The runtime-pin-compat.yml byte-identical wording was moved to runtime-prbuild-compat.yml by #2199's per-job-paths split, where the wording is already softened to 'same build script with the same args'.

Both target comments are now correct on staging without this PR.

Issue #1815, opened 2026-04-23
[testing] Canvas (vitest) has no coverage instrumentation — set 70% floor + CI gate


HongmingWang-Rabbit:
## Context

Canvas currently runs 900 tests across 58 files but has **zero coverage instrumentation**:

- `canvas/package.json` "test" script is `vitest run` (no `--coverage`)
- `canvas/vitest.config.ts` has no `coverage` block
- CI has no `vitest --coverage` step

Impact: we discover DOM/event handler bugs in production. Example from today — the ContextMenu Zustand mock silently lost `getState` during a rebase and no test caught it until an E2E failure.

### Step 1 — add coverage instrumentation

Update `canvas/vitest.config.ts`:

```ts
export default defineConfig({
  test: {
    coverage: {
      provider: 'v8',
      reporter: ['text', 'html', 'json-summary'],
      include: ['src/**/*.{ts,tsx}'],
      exclude: [
        'src/**/*.test.{ts,tsx}',
        'src/**/__tests__/**',
        'src/**/*.d.ts',
        'src/types/**',
      ],
      thresholds: {
        lines: 70,
        functions: 70,
        branches: 65,
        statements: 70,
      },
    },
  },
});
```

### Step 2 — add test:coverage script + CI step

`canvas/package.json`:
```json
"test:coverage": "vitest run --coverage"
```

`.github/workflows/ci.yml` — add `npm run test:coverage` after `npm run test` in the Canvas (Next.js) job.

### Step 3 — fix any files currently below threshold

The 70% floor should be achievable based on current test density (900 tests / 58 files), but some files will fall below. Options:
- Write missing tests
- Exclude files that are genuinely not unit-testable (e.g. `next.config.js`) via the `exclude` list

## Acceptance criteria

- [ ] Coverage instrumentation runs locally + in CI
- [ ] Initial threshold set at 70% (lines/functions/statements), 65% (branches)
- [ ] CI fails if coverage drops below threshold
- [ ] Documented in `canvas/README.md`

## Priority

**Medium** — not blocking prod but is the #1 reason we keep hitting regressions that feel like they should've been caught pre-merge.

HongmingWang-Rabbit:
PR #2302 opened — Step 2 of this issue (CI wiring + artifact upload). Step 1 (vitest.config.ts coverage block) already shipped; the inline comment there refs this issue. Step 3 (thresholds + hard gate) deferred until baseline visible.

PR #2302, opened 2026-04-29
ci(canvas): wire vitest --coverage into CI for baseline observability (#1815)

HongmingWang-Rabbit:
## Summary
Step 2 of #1815. Replaces `npx vitest run` with `npx vitest run --coverage` in the `canvas-build` job and uploads the coverage report (HTML + json-summary) as a 7-day artifact. No threshold gate yet — Step 3.

## Why staged
The inline comment in `canvas/vitest.config.ts` (already shipped under Step 1) explicitly defers thresholds because a blind 70% gate would either fail CI immediately or force an ad-hoc exclude list that papers over real gaps. This PR ships the observability so the team can see the baseline before dialing in thresholds.

## Test plan
- [x] `python3 -c 'import yaml; yaml.safe_load(open("..."))'` — yaml valid
- [ ] CI passes on this PR (vitest still runs the same 900 tests; coverage is additive)
- [ ] Coverage artifact appears under Actions → run → canvas-coverage-`<run_id>`
- [ ] No threshold failures (vitest.config.ts has none set)

## Follow-up
Step 3 (set thresholds + flip to a hard gate) becomes a fresh issue once baseline numbers are visible. Issue body proposed lines:70 / functions:70 / branches:65 / statements:70; the actual values may need adjustment.

Closes the Step-2 portion of #1815.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -193,8 +193,24 @@ jobs:
           node-version: '22'
       - run: rm -f package-lock.json && npm install
       - run: npm run build
-      - name: Run tests
-        run: npx vitest run
+      - name: Run tests with coverage
+        # Coverage instrumentation is configured in canvas/vitest.config.ts
+        # (provider: v8, reporters: text + html + json-summary). Step 2 of
+        # #1815 — wires coverage into CI so we get a baseline visible on
+        # every PR. No threshold gate yet; thresholds dial in (Step 3, also
+        # tracked in #1815) after the team sees what current coverage is.
+        # Per the inline comment in vitest.config.ts: "first land
+        # observability so we can see the baseline, then dial in
+        # thresholds + a hard gate" — this PR ships the observability half.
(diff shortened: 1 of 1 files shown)


HongmingWang-Rabbit (commented):
Workflow change (.github/workflows/ci.yml) — comment-only per infra-change policy.

Reviewed and looks clean:
- Additive: replaces `vitest run` with `vitest run --coverage` plus an artifact upload step. No threshold gate yet (Step 2 of #1815 is observability-only, threshold gate is Step 3).
- `actions/upload-artifact` is SHA-pinned (`ea165f8d…` # v4.6.2) — matches the org-wide pinning policy.
- `if: always()` on the upload + `if-no-files-found: warn` keeps a failed test run from masking missing coverage output.
- 7-day retention is reasonable for a baseline-observability stage.

No CI drag impact since coverage instrumentation is already configured in canvas/vitest.config.ts. Approve once a human signs off on the infra category.

PR #2314, opened 2026-04-29
feat(workspace): /internal/chat/uploads/ingest endpoint (RFC #2312, PR-B)

HongmingWang-Rabbit:
## Summary

Stacked on PR #2313 (PR-A). Adds the workspace-side endpoint that PR-C will forward to. **Still no platform-side caller**, so still no behavior visible to users — auth fails closed, the route just sits there waiting for PR-C.

## What this adds

- **\`platform_inbound_auth.py\`** — auth gate, mirrors \`transcript_auth.py\`. Reads \`/configs/.platform_inbound_secret\` (the file PR-A's provisioner writes). Fail-closed when missing/empty/unreadable. \`hmac.compare_digest\` for constant-time compare.
- **\`internal_chat_uploads.py\`** — \`POST /internal/chat/uploads/ingest\`. Multipart parse → \`sanitize_filename\` → write to \`/workspace/.molecule/chat-uploads/<random>-<name>\` with \`O_CREAT|O_EXCL|O_NOFOLLOW\`. Same response shape (\`uri\`/\`name\`/\`mimeType\`/\`size\` + \`workspace:\` URI scheme) as the legacy Go handler so canvas + agent code resolving \`workspace:...\` URIs keeps working.
- **\`workspace/main.py\`** — \`add_route(\"/internal/chat/uploads/ingest\", ingest_handler, methods=[\"POST\"])\` next to the existing \`/transcript\`.
- **\`workspace/requirements.txt\`** — \`python-multipart>=0.0.18\` (Starlette \`Request.form()\` needs it; ≥ 0.0.18 closes CVE-2024-53981).

## Test coverage

| File | Tests | Covers |
|---|---|---|
| \`tests/test_platform_inbound_auth.py\` | 14 | happy path, fail-closed on missing/empty/whitespace-only file, missing/case-wrong/empty Bearer prefix, in-process cache, default \`CONFIGS_DIR\` fallback, end-to-end file → authorized |
| \`tests/test_internal_chat_uploads.py\` | 22 | \`sanitize_filename\` matrix (\`../traversal\`, CJK, length truncation), 401 on missing/wrong/no-secret-file bearer, single + batch upload happy paths, unique random prefix on duplicate names, mimetype guess fallback, 400 on missing files field, 413 on per-file + total-body oversize, **symlink-at-target refusal** (sentinel-content unchanged assertion) |

\`WORKSPACE_ID=test-workspace pytest tests/ --no-cov\` → **1266 passed, 2 xfailed**.

[...]

PR #2322, opened 2026-04-29
ci: collapse Canvas (Next.js) to single job with conditional steps (supersedes #2321)

HongmingWang-Rabbit:
## Summary

Supersedes PR #2321's two-jobs-sharing-a-name approach, which **didn't actually clear branch-protection's required-check evaluation**.

Live test on PR #2314 after #2321 merged: GraphQL \`isRequired\` confirmed BOTH check runs registered under the \"Canvas (Next.js)\" name (one SUCCESS via no-op, one SKIPPED via real job), and the SKIPPED one kept \`mergeStateStatus: BLOCKED\` despite the SUCCESS sibling. Branch protection's \"set of matching contexts\" semantic is stricter than the \`branch_protection_check_name_parity\` durable feedback memory documented — at least one passing isn't enough; SKIPPED counts as not-passed regardless of how many SUCCESS siblings it has.

## Real fix

ONE job that always runs (no job-level \`if:\`), with all real work gated on the path filter via per-step \`if:\`. Produces exactly **one** \`Canvas (Next.js)\` check run per commit, always SUCCEEDS, regardless of which paths changed.

Cost: ~10s of always-on CI runtime per PR (the ubuntu-latest spin-up + the no-op echo step). Negligible vs. the manual-merge cost when BLOCKED catches us.

## Latent bug worth flagging

Three other required checks use the same \`if: needs.changes.outputs.X == 'true'\` pattern at the job level:

- \`Platform (Go)\` — \`platform\` filter
- \`Python Lint & Test\` — \`python\` filter
- \`Shellcheck (E2E scripts)\` — \`scripts\` filter

PR-A and PR #2321 merged because their diffs happened to trigger every filter; PR-B (#2314) only missed canvas. **Future PRs that miss any of these filters will hit the same BLOCKED state.** Filing a follow-up to apply the same single-job-conditional-steps pattern across all four required checks.

## Test plan

- [x] YAML validates (\`python3 -c \"import yaml; yaml.safe_load(open('.github/workflows/ci.yml'))\"\`)
- [ ] After merge: PR #2314's next CI run produces a single SUCCESS \`Canvas (Next.js)\` check → \`mergeStateStatus\` clears → auto-merge fires
- [ ] Verify on a PR that DOES touch canvas/** that the real build still runs

## Refs

- PR #2321 (the fix-attempt that needs replacing)
- Memory note \`feedback_branch_protection_check_name_parity\` needs correcting (the documented "two jobs same name" pattern doesn't actually work — branch protection sees both check runs as a SET; SKIPPED in the set blocks).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -178,42 +178,44 @@ jobs:
             exit 1
           fi
 
-  # Path-filter no-op shadow for Canvas (Next.js).
+  # Canvas (Next.js) is a required check on staging branch protection.
+  # The job ALWAYS runs (no `if:` gate at job level) so branch protection
+  # always sees a SUCCESS conclusion. Inside, every meaningful step is
+  # gated on `needs.changes.outputs.canvas == 'true'` — when the PR
+  # doesn't touch canvas/**, the job spins up, skips all real work, and
+  # exits clean in ~10s.
   #
-  # Branch protection on staging requires a "Canvas (Next.js)" check.
-  # When a PR doesn't touch canvas/** paths, the real canvas-build job
-  # below is skipped via `if:`, and GitHub reports its conclusion as
-  # SKIPPED — which branch protection treats as not-passed → merge
-  # BLOCKED on every workspace-server-only or migration-only PR.
-  #
-  # Pattern (per durable feedback memory: branch_protection_check_name_parity):
-  # split into a real job + a no-op shadow that share the same `name:`.
-  # Exactly one runs per PR; both report the same check context, and at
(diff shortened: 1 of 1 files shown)


hongmingwang-moleculeai (approved):
Single-job-with-conditional-steps is the correct shape — branch protection sees exactly one Canvas (Next.js) check run per commit and it always reports SUCCESS. Confirmed via the live failure mode on PR #2314 that the previous two-jobs-same-name pattern doesn't satisfy 'required check' evaluation when one of the check runs is SKIPPED.

PR #2327, opened 2026-04-30
fix(e2e-canvas): kill teardown race that poisons concurrent runs (unblocks staging→main #2264)

HongmingWang-Rabbit:
## Summary

The canvas-E2E teardown safety net was pattern-sweeping every \`e2e-canvas-<today>-*\` org as compensation for setup-crash-before-state-file-write. That sweep raced concurrent canvas-E2E runs and deleted their live tenants mid-test — surfacing as \`getaddrinfo ENOTFOUND\` once CP cleaned up the just-deleted DNS record.

Fix: write the state file as setup's **first action**, before any CP call. Teardown always has the slug, so the workflow safety-net can read it and delete only that slug.

## Root cause

| When | Old behavior | New behavior |
|---|---|---|
| Crash before slug generation | No state file → safety-net pattern-sweeps everyone's tenants | No state file → safety-net no-ops (no orphan to clean) |
| Crash during steps 1-6 (org created, but workspace-online never resolved) | No state file → safety-net pattern-sweeps everyone's tenants | State file has slug → safety-net deletes only that slug |
| Setup completes successfully | State file has full state → globalTeardown deletes own slug | Same — state file just gets overwritten with full state at step 7 |

## Race observed

PR #2264 staging→main, 2026-04-30. Three real-test canvas-E2E runs killed each other mid-test. Timeline (each row = a peer's teardown nuking the row below it):

| Time | Run | Result | What killed it |
|---|---|---|---|
| 00:30:06 | PR (real, 12min) | ✓ — finishes ~00:42, runs teardown | — |
| 00:36:08 | push (real, 7.8min) | ✗ fail at **00:43:57** | 00:30 PR's teardown |
| 00:36:10 | PR (real, 20min) | ✓ — finishes ~00:56, runs teardown | — |
| 00:50:01 | push (real, 8.3min) | ✗ fail at **00:58:16** | 00:36 PR's teardown |
| 00:50:04 | PR (real, 21min) | ✓ — finishes ~01:11, runs teardown | — |
| 00:50:18 | PR rerun (real, 24min) | ✗ fail at **01:14:25** | 00:50:04 PR's teardown |

Every failure is within ~2 min of a peer's teardown firing. Not a CloudFlare DNS bug, not a sweep-cf-orphans bug — both verified clean.

## Files changed

- \`canvas/e2e/staging-setup.ts\`: write state file right after \`makeSlug()\`; promote to full state at step 7 (overwrite)
- \`canvas/e2e/staging-teardown.ts\`: comment update — missing state file genuinely means nothing to clean now
- \`.github/workflows/e2e-staging-canvas.yml\`: replace pattern-sweep with state-file lookup; delete only the recorded slug

## Test plan

- [ ] CI passes on this PR (path filter SHOULD trigger real run since canvas/e2e/ changed)
- [ ] After merge, verify PR #2264 staging→main can advance — Canvas tabs E2E should no longer self-poison
- [ ] Teardown still cleans up successfully on a normal happy-path run
- [ ] Teardown still cleans up the orphan when staging-setup crashes mid-flight (e.g., on a workspace-online timeout)

### DIFF
--- a/.github/workflows/e2e-staging-canvas.yml
+++ b/.github/workflows/e2e-staging-canvas.yml
@@ -153,43 +153,39 @@ jobs:
           path: canvas/test-results/
           retention-days: 14
 
-      # Safety-net teardown mirrors the bash-harness workflow — if
-      # globalTeardown didn't run (worker crash, runner cancel), this
-      # step sweeps any e2e-canvas-* org tagged with today's date.
+      # Safety-net teardown — fires only when Playwright's globalTeardown
+      # didn't (worker crash, runner cancel). Reads the slug from
+      # canvas/.playwright-staging-state.json (written by staging-setup
+      # as its first action, before any CP call) and deletes only that
+      # slug.
+      #
+      # Earlier versions of this step pattern-swept `e2e-canvas-<today>-*`
+      # orgs to compensate for setup-crash-before-state-file-write. That
+      # over-aggressive cleanup raced concurrent canvas-E2E runs and
+      # poisoned each other's tenants — observed 2026-04-30 when three
+      # real-test runs killed each other mid-test, surfacing as
+      # `getaddrinfo ENOTFOUND` once CP had cleaned up the just-deleted
+      # DNS record. Pattern-sweep removed; setup now writes the state
+      # file before any CP work, so the slug is always recoverable.
       - name: Teardown safety net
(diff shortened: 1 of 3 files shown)


hongmingwang-moleculeai (approved):
Approved — root-cause fix for the teardown race. State file written before CP calls means concurrent runs can no longer poison each other.

Issue #2312, opened 2026-04-29
RFC: replace Docker-exec handlers with HTTP-forward to workspace's own URL (chat upload + 5 follow-ons)


HongmingWang-Rabbit:
## Problem

Six `workspace-server` handlers reach into a workspace by acquiring a local Docker client and either `exec`'ing or `tar`-copying through it:

- `POST /workspaces/:id/chat/uploads`  (chat_files.go) — file upload
- `GET  /workspaces/:id/chat/download` (chat_files.go) — file download
- `GET/PUT/DELETE /files/*`            (templates.go) — template/config files
- `POST /templates/import`             (templates.go) — bulk template import
- `WS   /terminal`                     (terminal handler)
- plugin install / uninstall flows

All of them call `TemplatesHandler.findContainer(ctx, workspaceID)` which short-circuits to `\"\"` when `h.docker == nil`.

In SaaS (`MOLECULE_ORG_ID` set), `cmd/server/main.go` selects the **CP provisioner** instead of the local Docker provisioner — `dockerCli` is nil. Result: every one of the six surfaces returns 503 \"workspace container not running\" on Railway-hosted production, regardless of runtime. Self-hosted local Docker continues to work because `h.docker` is non-nil and the workspace runs on the same host.

## Proposed architecture

Stop reaching into containers entirely. Use HTTP everywhere:

```
canvas / molecli / SDK
     │
     ▼
platform workspace-server  ──HTTPS──▶  workspace's own server (runs inside container)
     │                          │            │
     │                          │            ▼
     │                          │      writes to local fs
     │                          │      (e.g. /workspace/.molecule/chat-uploads)
     │                          │
     └─ workspace bearer token ─┘
```

`workspaces.url` already differs correctly per environment:
- Local Docker: `http://ws-<id>:8000` (Docker-network hostname)
- SaaS EC2: `https://<id>.<tenant>.moleculesai.app`
- External: caller-provided public URL

So one platform-side call site — \"POST/GET to `<workspaces.url><internal_path>` with `Authorization: Bearer <wsAuthToken>`\" — works in every environment without branching.

## Adapter shape (universal default + per-runtime override)

Each operation lands two pieces:

[...]

PR #2334, opened 2026-04-30
docs(chat_files): update header — Download is HTTP-forward, not docker-cp

HongmingWang-Rabbit:
Pure docs change. The \`chat_files.go\` header still claimed Download uses the v1 docker-cp path and that the migration was a follow-up PR — both stale. RFC #2312 PR-D landed Download HTTP-forward at line 336 (\`http.NewRequestWithContext\` to \`/internal/file/read\`).

Updated header reflects actual code: both upload + download are HTTP-forward, share the same \`platform_inbound_secret\` auth, work uniformly on local Docker and SaaS EC2.

No behavior change, no build/test impact.

### DIFF
--- a/workspace-server/internal/handlers/chat_files.go
+++ b/workspace-server/internal/handlers/chat_files.go
@@ -1,18 +1,26 @@
 package handlers
 
-// chat_files.go — file upload (HTTP-forward) + download (Docker-exec)
-// for workspace chat.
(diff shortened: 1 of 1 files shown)


hongmingwang-moleculeai (approved):
Approved — header now matches reality (both upload+download HTTP-forward post RFC #2312).

Issue #2339, opened 2026-04-30
Design: unified poll/push delivery for external runtimes (Telegram-getUpdates / Discord-Gateway shape)


HongmingWang-Rabbit:
## TL;DR

External runtimes (operator's laptop, Python SDK consumers, custom agents) need to talk to the platform from anywhere — including networks where they have no public URL. Today the platform always tries to push to a registered URL, fails the SSRF check silently if the URL is unroutable, and **doesn't even log the inbound activity** — so the polling channel plugin sees nothing and the canvas user sees a red "agent unreachable" banner.

This issue proposes a Telegram-getUpdates / Discord-Gateway-shaped pull mode as a first-class delivery mode, with cursor-based correctness and an SSE upgrade path. Single architecture, works local + SaaS, no tunnels.

## Concrete bug that motivated this

Steps to reproduce on `hongmingwang.moleculesai.app`:
1. Create workspace with `runtime=external`, register agent, heartbeat keeps it `online`.
2. Open canvas chat for the workspace, type "hi".
3. Canvas shows red "Failed to send message — agent may be unreachable".
4. `GET /workspaces/:id/activity?since_secs=3600` returns `[]` — no record of the inbound at all.

Root cause: `workspace-server/internal/handlers/a2a_proxy.go:454-461` rejects on SSRF and short-circuits *before* `logA2AFailure` runs. The polling channel can't surface what was never logged.

## Prior art

| Platform | Pull mode | Push mode | Cursor | Mutually exclusive? |
|---|---|---|---|---|
| **Slack** | Socket Mode (WS) | Events API (webhook) | per-envelope ack | yes (operator picks) |
| **Discord** | Gateway (WS) | HTTP Interactions | sequence + RESUME | gateway is the only full-feature path |
| **Telegram** | getUpdates (long-poll) | setWebhook | update_id offset | **enforced exclusive** |
| **Stripe** | `stripe listen` (CLI WS relay) | direct webhooks | event ID | dev-only relay |
| **AWS IoT** | MQTT-over-WSS | n/a | message ID + QoS | pull-only by design |

Five patterns:
1. Pull mode is mandatory at scale — laptops, CI runners, NAT, corp firewalls don't have public URLs.
2. Mutual exclusivity is operationally cleaner. Telegram refuses to deliver via both at once; Slack Socket Mode silences Events API. Both transports half-working at once is a known anti-pattern.
[...]

PR #2348, opened 2026-04-30
feat(workspaces): delivery_mode column + poll-mode register flow (#2339 PR 1)

HongmingWang-Rabbit:
First PR in the [#2339](https://github.com/Molecule-AI/molecule-core/issues/2339) series — unified poll/push delivery for external runtimes (Telegram-getUpdates / Discord-Gateway shape).

This PR adds the schema column and reshapes the register handler. **No behavior change for existing push-mode workspaces** — every existing row keeps the schema default (`push`) and the same dispatch path. The poll-mode short-circuit lands in PR 2; the cursor lands in PR 3; the plugin upgrade lands separately in `Molecule-AI/molecule-mcp-claude-channel`.

## Summary

- **Schema:** new `workspaces.delivery_mode` column (NOT NULL, default `push`, CHECK constraint on `push|poll`). Migration 045.
- **Models:** `Workspace.DeliveryMode`, `RegisterPayload.DeliveryMode`, `CreateWorkspacePayload.DeliveryMode`. `RegisterPayload.URL` drops the `binding:\"required\"` tag — handler enforces conditionally.
- **Register handler:**
  - Validates explicit `delivery_mode` if set (rejects unknown values with 400).
  - Resolves effective mode AFTER C18 token check: payload value, else stored row value, else `push`.
  - URL required + SSRF-validated **only** when effective mode is `push`.
  - Poll-mode skips URL caching (no cache pollution with empty strings).
  - Persists `delivery_mode` in upsert; returns it in the response body so the client can confirm what the server resolved.
- **Create handler:** persists `delivery_mode` (defaults to `push`), validates before any side effects.

## Why this design

- **Sticky mode (Telegram-shape):** absent payload value defers to the existing row's mode, so an established poll-mode workspace doesn't silently downgrade to push on a re-register that omits the field. New `TestRegister_PollMode_PreservesExistingValue` enforces this.
- **Mode resolution AFTER token check:** prevents a side-channel where an attacker could probe \"does this workspace exist?\" via the delivery_mode lookup before the C18 auth gate fires.
- **Empty URL vs sentinel:** chose nullable URL + delivery_mode flag over a sentinel like `urn:molecule:poll-only`. Cleaner to query, no parsing, and the column tells the truth about what the workspace expects.

## Tests

New (4):
- `TestRegister_PollMode_AcceptsEmptyURL` — register with `delivery_mode=poll`, no URL → 200, response includes `delivery_mode: poll`, no URL caching, token issuance still works.
- `TestRegister_PushMode_RejectsEmptyURL` — register with no URL and no explicit mode → 400 \"url is required\".
- `TestRegister_InvalidDeliveryMode` — register with `delivery_mode=webhook` → 400.
- `TestRegister_PollMode_PreservesExistingValue` — re-register without `delivery_mode`, row already has `poll` → mode stays poll (no silent downgrade).

Updated (existing tests touching `INSERT INTO workspaces` mocks):
- `registry_test.go` — 6 register tests get the new `SELECT delivery_mode` mock + the new INSERT arg.
- `handlers_test.go`, `handlers_additional_test.go`, `workspace_test.go`, `workspace_budget_test.go` — every CreateWorkspace mock + every Register mock gets the `\"push\"` arg appended.

`go test ./...` clean across all 17 packages.

## Test plan

- [x] All new tests pass
- [x] All updated existing tests pass (full handler suite green)
- [x] `go build ./...` clean
- [x] Migration .up + .down syntactically valid; .down is destructive (drops column) but consumers fall back to push (pre-2339 behavior)
- [ ] Reviewer: confirm the resolve-mode-after-token-check ordering is the right call (alternatives: combine into a single query, or always do the lookup unconditionally)
- [ ] Reviewer: confirm the `binding:\"required\"` removal on `RegisterPayload.URL` is acceptable — handler enforces conditionally, no test covers a missing `url` field arriving with `delivery_mode=push` other than the new `TestRegister_PushMode_RejectsEmptyURL`

## Follow-ups in this series

1. **PR 2 — ProxyA2A poll-mode short-circuit:** when `delivery_mode=poll`, log `a2a_receive` activity and return 200 to caller without invoking SSRF or dispatch. Closes the silent-404 failure mode that motivated #2339.
2. **PR 3 — `since_id` cursor on `/activity`:** ascending cursor pagination so polling clients survive arbitrary downtime without losing events.
3. **Plugin v0.2 in `Molecule-AI/molecule-mcp-claude-channel`:** cursor persistence + `delivery_mode=poll` register helper.
4. **PR 4 — SSE upgrade:** `/workspaces/:id/events` streaming endpoint with `Last-Event-ID` resume.
5. **E2E test:** full external+poll round-trip in `tests/e2e/`.

### DIFF
--- a/workspace-server/internal/handlers/handlers_additional_test.go
+++ b/workspace-server/internal/handlers/handlers_additional_test.go
@@ -31,8 +31,9 @@ func TestWorkspaceCreate_WithParentID(t *testing.T) {
 	parentID := "parent-ws-123"
 	mock.ExpectBegin()
 	// Default tier is 3 (Privileged) — see workspace.go create-handler comment.
+	// delivery_mode defaults to "push" when payload omits it (#2339).
 	mock.ExpectExec("INSERT INTO workspaces").
-		WithArgs(sqlmock.AnyArg(), "Child Agent", nil, 3, "langgraph", sqlmock.AnyArg(), &parentID, nil, "none", (*int64)(nil), models.DefaultMaxConcurrentTasks).
+		WithArgs(sqlmock.AnyArg(), "Child Agent", nil, 3, "langgraph", sqlmock.AnyArg(), &parentID, nil, "none", (*int64)(nil), models.DefaultMaxConcurrentTasks, "push").
 		WillReturnResult(sqlmock.NewResult(0, 1))
 	mock.ExpectCommit()
 	mock.ExpectExec("INSERT INTO canvas_layouts").
@@ -66,8 +67,9 @@ func TestWorkspaceCreate_ExplicitClaudeCodeRuntime(t *testing.T) {
 	handler := NewWorkspaceHandler(broadcaster, nil, "http://localhost:8080", t.TempDir())
 
 	mock.ExpectBegin()
+	// delivery_mode defaults to "push" when payload omits it (#2339).
 	mock.ExpectExec("INSERT INTO workspaces").
-		WithArgs(sqlmock.AnyArg(), "CC Agent", nil, 2, "claude-code", sqlmock.AnyArg(), (*string)(nil), nil, "none", (*int64)(nil), models.DefaultMaxConcurrentTasks).
+		WithArgs(sqlmock.AnyArg(), "CC Agent", nil, 2, "claude-code", sqlmock.AnyArg(), (*string)(nil), nil, "none", (*int64)(nil), models.DefaultMaxConcurrentTasks, "push").
 		WillReturnResult(sqlmock.NewResult(0, 1))
 	mock.ExpectCommit()
 	mock.ExpectExec("INSERT INTO canvas_layouts").
@@ -288,7 +290,7 @@ func TestWorkspaceCreate_MaxConcurrentTasksOverride(t *testing.T) {
 
 	mock.ExpectBegin()
 	mock.ExpectExec("INSERT INTO workspaces").
-		WithArgs(sqlmock.AnyArg(), "Leader Agent", nil, 3, "claude-code", sqlmock.AnyArg(), (*string)(nil), nil, "none", (*int64)(nil), 3).
+		WithArgs(sqlmock.AnyArg(), "Leader Agent", nil, 3, "claude-code", sqlmock.AnyArg(), (*string)(nil), nil, "none", (*int64)(nil), 3, "push").
 		WillReturnResult(sqlmock.NewResult(0, 1))
 	mock.ExpectCommit()
 	mock.ExpectExec("INSERT INTO canvas_layouts").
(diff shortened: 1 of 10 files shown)


HongmingWang-Rabbit (commented):
Five-axis pass on a fast read of the diff + description. Substantive logic change so leaving as comment only — final approve from a human after the open questions land.

**Correctness ✓** — Sticky-mode preservation on re-register is the right invariant for a poll-mode workspace (Telegram clients regularly re-issue `/setWebhook`-equivalents without re-asserting their delivery shape, and a silent downgrade to push would silently send dispatches to whatever the URL field was last set to). The new `TestRegister_PollMode_PreservesExistingValue` pins this and is the one I'd most want to keep stable as future PRs in this series land. Migration 045 with `NOT NULL DEFAULT 'push'` + CHECK is the safe shape — existing rows get the right default, and the `.down.sql` drop-column path is the right destructive choice (rollback semantics: workspaces fall back to push, which is pre-#2339 behaviour).

**Readability ✓** — File naming matches repo convention; description is unusually thorough about *why* (sticky mode, ordering rationale, alternative-considered notes).

**Architecture ✓** — Single column add, models updated symmetrically (Workspace + RegisterPayload + CreateWorkspacePayload), test mocks updated across all 4 suites that touch register/create. The 243-line growth in registry_test.go is mostly mock-ledger churn (6 register tests × added `SELECT delivery_mode` mock + INSERT arg), not new asserts — appropriate.

**Security ✓** — On the explicit reviewer question about *resolve-mode-after-token-check ordering*: this is the right call. Combining into a single query would couple auth and delivery-mode lookup, which makes a future "add another resolved field" PR have to think about both. Keeping them separate is cleaner, and the side-channel argument (no enumeration via differential errors before C18) is sound. The token check is cheap; the savings of skipping the second lookup on auth-fail are also cheap. Don't combine.

**Performance ✓** — One extra `SELECT delivery_mode` per register. Register isn't a hot path (it's per-workspace-establish), so the cost is negligible.

**On the `binding:\"required\"` removal**: acceptable. The conditional handler-side enforcement is now the source of truth, and `TestRegister_PushMode_RejectsEmptyURL` exercises the "empty URL + default push mode" path. One extra test you might consider adding: explicit `delivery_mode=push` + empty URL → 400 (not just default-mode + empty URL). Same code path, but pins that explicit-push doesn't accidentally bypass the requirement. Optional.

**Branch coverage on the new register-handler logic**:
- explicit `delivery_mode=poll` → `TestRegister_PollMode_AcceptsEmptyURL` ✓
- explicit `delivery_mode=invalid` → `TestRegister_InvalidDeliveryMode` ✓
- absent `delivery_mode` + push default + empty URL → `TestRegister_PushMode_RejectsEmptyURL` ✓
- absent `delivery_mode` + existing row=poll → `TestRegister_PollMode_PreservesExistingValue` ✓
- (missing) absent `delivery_mode` + existing row=push → behaves identically to before; existing tests cover ✓

5/5 — every prod branch in the new ordering has a test.

**Nit (not blocking)**: in registry.go the resolution comment could note *why* the order matters (no enumeration via differential errors). The description has it; the code doesn't. Future code-reader benefits.

PR ready for human merge once the in-progress check completes and the followup PR-2/PR-3 plan is confirmed.

Issue #2345, opened 2026-04-30
A2A v0.3 jsonrpc adapter silently drops messages with old (.content) format — should accept both or reject loud


HongmingWang-Rabbit:
## Symptom

Design Director appeared to accept A2A briefs (HTTP 200/202) but never delegated to UX Researcher. Pipeline silently idle. Investigation found the adapter is rejecting messages at parse time:

\`\`\`
params.message.parts — Field required [type=missing]
\`\`\`

Sender uses the **v0.2 format** (\`params.message.content\`). Adapter expects the **v0.3 format** (\`params.message.parts\`). The mismatch isn't surfaced to the sender — the HTTP layer accepts the request, the parse layer fails, and the message disappears into a log line nobody reads.

## What's actually broken

1. **A2A v0.3 jsonrpc adapter rejects v0.2 format silently.** \`a2a/compat/v0_3/jsonrpc_adapter.py\` (path per the dogfooding report). Should accept BOTH \`content\` and \`parts\` for compatibility — or REJECT loud (return a structured error to sender) instead of dropping silently.

2. **Design Director → UX Researcher delegation likely also broken.** Same JSON-RPC adapter, same code path. Need to verify by re-running an end-to-end org-template kickoff after the adapter fix.

3. **Bind-mount template sync may have failed.** 7 workspaces still on the old template. May or may not be related — orthogonal symptom worth checking.

## Why this is a hard-gate-class failure

This bug:

- Passed unit tests (the adapter parses valid v0.3 fine; the test never sent v0.2)
- Passed PR-time E2E (we don't run a continuous mixed-version-sender test)
- Was invisible to operators (HTTP 200 + silent parse failure looks like \"accepted, processing\")
- Manifested as \"the multi-agent pipeline is idle\" — generic symptom, hard to root-cause

It would have been caught by:

- **Hard gate #2 (continuous synthetic E2E, #2342)** — a periodic full org-template kickoff would have failed within 15 min of the deploy that broke compat
- **Schema-strictness contract test** — an adapter test that asserts \"unknown fields in inbound messages must produce a structured 4xx error,\" not silent drop

### Tier 1 — accept both formats (~30 min)

In \`a2a/compat/v0_3/jsonrpc_adapter.py\`:

\`\`\`python
def parse_message(payload):
    msg = payload[\"params\"][\"message\"]
    if \"parts\" in msg:
        parts = msg[\"parts\"]                 # v0.3 native
    elif \"content\" in msg:
[...]

PR #2350, opened 2026-04-30
ci: continuous synthetic E2E against staging (#2342)

### LINKED ISSUE #2342: Hard gate: continuous synthetic E2E against staging (15-min cadence)
Hard-gates Tier 2, item 2 of 4. Highest-leverage gate on the deck.

## Problem

\`e2e-staging-saas\` runs on PR push. It does NOT run continuously. Result: regressions that escape PR-time CI (e.g., deployment-pipeline gaps, third-party API drift, infra config changes outside this repo) are caught only when a human dogfoods the system or another PR's E2E happens to fire.

Today's case: RFC #2312 chat upload landed on staging-branch but \`:staging-latest\` image wasn't rebuilt because publish-workspace-server-image was main-only. Staging tenants kept failing chat upload for hours. Caught by dogfooding, not by automation. A continuous synthetic E2E would have alerted within ~15 min of the regression manifesting.

## Proposed gate

Cron-scheduled workflow that runs \`tests/e2e/test_staging_full_saas.sh\`-style provisioning + smoke test against staging every 15 min:

1. Provision a fresh tenant (slug \`canary-synth-<timestamp>\`)
2. Create a workspace inside it
3. Run the chat-upload happy path against the tenant URL
4. Run an A2A delegate happy path
5. Tear down

On failure: alert via PushNotification, GitHub issue auto-created, optionally PagerDuty for prod equivalent.

Per-run cost: ~$0.05 GHA + ~$0.001 staging tenant. ~$5/day. Negligible.

## What it catches

- Today's RFC #2312 \"deployed on staging branch but not on staging tenants\" gap
- Stale dependencies / vendor outages affecting our flow
- Staging-CP env var drift (e.g., admin token rotated without updating)
- DNS / Cloudflare / Railway side-changes that break tenant provisioning silently
- \"applyRuntimeModelEnv missing\" 2026-04-24 incident class

## What it does NOT catch

- Bugs in code that's not on the synthetic path (e.g., specific runtimes other than the one tested)
- Pre-merge regressions (those are PR-time E2E's job)

## Acceptance criteria

- New \`continuous-synth-e2e.yml\` workflow with \`schedule: '*/15 * * * *'\` cron
- Tenant provisioning + chat upload + A2A delegate runs to completion in <8 min
- Failure produces a GitHub issue with diagnostic output (or comments on a single rolling \"continuous-E2E-status\" issue)
- Tenant cleanup is bulletproof — leveraging the per-run state file pattern from PR #2327

## Estimated effort

~1 day. Most of the harness exists in \`canvas/e2e/staging-setup.ts\` and \`tests/e2e/test_staging_full_saas.sh\`.

## Refs

Hard-gates discussion 2026-04-30. Tier 2, item 2. Closes the dogfooding observation cycle (problem visible-only-to-humans → problem visible-to-CI within 15 min).

HongmingWang-Rabbit:
Hard gate Tier 2 item 2 of 4. Cron-driven full-lifecycle E2E that catches regressions visible only at runtime.

## Empirical motivation (from today)

Two real failures from this session that PR-time CI couldn't catch but a continuous synth E2E would have surfaced within ~20 min:

1. **#2345 — A2A v0.2 silent drop** — passed unit tests, broke at JSON-RPC parse layer. Visible only with a sender exercising the full path. Fixed by #2349; would have caught the regression at deploy time.
2. **RFC #2312 chat upload** — landed staging-branch, never reached staging tenants because publish-workspace-server-image was main-only. Caught hours later by manual dogfooding. Same pattern.

Both classes are invisible to PR-time CI. The continuous gate fires every 20 min against a real staging tenant.

## Design

| Decision | Why |
|---|---|
| Cron `0,20,40 * * * *` (3x/hour) | Offsets sweep-cf-orphans (:15) and sweep-cf-tunnels (:45) so CF/AWS APIs don't get bursted at the same minute |
| Reuses `tests/e2e/test_staging_full_saas.sh` | No new harness to maintain; same script that PR-time E2E uses |
| `E2E_RUNTIME=langgraph` (default) | Fastest cold start; the regression classes this catches don't need hermes-specific paths. Operator can dispatch with hermes for SDK-native coverage |
| `E2E_PROVISION_TIMEOUT_SECS=600` (10 min) | Stuck runs fail fast, don't hold up next firing |
| `concurrency: continuous-synth-e2e, cancel-in-progress: false` | Two firings can never overlap |
| Hard-fail-on-missing-secret for cron | Mirror sweep-cf-* hardening — silent-skip would mask real outages |
| Soft-skip on dispatch | Operators can ad-hoc dispatch to verify a fix without secret setup |

## Cost

~3 runs/hour × 5-10 min × $0.008/min GHA = **$0.50-$1/day**. Plus pennies of staging tenant lifecycle (Railway + AWS). Negligible.

## What this catches that PR-time CI doesn't

- Schema drift between sender + receiver (today's #2345 class)
- Deployment-pipeline gaps (today's RFC #2312 class — code merged but image not built)
- Vendor outages (Cloudflare DNS, Railway, AWS, GHCR)
- Staging-CP env-var rotations (admin token rolled, secret-scan key rotated, etc.)
- AMI / cloudflared / DNS / TLS-cert drift over time

## What this doesn't replace

- PR-time CI on \`e2e-staging-saas.yml\` — that's the on-PR gate; this is the **between-PR** gate
- Production monitoring for prod CP — staging-only here
- Per-runtime coverage — defaults to langgraph; operator dispatches hermes/claude-code as needed

## Failure surface

Failed runs surface in the workflow's history (red). Job summary embeds a triage guide. No auto-issue creation yet — keeping surface minimal until that's actually wanted; can layer on top later.

## Refs

- Closes #2342 hard-gates Tier 2 item 2
- Empirical motivation: #2345 (just-fixed), RFC #2312 stack, #2335
- Hard-gates discussion 2026-04-30

### DIFF
--- a/.github/workflows/continuous-synth-e2e.yml
+++ b/.github/workflows/continuous-synth-e2e.yml
@@ -1,0 +1,160 @@
+name: Continuous synthetic E2E (staging)
+
+# Hard gate (#2342): cron-driven full-lifecycle E2E that catches
+# regressions visible only at runtime — schema drift, deployment-pipeline
+# gaps, vendor outages, env-var rotations, DNS / CF / Railway side-effects.
+#
+# Why this gate exists:
+#   PR-time CI catches code-level regressions but not deployment-time or
+#   integration-time ones. Today's empirical data:
+#     • #2345 (A2A v0.2 silent drop) — passed all unit tests, broke at
+#       JSON-RPC parse layer between sender and receiver. Visible only
+#       to a sender exercising the full path.
+#     • RFC #2312 chat upload — landed on staging-branch but never
+#       reached staging tenants because publish-workspace-server-image
+#       was main-only. Caught by manual dogfooding hours after deploy.
+#   Both would have surfaced within 15-20 min of regression if a
+#   continuous synth-E2E was running.
+#
+# Cadence: every 20 min (3x/hour). The script is conservatively
+# bounded at 10 min wall-clock; even on degraded staging it should
+# finish before the next firing. cron-overlap is guarded by the
+# concurrency group below.
+#
+# Cost: ~3 runs/hour × 5-10 min × $0.008/min GHA = ~$0.50-$1/day.
+# Plus a fresh tenant provisioned + torn down each run (Railway +
+# AWS pennies). Negligible.
+#
+# Failure handling: when the run fails, the workflow exits non-zero
+# and GitHub's standard email/notification path fires. Operators
+# can subscribe to this workflow's failure channel for paging-grade
+# alerting.
+
+on:
+  schedule:
+    # Every 20 minutes, on the :00 :20 :40. Offsets the existing :15
+    # sweep-cf-orphans and :45 sweep-cf-tunnels so the three
+    # operations don't all hit Cloudflare/AWS at the same minute.
+    - cron: '0,20,40 * * * *'
+  workflow_dispatch:
+    inputs:
+      runtime:
+        description: "Runtime to provision (langgraph = fastest, default; hermes = slower but covers SDK-native path; claude-code = needs OAUTH token in tenant env)"
+        required: false
+        default: "langgraph"
+        type: string
+      keep_org:
+        description: "Skip teardown for post-mortem debugging (only manual dispatch — never set this for cron runs)"
+        required: false
+        default: false
+        type: boolean
+
+permissions:
+  contents: read
+  # No issue-write here — failures surface as red runs in the workflow
+  # history. If you want auto-issue-on-fail, add a follow-up step that
+  # uses gh issue create gated on `if: failure()`. Keeping the surface
+  # minimal until that's actually wanted.
+
+# Serialize so two firings can never overlap. Cron firing every 20 min
+# but scripts conservatively bounded at 10 min — overlap shouldn't
+# happen in steady state, but if a run hangs we don't want N more
+# stacking up.
+concurrency:
+  group: continuous-synth-e2e
+  cancel-in-progress: false
+
+jobs:
+  synth:
+    name: Synthetic E2E against staging
+    runs-on: ubuntu-latest
+    timeout-minutes: 12
+    env:
+      # langgraph default keeps cold-start under 5 min on staging EC2.
+      # hermes is slower (~7-10 min) and isn't needed for the
+      # regression class this gate exists to catch (deployment-pipeline
+      # + schema-drift + integration). Operators can pick hermes via
+      # workflow_dispatch when they need to exercise the SDK-native
+      # session path.
+      E2E_RUNTIME: ${{ github.event.inputs.runtime || 'langgraph' }}
+      # Bound to 10 min so a stuck provision fails the run instead of
+      # holding up the next cron firing. 15-min default in the script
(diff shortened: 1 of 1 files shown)


HongmingWang-Rabbit (commented):
Five-axis pass on the workflow file. Per loop rules, .github/workflows/* changes get COMMENT-only.

**Correctness ✓** — Cron offset (`:00/:20/:40`) is intentional and *correct* against the existing operations cadence: sweep-cf-orphans fires at `:15` and sweep-cf-tunnels at `:45`, so the 3 minute-buckets don't burst the same APIs simultaneously. Schedule-vs-dispatch secret hardening (hard-fail on cron missing secret, soft-skip on workflow_dispatch) matches the pattern used by the sweep-cf-* and redeploy-tenants-on-* workflows — and it's also encoded in the project's feedback memory as a learned-the-hard-way lesson, so this is the right shape.

**Readability ✓** — Comments explain the why (cron offset, soft-skip-on-dispatch reason, slug namespacing). Failure summary writes a triage runbook into `$GITHUB_STEP_SUMMARY` rather than raw log scroll. Operator-friendly `workflow_dispatch.inputs.runtime` description spells out the cold-start tradeoff.

**Architecture ✓** — Reuses `tests/e2e/test_staging_full_saas.sh`; doesn't fork the harness. `concurrency: continuous-synth-e2e` + `cancel-in-progress: false` is the right call for a slow-but-bounded job (a 10-min run hanging shouldn't cancel; the next firing should just queue). Separates between-PR gate from on-PR gate cleanly and documents what each catches.

**Security ✓** —
- `actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4` is SHA-pinned (per the project's third-party-action pinning policy)
- `permissions: contents: read` only — no write tokens issued
- `secrets.CP_STAGING_ADMIN_API_TOKEN` referenced via secret context, never logged
- Tool presence check (`jq`, `curl`, `python3`) fails fast on runner-image regression
- The script's EXIT-trap teardown is the right call for ensuring no orphan tenant on assertion failure

**Performance ✓** — 3x/hour × 10-min cap = ~/bin/zsh.50-/day GHA budget, well within reasonable. timeout-minutes:12 (just over the 10-min E2E_PROVISION_TIMEOUT_SECS:600) gives the script's EXIT trap time to teardown cleanly before the runner kills the job.

**One small thing worth considering (not blocking):** the failure summary mentions "Likely classes" as bullets but doesn't pin a specific failure mode by parsing the script output. Future iteration: the script could emit a single-line classifier (e.g. `E2E_FAIL_REASON=schema_drift|deploy_gap|vendor_outage|env_drift`) that this step echoes verbatim. Not needed for v1 but a small lift if the post-mortem-by-summary pattern proves useful.

Branch coverage: workflow file is the artifact under test; cron path + dispatch path + missing-secret path + tool-missing path are all exercised by inspection (no Go unit tests on YAML, but the path semantics match the existing sweep-cf-*/redeploy-tenants-on-* siblings).

Clean implementation. Ready for human merge once you've verified `CP_STAGING_ADMIN_API_TOKEN` is set in repo secrets and the next cron firing observed once before relying on it.

PR #2362, opened 2026-04-30
fix(a2a): detect dead EC2 agents on upstream 5xx + reactive auto-restart for SaaS

HongmingWang-Rabbit:
## Summary

Class-of-bugs fix surfaced by `hongmingwang.moleculesai.app`'s canvas chat to a dead workspace returning a generic Cloudflare 502 page on 2026-04-30. Three independent gaps in the reactive-health path that together leak dead-agent failures to canvas with no auto-recovery on every SaaS tenant.

### Bug 1 — `maybeMarkContainerDead` is a no-op for SaaS tenants

`maybeMarkContainerDead` only consulted `h.provisioner` (local Docker provisioner). SaaS tenants set `h.cpProv` (CP-backed EC2 provisioner) and leave `h.provisioner` nil — so the function early-returned `false` on every call and dead EC2 agents never triggered the offline-flip / broadcast / restart cascade. **All SaaS tenants pre-this-fix had zero reactive auto-recovery.**

**Fix:** extend `CPProvisionerAPI` interface with `IsRunning(ctx, id) (bool, error)` (already implemented on `*CPProvisioner`; just needs to surface on the interface). `maybeMarkContainerDead` now branches: local-Docker uses `h.provisioner.IsRunning`; SaaS uses `h.cpProv.IsRunning` which hits CP's `/cp/workspaces/:id/status`.

### Bug 2 — `RestartByID` short-circuits on `h.provisioner == nil`

Same shape as Bug 1: the auto-restart cascade triggered by `maybeMarkContainerDead` calls `RestartByID` which short-circuited when local Docker was missing. So even if Bug 1 were fixed, the workspace-offline state would never recover.

**Fix:** change the gate to `h.provisioner == nil && h.cpProv == nil` and update `runRestartCycle` to branch on which provisioner is wired for the Stop call. (The HTTP `Restart` handler at workspace_restart.go:97-135 already does this branching correctly — we're just bringing the auto-restart path to parity.)

### Bug 3 — upstream 502/503/504 propagated as-is, masked by Cloudflare

When the agent's tunnel returns 5xx (the "tunnel up but no origin" shape — agent process dead but cloudflared connection still healthy), `dispatchA2A` returns successfully at the HTTP layer with a 5xx body. `handleA2ADispatchError`'s reactive-health path doesn't run because that path is only triggered on transport-level errors. Pre-fix the code propagated the 502 status to canvas; Cloudflare in front of the platform then masked it with its own opaque \`error code: 502\` page, hiding any structured response and any Retry-After hint.

**Fix:** in `proxyA2ARequest`, when upstream returns 502/503/504, run `maybeMarkContainerDead` BEFORE propagating. If IsRunning confirms the agent is dead → return structured 503 with `restarting:true` + Retry-After (CF doesn't mask 503s the same way). If running, propagate the original status (safety check — don't recycle a healthy agent on a transient hiccup; it might have legitimately returned 502).

## Drive-by — `a2aClient` transport timeouts

`a2aClient` was `&http.Client{}` with no Transport timeouts. When a workspace's EC2 black-holes TCP connects (instance terminated mid-flight, SG flipped, NACL bug), the OS default is 75s on Linux / 21s on macOS — long enough for CF's ~100s edge timeout to fire first and surface a generic 502. Added:

- `DialContext` 10s connect
- `TLSHandshakeTimeout` 10s
- `ResponseHeaderTimeout` 60s (covers OAuth cold-start first-byte budgets like Claude Code's 30-60s)

`Client.Timeout` DELIBERATELY unset — that's a hard wall on the entire request including streamed body reads, which would pre-empt multi-minute agent synthesis. Body streaming after headers stays governed by the per-request context deadline.

## Tests

| Test | Verifies |
|---|---|
| `TestMaybeMarkContainerDead_CPOnly_NotRunning` | IsRunning(false) → marks workspace offline, returns true |
| `TestMaybeMarkContainerDead_CPOnly_Running` | IsRunning(true) → no offline-flip, returns false (safety check) |
| `TestProxyA2A_Upstream502_TriggersContainerDeadCheck` | agent returns 502 + cpProv dead → caller gets structured 503 with restarting=true + Retry-After: 15 |
| `TestProxyA2A_Upstream502_AliveAgent_PropagatesAsIs` | upstream 502 + cpProv alive → propagates 502 (existing behavior) |
| `TestMaybeMarkContainerDead_NilProvisioner` (existing) | still passes — both nil → false |
| `TestMaybeMarkContainerDead_ExternalRuntime` (existing) | still passes — external runtime → false |

Full handlers + provisioner test suites green.

## Branch-count audit

Code paths through `maybeMarkContainerDead` after fix:
1. wsRuntime == \"external\" → false ✓ (existing test)
2. both provisioners nil → false ✓ (existing test)
3. h.provisioner != nil → uses Docker IsRunning → existing tests cover dead+alive
4. h.cpProv != nil → uses CP IsRunning → **new tests cover dead+alive**

Code paths through proxyA2A's upstream 5xx handling:
1. status not in {502,503,504} → propagate (covered by `TestProxyA2A_AgentReturnsError` with 500)
2. status in {502,503,504} + agent dead → structured 503 (**new test**)
3. status in {502,503,504} + agent alive → propagate (**new test**)

All 4 cpProv-related branches and all 3 upstream-5xx branches have at least one test.

## Impact

**Pre-fix:** dead EC2 agent on a SaaS tenant → CF-masked 502 to canvas, no auto-recovery, manual restart from canvas required. Reproduction: hongmingwang.moleculesai.app workspace 01bc1def...

**Post-fix:** dead EC2 agent on a SaaS tenant → structured 503 with `restarting:true` + Retry-After to canvas, workspace flipped to offline, auto-restart cycle triggered. Canvas can show \"agent is restarting, please wait\" instead of a generic 502.

### DIFF
--- a/workspace-server/internal/handlers/a2a_proxy.go
+++ b/workspace-server/internal/handlers/a2a_proxy.go
@@ -13,6 +13,7 @@ import (
 	"errors"
 	"io"
 	"log"
+	"net"
 	"net/http"
 	"os"
 	"strconv"
@@ -92,13 +93,47 @@ func isSystemCaller(callerID string) bool {
 const maxProxyResponseBody = 10 << 20
 
 // a2aClient is a shared HTTP client for proxying A2A requests to workspace agents.
-// No client-level timeout — timeouts are enforced per-request via context
-// deadlines: canvas = 5 min (Rule 3), agent-to-agent = 30 min (DoS cap). Do NOT
-// set a Client.Timeout here: it is enforced independently of ctx deadlines and
-// would pre-empt legitimate slow cold-start flows (e.g. Claude Code first-token
-// over OAuth can take 30-60s on boot). Callers that want a safety net should
-// build a context.WithTimeout themselves.
-var a2aClient = &http.Client{}
+//
+// Timeout model — three independent budgets, none of which gets in each other's way:
+//
+//   1. Client.Timeout — DELIBERATELY UNSET. Client.Timeout is a hard wall on
+//      the entire request including streamed body reads, and would pre-empt
+//      legitimate slow cold-start flows (Claude Code first-token over OAuth
+//      can take 30-60s on boot; long-running agent synthesis can stream
+//      tokens for minutes). Total-request budget is enforced per-request
+//      via context deadline (canvas = idle-only, agent-to-agent = 30 min ceiling).
+//
+//   2. Transport.DialContext — 10s connect timeout. When a workspace's EC2
+//      black-holes TCP connects (instance terminated mid-flight, security group
+//      flipped, NACL bug), the OS default is 75s on Linux / 21s on macOS — long
+//      enough that Cloudflare's ~100s edge timeout can fire first and surface
+//      a generic 502 page to canvas. 10s is well above realistic intra-region
+//      latencies and well below CF's edge timeout.
+//
+//   3. Transport.ResponseHeaderTimeout — 60s. From request-body-end to
+//      response-headers-start. Covers cold-start first-byte (the 30-60s OAuth
+//      flow above), with margin. Body streaming after headers is governed by
+//      the per-request context deadline, NOT this timeout — so multi-minute
+//      agent responses still work fine.
+//
+// The point of (2) and (3) is to surface a *structured* 503 from
+// handleA2ADispatchError when the workspace agent is unreachable, so canvas
+// gets `{"error":"workspace agent unreachable","restarting":true}` instead
+// of Cloudflare's opaque 502 error page. Without these, dead workspaces hang
+// long enough that CF gives up first and shows its own page.
+var a2aClient = &http.Client{
+	Transport: &http.Transport{
+		DialContext: (&net.Dialer{
+			Timeout:   10 * time.Second,
+			KeepAlive: 30 * time.Second,
+		}).DialContext,
+		ResponseHeaderTimeout: 60 * time.Second,
+		TLSHandshakeTimeout:   10 * time.Second,
+		// MaxIdleConns / IdleConnTimeout: stdlib defaults are fine; agent
+		// fan-in is bounded by the platform's broadcaster fan-out, not by
+		// connection-pool sizing.
+	},
+}
 
 type proxyA2AError struct {
 	Status   int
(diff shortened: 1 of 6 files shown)


HongmingWang-Rabbit (commented):
Five-axis review (HongmingWang-Rabbit identity) — all axes pass; flagging for human eye per loop policy on substantive logic changes.

**Correctness** — three real bugs, each fixed:
- `maybeMarkContainerDead` was hard-gated on `h.provisioner != nil` → no-op for every SaaS tenant. Branch-by-which-provisioner-is-wired is the right fix.
- `RestartByID` had the same shape; gate now `provisioner==nil && cpProv==nil` matches HTTP `Restart` handler's existing branching at workspace_restart.go:97-135.
- 5xx pass-through via empty-body CF-tunnel shape: probing IsRunning + translating to structured 503 + Retry-After:15 is correct, and the alive-agent safety check (propagate as-is when cpProv reports running) prevents over-eager recycling.
- `(true, err)` contract on transient IsRunning errors keeps caller on alive path → no restart-cascade thrashing on flaky CP calls.

**Readability** — `a2aClient` triple-block comment cleanly explains the three timeout budgets (Client.Timeout DELIBERATELY unset; DialContext 10s; ResponseHeaderTimeout 60s) and why each value was picked. `maybeMarkContainerDead` docstring updated with provisioner-selection block + 2026-04-30 incident breadcrumb.

**Architecture** — surfacing `IsRunning` on `CPProvisionerAPI` (not just concrete type) is correct for uniform reactive-health across deployment shapes + testability. `fakeCPProv` panics on unexpected Start/Stop/GetConsoleOutput → loud failure if a test reaches the wrong path. Existing `coalesceRestart` idempotency preserved.

**Security** — no leaked credentials, no auth bypass, no SQL string concat, no test deletion, no `.github/workflows/*` changes. Structured 503 body exposes only "workspace agent unreachable — container restart triggered" — no info leak. Parameterised SQL preserved (`QueryRowContext` with `$1`).

**Performance** — IsRunning consultation only on 5xx branch (happy path unchanged). DialContext 10s well above intra-region latencies, well below CF's ~100s edge timeout. ResponseHeaderTimeout 60s covers Claude-Code OAuth cold-start budget.

**Branch-count audit** verified per the PR's own table:
- `maybeMarkContainerDead`: 4 paths (external, both-nil, Docker, CP) → 4 tests ✓
- `proxyA2A` 5xx: 3 paths (not 5xx, 5xx+dead, 5xx+alive) → 3 tests ✓

**CI**: 17/17 SUCCESS + 1 pending. MERGEABLE/BLOCKED.

LGTM modulo human approval. Net behavior change: SaaS tenants get reactive auto-recovery for the first time. Approving this from secondary would let it merge, but loop policy reserves substantive logic changes for the human author's call.

HongmingWang-Rabbit:
Independent review (multi-model second-opinion) caught a **Critical** gap I missed: my first commit fixed the Stop dispatch in `runRestartCycle` but left the `provisionWorkspace` dispatch unconditionally Docker-only. So on SaaS the auto-restart cycle would Stop the EC2 successfully, then NPE in `provisionWorkspace.VolumeHasFile`, get swallowed by `coalesceRestart`'s `recover()`-without-re-raise (a deliberate platform-stability safeguard), and leave the workspace permanently stuck in `status='provisioning'` because the UPDATE on workspace_restart.go:450 had already run.

Net pre-amendment effect: the headline claim "SaaS auto-restart now works" was **false** on the path the PR shipped.

**Amendment** (just pushed in `28b4e380`):

1. `runRestartCycle:476-481` — dispatch reprovision the same way every other call site does: branch on `h.cpProv != nil` → `provisionWorkspaceCP`, else `provisionWorkspace`.
2. `fakeCPProv` test helper — methods now record calls and return nil/empty rather than panicking. The previous "panic on unexpected call" pattern was unsafe because the panic fires on the async goroutine spawned by `maybeMarkContainerDead` AFTER the test assertions ran — so the broken path "passed by accident."
3. New `TestRunRestartCycle_SaaSPath_DispatchesViaCPProv` — asserts `cpProv.Stop` is called when the SaaS path runs through `runRestartCycle` directly. Would have caught the NPE if `provisionWorkspace` had been called instead.

Reviewer also flagged Important items I'm tracking but not addressing in this PR (each is a follow-up, none would have caught the Critical bug):

- **No cooldown on `maybeMarkContainerDead`** — sustained 502 fans out to CP at request rate. Per-workspace TTL gate worth adding when a real load surface emerges.
- **`activity_log` records "success" before the offline flip on dead-agent 502** — audit trail is consistent enough; net inconsistency is acceptable.
- **`ResponseHeaderTimeout: 60s`** — could legitimately starve a 61s+ cold-start. Keeping current value pending real cold-start data.

Tests: full `workspace-server` test suite green (handlers + provisioner + everything else).

HongmingWang-Rabbit:
Addressed second-reviewer findings in `a27cf8f3`:

**C1 (Critical) — test goroutine race:** Extracted `stopForRestart(ctx, id)` from `runRestartCycle` as a pure dispatcher. The new test (`TestStopForRestart_SaaSPath_DispatchesViaCPProv`) calls the helper directly so no `sendRestartContext` goroutine is spawned, eliminating the race against subsequent tests' `db.DB` writes. Verified locally with `go test -race -count=1 ./internal/handlers/...` — green. Added a `TestStopForRestart_NoProvisioner_NoOp` guard for the both-nil branch.

**I2 (Important) — CF 524:** Added `resp.StatusCode == 524` to the upstream-dead status list at `a2a_proxy.go:489-498`. Same upstream signal as 502/503/504 (origin timed out — agent unresponsive); we still gate on `IsRunning` before triggering restart.

**I1 — CI red on Platform (Go) under `-race`:** Verified the failure was the C1 cascade (single race in the new test poisoned ~30 downstream tests). Re-running on this commit.

PR #11, opened 2026-04-14
refactor(platform): split 981-line plugins.go into per-domain modules

HongmingWang-Rabbit:
## Summary

Pure mechanical split of `platform/internal/handlers/plugins.go` — no behavior changes, no test changes. Groups the `PluginsHandler` surface area by responsibility.

## Before / after

| File | Before | After | Contents |
|---|---:|---:|---|
| `plugins.go` | 981 | **194** | `PluginsHandler` struct, constructor, `With*` options, `RuntimeLookup`, `pluginInfo` + `supportsRuntime`, manifest parsing (`parseManifestYAML`, `readPluginManifest`, `strDefault`), container helpers (`findRunningContainer`, `execInContainer`/`execAsRoot`/`execInContainerAs`) |
| `plugins_sources.go` | — | **14** | `ListSources` (GET /plugins/sources) |
| `plugins_listing.go` | — | **174** | Read-only endpoints: `ListRegistry`, `listRegistryFiltered`, `ListAvailableForWorkspace`, `ListInstalled`, `CheckRuntimeCompatibility` |
| `plugins_install.go` | — | **276** | Route entrypoints for mutating ops: `Install`, `Uninstall`, `Download` |
| `plugins_install_pipeline.go` | — | **368** | Install pipeline: install-layer consts + env-var caps, `httpErr` + `newHTTPErr`, `logInstallLimitsOnce`, `dirSize`, `validatePluginName`, `installRequest`/`stageResult`, `resolveAndStage`, `deliverToContainer`, `copyPluginToContainer`, `streamDirAsTar`, `readPluginSkillsFromContainer`, `stripPluginMarkersFromMemory`, `regexpEscapeForAwk` |

Function count preserved: 32 total (was 32 in original `plugins.go`).

Every file is under the 400-line soft cap; `plugins.go` hits the ≤200 target.

### Judgment calls

- `validatePluginName` is used by every domain file; parked it in `plugins_install_pipeline.go` (alongside the other sanitization helpers) rather than `plugins.go`, purely to keep `plugins.go` under the 200-line target. Same-package, so callers don't care.
- `Install`/`Uninstall`/`Download` stay as full handler bodies in `plugins_install.go` (not thin wrappers in `plugins.go`) so the split is mechanical — moving a `func (h *PluginsHandler) Foo()` method between files in the same package is a no-op to callers.
- `plugins_sources.go` is tiny (14 lines, one handler) but kept as its own file per the grouping requested in the task.

## Test plan

[...]

PR #10, opened 2026-04-14
refactor(canvas): split 650-line FilesTab.tsx into focused components

HongmingWang-Rabbit:
## Summary
- Pure restructure of `canvas/src/components/tabs/FilesTab.tsx` (650 lines) into focused sibling files under `canvas/src/components/tabs/FilesTab/`.
- No behavior change. Top-level `FilesTab.tsx` stays at the same import path (240 lines, glue + confirmation dialogs) and re-exports `buildTree` / `TreeNode` so the existing `buildTree.test.ts` and `SidePanel.tsx` imports are unaffected.
- Preserves all dark-theme classes, aria/keyboard handlers, fetch URLs/methods/bodies, error handling, and the in-panel `ConfirmDialog`-style banners (no `confirm()`/`alert()`/`prompt()` introduced).

## Files
- `FilesTab.tsx` — 240 lines (was 650): state orchestration, layout, inline confirm banners, re-exports.
- `FilesTab/FilesToolbar.tsx` — 77 lines: root selector, New/Upload/Export/Clear/Refresh buttons.
- `FilesTab/FileTree.tsx` — 115 lines: recursive `FileTree` + `TreeItem`.
- `FilesTab/FileEditor.tsx` — 112 lines: header, save/download buttons, textarea with Ctrl/Cmd+S and Tab handling.
- `FilesTab/useFilesApi.ts` — 172 lines: GET/PUT/DELETE API wrapper, directory expansion, bulk upload/download/delete.
- `FilesTab/tree.ts` — 86 lines: `FileEntry`, `TreeNode`, `getIcon`, `buildTree` utility.

## Test plan
- [x] `cd canvas && npm run build` — clean
- [x] `cd canvas && npm test -- --run` — 357/357 pass (incl. existing `buildTree` tests)
- [x] `wc -l` — top file 240 (≤250), largest sibling 172 (≤300)
- [x] `grep -n "confirm(\|alert(\|prompt(" FilesTab*` — 0 matches
- [ ] Manual smoke test: open a workspace's Files tab, browse tree, edit/save/delete a file, upload folder, download single + export all

PR #2382, opened 2026-04-30
feat(external): default external runtime to poll-mode + awaiting_agent

HongmingWang-Rabbit:
## Summary

Paired molecule-core change for [\`molecule connect\` RFC #10](https://github.com/Molecule-AI/molecule-cli/issues/10) (molecule-cli). After this PR an external-runtime workspace's full lifecycle matches the operator-driven model:

- boots in \`awaiting_agent\` (already the case via \`workspace.go:333\`)
- CLI registers without \`--mode\` and lands in poll-mode by default (no public HTTPS needed)
- heartbeat-loss flips back to \`awaiting_agent\` (re-registrable) instead of the terminal-feeling \`offline\`

The CLI side (PRs #11/#12/#13/#14 in molecule-cli) is now merged to main; this PR closes the loop on the platform side.

### 1. \`resolveDeliveryMode\` (registry.go)

New resolution order:

| Step | Condition | Returns |
|------|-----------|---------|
| 1 | payload.delivery_mode non-empty | payload value |
| 2 | row exists, delivery_mode non-empty | preserved |
| 3 | row.runtime = "external", delivery_mode empty | **\`poll\`** *(new)* |
| 4 | otherwise | \`push\` (historical default) |

Why poll for external: laptop operators have no public HTTPS, push-mode would hard-fail at \`validateAgentURL\` anyway.

### 2. Heartbeat-loss → awaiting_agent (external only)

- **\`liveness.go\`** (Redis TTL expiration): UPDATE uses a CASE expression so the conditional is one round-trip and there's no TOCTOU between runtime read and status write.
- **\`healthsweep.go::sweepStaleRemoteWorkspaces\`** (DB-side last_heartbeat age scan): query already filters to runtime=external, so the UPDATE hard-codes the new status.
- Docker-side \`sweepOnlineWorkspaces\` keeps \`offline\` — recovery there is "restart the container", not "re-register from operator's box".

## Test plan

- [x] 9 existing \`resolveDeliveryMode\` test sites updated to the new \`SELECT delivery_mode, runtime\` query shape (sqlmock columns + AddRow values)
- [x] **New:** \`TestRegister_ExternalRuntime_DefaultsToPoll\` — external + empty mode → poll
- [x] **New:** \`TestRegister_NonExternalRuntime_StillDefaultsToPush\` — guards against the new branch overshooting (langgraph keeps push)
- [x] Liveness regex updated to match the CASE expression
- [x] \`TestSweepStaleRemoteWorkspaces_MarksStaleAwaitingAgent\` (renamed from \`_MarksStaleOffline\` for grep-ability)
- [x] Docker-side \`sweepOnlineWorkspaces\` test verified unchanged (still expects \`'offline'\`)
- [x] \`go test -race -count=1 ./internal/handlers/... ./internal/registry/...\` green (12.873s + 2.264s)

## No migration needed

\`status\` is a free-form text column. \`awaiting_agent\` is already a known value used at initial external workspace creation (\`workspace.go:333\`).

## Out of scope

- Push-mode for external (\`--mode push\` on the CLI is M4 in the RFC)
- \`/registry/register\` returning the canonical poll cursor — RFC §M1 noted but not blocking; CLI uses since_secs on first poll for now

### DIFF
--- a/workspace-server/internal/handlers/handlers_additional_test.go
+++ b/workspace-server/internal/handlers/handlers_additional_test.go
@@ -323,7 +323,7 @@ func TestRegister_ProvisionerURLPreserved(t *testing.T) {
 	handler := NewRegistryHandler(broadcaster)
 
 	// resolveDeliveryMode preflight — no row yet, default push (#2339).
-	mock.ExpectQuery("SELECT delivery_mode FROM workspaces WHERE id").
+	mock.ExpectQuery(`SELECT delivery_mode, runtime FROM workspaces WHERE id`).
 		WithArgs("ws-prov").
 		WillReturnError(sql.ErrNoRows)
 
--- a/workspace-server/internal/handlers/handlers_test.go
+++ b/workspace-server/internal/handlers/handlers_test.go
@@ -102,7 +102,7 @@ func TestRegisterHandler(t *testing.T) {
 	handler := NewRegistryHandler(broadcaster)
 
 	// resolveDeliveryMode preflight — no row yet, default push (#2339).
-	mock.ExpectQuery("SELECT delivery_mode FROM workspaces WHERE id").
+	mock.ExpectQuery(`SELECT delivery_mode, runtime FROM workspaces WHERE id`).
 		WithArgs("ws-123").
 		WillReturnError(sql.ErrNoRows)
 
--- a/workspace-server/internal/handlers/registry.go
+++ b/workspace-server/internal/handlers/registry.go
@@ -125,21 +125,30 @@ func (h *RegistryHandler) SetQueueDrainFunc(f QueueDrainFunc) {
 // Resolution order:
 //  1. payload value if non-empty (caller validated it's push/poll already)
 //  2. existing row's delivery_mode if the row exists
-//  3. "push" (the schema default — safe fallback for both new rows and
-//     a row whose delivery_mode is somehow NULL despite the NOT NULL
-//     CHECK constraint, which is forward-defensive only)
+//  3. "poll" if the existing row's runtime is "external" — most external
+//     operators run on a laptop without public HTTPS; poll is the
+//     no-public-URL path. This default flipped 2026-04-30 (issue #10
+//     in molecule-cli) when `molecule connect` shipped — push-mode
+//     stays available via explicit payload.delivery_mode="push" for
+//     VM/server operators who opt in.
+//  4. "push" (the schema default — safe fallback for non-external
+//     runtimes whose row exists with NULL delivery_mode, which is
+//     forward-defensive only)
 //
 // Returns ("", err) only on a real DB error; sql.ErrNoRows is treated
-// as "no row yet, default to push" — that's the first-register flow.
+// as "no row yet, default to push" — that's the first-register flow,
+// and at that point we don't know the runtime yet so push is the
+// historical compatible default.
 func (h *RegistryHandler) resolveDeliveryMode(ctx context.Context, workspaceID, payloadMode string) (string, error) {
 	if payloadMode != "" {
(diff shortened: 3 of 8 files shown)


HongmingWang-Rabbit (commented):
Five-axis review — endorsed for human merge.

**Correctness:** ✅ `resolveDeliveryMode` extends the SELECT to (delivery_mode, runtime) and adds a fourth resolution step (external + empty mode → poll). Order is right: payload wins, then existing-row delivery_mode (preserves explicit settings), then runtime-driven default, then push as historical fallback. `sql.NullString` for both columns handles NULL. `sql.ErrNoRows` keeps push as the default for first-register (runtime is unknown at that point — correct). Liveness UPDATE uses `CASE WHEN runtime = 'external' THEN 'awaiting_agent' ELSE 'offline' END` — single round-trip, no TOCTOU between runtime read and status write. healthsweep is already runtime=external-filtered, so hardcoding `awaiting_agent` is the right call. NOT IN guards on (removed, provisioning, paused) and (removed, paused, hibernated) preserve admin-set states.

**Readability:** ✅ Resolution-order doc table is excellent — future maintainers will know exactly which branch fires. Date stamp + RFC #10 cross-ref for archeology. `_MarksStaleOffline` → `_MarksStaleAwaitingAgent` rename for grep-ability is the right hygiene move. Liveness comment captures the user-experience rationale (operator-closed-laptop case) which is what justifies the status-flip in the first place.

**Architecture:** ✅ Aligns with the pluggable-runtime principle: external workspaces now have a coherent lifecycle (boot awaiting_agent → register → online → heartbeat-loss → awaiting_agent → re-register → online), distinct from Docker/CP-managed runtimes whose recovery path is "restart the container" (still `offline`). The CASE expression keeps the conditional in SQL where it belongs — Go code stays clean. No schema migration needed (status is free-form text and `awaiting_agent` already exists at workspace.go:333).

**Security:** ✅ Parameterized query (``). No SQL concat. No new auth surface. Tests rename rather than delete. No `.github/workflows/*` changes. No leaked credentials.

**Performance:** ✅ `resolveDeliveryMode` adds one column to an existing query — same round-trip. Liveness CASE expression keeps the non-external case at one round-trip (was already one). healthsweep query was already runtime=external-filtered.

**Test coverage:**
- ✅ `TestRegister_ExternalRuntime_DefaultsToPoll` — covers the new branch (external + empty mode → poll)
- ✅ `TestRegister_NonExternalRuntime_StillDefaultsToPush` — guards against overshoot (langgraph keeps push)
- ✅ All 9 existing resolveDeliveryMode test sites updated to the (delivery_mode, runtime) SELECT shape
- ✅ `TestRegister_PollMode_PreservesExistingValue` updated to include runtime=langgraph for the existing-mode preservation case
- ✅ `TestSweepStaleRemoteWorkspaces_MarksStaleAwaitingAgent` — renamed for grep
- ✅ Liveness regex updated to match the CASE expression
- ✅ Docker-side `TestSweepOnlineWorkspaces_DeadContainer` unchanged (still expects 'offline') — confirms the divergence is intentional
- ✅ `go test -race -count=1` green per PR body

**One question worth flagging (not a blocker):**
- After staging deploys, you'll want a one-time backfill consideration: any external workspace currently in `offline` that's still TTL-stale will only flip to `awaiting_agent` on its NEXT heartbeat-loss event. If you want to eagerly heal them, a migration-as-data-fix `UPDATE workspaces SET status = 'awaiting_agent' WHERE runtime = 'external' AND status = 'offline'` could land alongside. Optional — the operator's next `molecule connect` invocation will set it right too.

Holding the merge for the user since this is substantive net-new logic, not a mechanical diff.

Issue #2357, opened 2026-04-30
auto-promote-staging uses GITHUB_TOKEN — main pushes never trigger publish/canary/redeploy chain


HongmingWang-Rabbit:
## Root cause

\`.github/workflows/auto-promote-staging.yml\` enables auto-merge on the staging→main promote PR using \`gh pr merge --auto --merge\` with:

\`\`\`yaml
GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
\`\`\`

When the merge queue lands the promote PR, the merge action is performed on behalf of GITHUB_TOKEN. Per [GitHub's documented limitation](https://docs.github.com/en/actions/using-workflows/triggering-a-workflow#triggering-a-workflow-from-a-workflow):

> When you use the repository's GITHUB_TOKEN to perform tasks, events triggered by the GITHUB_TOKEN (with the exception of workflow_dispatch and repository_dispatch) will not create a new workflow run.

Result: the resulting \`push\` event on \`main\` is suppressed. Every workflow that depends on it stays dark.

## Observed impact (2026-04-30)

| Time (UTC) | Event | Triggered? |
|---|---|---|
| 05:48:04 | Promote PR #2352 merged into main (c140ad28) | No workflows fired |
| 06:07:29 | Promote PR #2356 merged into main (5973c9bd) | No workflows fired |

Stuck workflows:
- \`publish-workspace-server-image.yml\` — \`on: push: branches: [main]\` never fires
- \`canary-verify.yml\` — \`workflow_run\` from publish-workspace-server-image never chains
- \`redeploy-tenants-on-main.yml\` — \`workflow_run\` from publish-workspace-server-image, branches: [main] never chains

User-visible symptom: tenants stay on stale code after main updates. #2339 PRs landed on main 06:07Z; user's tenant (\`hongmingwang.moleculesai.app\`, the canary slug) was still serving pre-PR-1 code 30 min later — required a manual \`gh workflow run redeploy-tenants-on-main.yml\` to recover.

## Why staging works but main doesn't

Staging pushes come from regular human-account merges (\`gh pr merge\` with the user's PAT, not GITHUB_TOKEN). Those DO trigger downstream workflows. The asymmetry is that only the promote PRs use GITHUB_TOKEN-driven auto-merge.

## Fix options (ranked)

[...]

PR #2389, opened 2026-04-30
ci(auto-promote): dispatch publish via molecule-ai App token to unblock workflow_run chain

HongmingWang-Rabbit:
## Root cause

\`secrets.GITHUB_TOKEN\`-initiated \`workflow_dispatch\` creates the dispatched run, but the resulting run's completion event **does not fire downstream \`workflow_run\` triggers** ([no-recursion docs](https://docs.github.com/en/actions/using-workflows/triggering-a-workflow#triggering-a-workflow-from-a-workflow)). This is the well-known limitation \`actions/create-github-app-token\` exists to work around.

## Evidence (publish-workspace-server-image on main, 2026-04-30)

| run_id | head_sha | \`triggering_actor\` | canary chained? | redeploy chained? |
|---|---|---|---|---|
| 25151545007 | 6ef562ee | \`HongmingWang-Rabbit\` (PAT) | yes | yes |
| 25171773918 | 21313dc | \`github-actions[bot]\` (GITHUB_TOKEN) | **no** | **no** |
| 25173801008 | 59dec57 | \`github-actions[bot]\` (GITHUB_TOKEN) | **no** | **no** |

The 06:52Z run that "worked" was a manual operator-fired dispatch from the terminal — actor was the operator's PAT. The two runs that "dropped" were dispatched by \`auto-promote-staging.yml\`'s polling tail step authenticated via \`secrets.GITHUB_TOKEN\`. Same workflow file, same dispatch call — only the auth token differed.

## Fix

Mint a molecule-ai GitHub App installation token before the dispatch step and use it as \`GH_TOKEN\`. App-initiated dispatches DO propagate the workflow_run cascade because the App user is a real identity, not the GITHUB_TOKEN bot pseudonym.

\`\`\`yaml
- name: Mint App token for downstream dispatch
  id: app-token
  uses: actions/create-github-app-token@1b10c78c... # v3.1.1
  with:
    app-id: \${{ secrets.MOLECULE_AI_APP_ID }}
    private-key: \${{ secrets.MOLECULE_AI_APP_PRIVATE_KEY }}

- name: Wait for promote merge, then dispatch publish + redeploy
  env:
    GH_TOKEN: \${{ steps.app-token.outputs.token }}   # was: secrets.GITHUB_TOKEN
  ...
\`\`\`

## Why molecule-ai App

- **Already installed** on the Molecule-AI org (\`installation_id=124443072\`, \`app_id=3398844\`)
- Has the required \`actions:write\` permission (verified via \`GET /orgs/Molecule-AI/installations\`)
- Repository_selection: \`all\` — covers molecule-core out of the box

## ⚠️ Required setup BEFORE merging

These two repo secrets must exist at  https://github.com/Molecule-AI/molecule-core/settings/secrets/actions or auto-promote will hard-fail at the new \"Mint App token\" step:

- \`MOLECULE_AI_APP_ID\` = \`3398844\`
- \`MOLECULE_AI_APP_PRIVATE_KEY\` = contents of a .pem file generated at https://github.com/organizations/Molecule-AI/settings/installations/124443072 → \"Private keys\" → \"Generate a private key\"

(If a key has already been generated previously, you'll need to either reuse the stored \`.pem\` or generate a fresh one — old key still works in parallel for ~7 days.)

## Long-term cleanup (NOT in this PR)

The polling tail step still exists because the auto-merge call uses GITHUB_TOKEN, so the FF push to main doesn't fire publish-image's \`push\` trigger. Switching the auto-merge call to use the **same** App token would eliminate the polling tail entirely. Tracked in #2357.

## Two skipped main bumps

The 21313dc (14:41Z) and 59dec57 (15:21Z) bumps still need their tenant images redeployed. After merging this + adding secrets, fire one manually:

\`\`\`
gh workflow run redeploy-tenants-on-main.yml --ref main
\`\`\`

## Test plan

- [ ] User generates molecule-ai App private key + adds repo secrets
- [ ] PR merges via auto-promote (this same chain — meta-test)
- [ ] First post-merge auto-promote run dispatches publish-image with \`triggering_actor: molecule-ai[bot]\`
- [ ] canary-verify chains via workflow_run
- [ ] redeploy-tenants-on-main chains via workflow_run on canary success
- [ ] Verify by inspecting \`gh run view <publish-id> --json triggering_actor\` on the next staging→main bump

### DIFF
--- a/.github/workflows/auto-promote-staging.yml
+++ b/.github/workflows/auto-promote-staging.yml
@@ -267,6 +267,32 @@ jobs:
           echo "promote_pr_num=${PR_NUM}" >> "$GITHUB_OUTPUT"
         id: promote_pr
 
+      # Mint a short-lived GitHub App installation token for the dispatch
+      # step below. We CANNOT use `secrets.GITHUB_TOKEN` to dispatch the
+      # downstream publish chain — workflow runs created by GITHUB_TOKEN
+      # do not fire `workflow_run` triggers on completion (the
+      # documented "no recursion" rule —
+      # https://docs.github.com/en/actions/using-workflows/triggering-a-workflow#triggering-a-workflow-from-a-workflow).
+      #
+      # Symptom this caused (root-caused on 2026-04-30): publish-image
+      # ran successfully twice (21313dc 14:41Z, 59dec57 15:21Z) but
+      # canary-verify and redeploy-tenants-on-main never chained,
+      # because the publish run's `triggering_actor` was
+      # `github-actions[bot]` (i.e. GITHUB_TOKEN). A manual dispatch
+      # earlier in the day with the operator's PAT (d850ec7 06:52Z) did
+      # chain — same workflow file, only the actor differed.
+      #
+      # An App token's triggering_actor is the App user (e.g.
+      # `molecule-ai[bot]`), which IS allowed to fire downstream
+      # workflow_run cascades.
+      - name: Mint App token for downstream dispatch
+        if: steps.promote_pr.outputs.promote_pr_num != ''
+        id: app-token
+        uses: actions/create-github-app-token@1b10c78c7865c340bc4f6099eb2f838309f1e8c3 # v3.1.1
+        with:
+          app-id: ${{ secrets.MOLECULE_AI_APP_ID }}
+          private-key: ${{ secrets.MOLECULE_AI_APP_PRIVATE_KEY }}
+
       - name: Wait for promote merge, then dispatch publish + redeploy (#2357)
         # GITHUB_TOKEN-initiated merges suppress downstream `push` events
         # (https://docs.github.com/en/actions/using-workflows/triggering-a-workflow#triggering-a-workflow-from-a-workflow).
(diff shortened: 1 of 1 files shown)


HongmingWang-Rabbit (commented):

**Correctness.** Root-cause analysis matches the GitHub-documented behavior: `workflow_run` triggers do not cascade for runs initiated by `secrets.GITHUB_TOKEN` (the no-recursion rule). The evidence table (PAT-actor 6ef562ee chained; bot-actor 21313dc/59dec57 dropped) is the smoking gun — same workflow file, only the actor differed. App-installed identity (`molecule-ai[bot]`) is a real user, not the GITHUB_TOKEN pseudonym, so its dispatches do propagate. Fix is correct.

**Readability.** Both comment blocks updated to point at the new flow + the long-term cleanup tracker (#2357). Inline rationale on the new step links the GitHub doc and the symptom date — future maintainers will not have to re-derive this.

**Architecture.** Reuses the already-installed molecule-ai App (`installation_id=124443072`) — no new infra, no new permissions surface. The PR correctly leaves the polling tail step in place and tracks its elimination separately in #2357 (tail goes away once auto-merge ALSO uses the App token, so the FF push fires `publish-image` naturally). Right scope split.

**Security.**
- `actions/create-github-app-token@1b10c78c7865c340bc4f6099eb2f838309f1e8c3` — SHA-pinned to v3.1.1, matches org convention.
- App token scope is `actions:write` only, same surface GITHUB_TOKEN already had — no permission expansion.
- Token is consumed in a single step's `env:` and not echoed; private-key secret stays in GitHub-managed encrypted storage.
- Failure mode if `MOLECULE_AI_APP_PRIVATE_KEY` is missing: `create-github-app-token` errors at the mint step, dispatch is skipped, tail step exits non-zero. Hard-fail aligns with `feedback_schedule_vs_dispatch_secrets_hardening` for automated triggers.

**Performance.** One additional App-token mint API call per auto-promote run. Trivial.

**Action items captured by the PR body** (not in-diff, but required for green operation):
- Add repo secrets `MOLECULE_AI_APP_ID` (=`3398844`) + `MOLECULE_AI_APP_PRIVATE_KEY` BEFORE merge — otherwise the mint step hard-fails.
- After merge, fire `gh workflow run redeploy-tenants-on-main.yml --ref main` once to recover the two skipped bumps (21313dc 14:41Z, 59dec57 15:21Z) whose redeploys never chained.

LGTM. The meta-test (this PR's own merge → publish-image → canary-verify → redeploy chain succeeds) will confirm the fix end-to-end.

PR #2392, opened 2026-04-30
test(e2e): live staging regression for external-runtime awaiting_agent transitions

HongmingWang-Rabbit:
## What this is

A new staging E2E test that pins the four \`workspaces.status='awaiting_agent'\` transitions end-to-end on a real tenant. Covers exactly the silent-enum-failure class that migration 046 just fix-forwarded.

This is the test I should have had before merging PR #2382 — without it the bug shipped to prod and ran for hours before discovery during /code-review-and-quality.

## What it exercises (live, against staging)

| Step | Path | Pre-046 behavior |
|------|------|---|
| POST /workspaces (external, no URL) | \`workspace.go:333\` | UPDATE silently failed; row stuck on 'provisioning' |
| GET /workspaces/<id> | DB read after create | Returned 'provisioning' (lie about response status) |
| POST /registry/register | \`registry.go:resolveDeliveryMode\` + transition to 'online' | Worked (writes 'online', a valid enum value), but delivery_mode defaulting was wrong |
| Wait STALE window with no heartbeat | \`registry/healthsweep.go:sweepStaleRemoteWorkspaces\` | UPDATE silently failed; row stuck on 'online' indefinitely |
| Re-register | recovery from awaiting_agent → 'online' | Worked, but no test caught it |

Plus an idempotent teardown trap matching \`test_staging_full_saas.sh\`'s safety-net pattern (per memory \`feedback_schedule_vs_dispatch_secrets_hardening\`: schedule + push hard-fail when secrets missing; not silent-skip).

## Why a separate workflow (not folded into e2e-staging-saas.yml)

- The full-saas harness defaults to \`runtime=hermes\` and the external lifecycle has unique timing (90s+ stale window). Folding doubles the cold-start budget.
- Both can run in parallel — distinct slug prefixes (\`e2e-ext-...\` vs \`e2e-...\`).
- Path filter targets exactly the source touching the enum-write surface.

## Test plan

- [ ] **Manual workflow_dispatch on \`staging\` ref** to verify the test goes green against the live staging tenant fleet (which already has migration 046 applied as of 16:05Z).
- [ ] Watch the path filter trigger on a follow-up edit to one of the gated files.
- [ ] After green, evaluate promoting to a required check on staging branch protection.

### DIFF
--- a/.github/workflows/e2e-staging-external.yml
+++ b/.github/workflows/e2e-staging-external.yml
@@ -1,0 +1,164 @@
+name: E2E Staging External Runtime
+
+# Regression for the four/five workspaces.status=awaiting_agent transitions
+# that silently failed in production for five days before migration 046
+# extended the workspace_status enum (see
+# workspace-server/migrations/046_workspace_status_awaiting_agent.up.sql).
+#
+# Why this is its own workflow (not folded into e2e-staging-saas.yml):
+#   - The full-saas harness defaults to runtime=hermes, never exercises
+#     external-runtime. Adding an `external` parameter to that script
+#     would force every push to staging through both lifecycles in
+#     series, doubling the EC2 cold-start budget.
+#   - The external lifecycle has unique timing (REMOTE_LIVENESS_STALE_AFTER
+#     window, 90s default + sweep interval), which we wait through
+#     deliberately. Folding it into hermes would make the long path
+#     even longer.
+#   - It can run in parallel with the hermes E2E since both create
+#     fresh tenant orgs with distinct slug prefixes (`e2e-ext-...` vs
+#     `e2e-...`).
+#
+# Triggers:
+#   - Push to staging when any source affecting external runtime,
+#     hibernation, or the migration set changes.
+#   - PR review for the same set.
+#   - Manual workflow_dispatch.
+#   - Daily cron at 07:30 UTC (catches drift on quiet days; staggered
+#     30 min after e2e-staging-saas.yml's 07:00 UTC cron).
+#
+# Concurrency: serialized so two staging pushes don't fight for the
+# same EC2 quota window. cancel-in-progress=false so a half-rolled
+# tenant always finishes its teardown.
+
+on:
+  push:
+    branches: [staging, main]
+    paths:
+      - 'workspace-server/internal/handlers/workspace.go'
+      - 'workspace-server/internal/handlers/registry.go'
+      - 'workspace-server/internal/handlers/workspace_restart.go'
+      - 'workspace-server/internal/registry/healthsweep.go'
+      - 'workspace-server/internal/registry/liveness.go'
+      - 'workspace-server/migrations/**'
+      - 'workspace-server/internal/db/workspace_status_enum_drift_test.go'
+      - 'tests/e2e/test_staging_external_runtime.sh'
+      - '.github/workflows/e2e-staging-external.yml'
+  pull_request:
+    branches: [staging, main]
+    paths:
+      - 'workspace-server/internal/handlers/workspace.go'
+      - 'workspace-server/internal/handlers/registry.go'
+      - 'workspace-server/internal/handlers/workspace_restart.go'
+      - 'workspace-server/internal/registry/healthsweep.go'
+      - 'workspace-server/internal/registry/liveness.go'
+      - 'workspace-server/migrations/**'
+      - 'workspace-server/internal/db/workspace_status_enum_drift_test.go'
+      - 'tests/e2e/test_staging_external_runtime.sh'
+      - '.github/workflows/e2e-staging-external.yml'
+  workflow_dispatch:
+    inputs:
+      keep_org:
+        description: "Skip teardown for debugging (only via manual dispatch)"
+        required: false
+        type: boolean
+        default: false
+      stale_wait_secs:
+        description: "Seconds to wait for the heartbeat-staleness sweep (default 180 = 90s window + 90s buffer)"
+        required: false
+        default: "180"
+  schedule:
+    - cron: '30 7 * * *'
+
+concurrency:
+  group: e2e-staging-external
+  cancel-in-progress: false
+
+permissions:
+  contents: read
+
+jobs:
+  e2e-staging-external:
+    name: E2E Staging External Runtime
+    runs-on: ubuntu-latest
+    timeout-minutes: 25
+
+    env:
+      MOLECULE_CP_URL: https://staging-api.moleculesai.app
+      MOLECULE_ADMIN_TOKEN: ${{ secrets.MOLECULE_STAGING_ADMIN_TOKEN }}
+      E2E_RUN_ID: "${{ github.run_id }}-${{ github.run_attempt }}"
+      E2E_KEEP_ORG: ${{ github.event.inputs.keep_org && '1' || '0' }}
+      E2E_STALE_WAIT_SECS: ${{ github.event.inputs.stale_wait_secs || '180' }}
+
+    steps:
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
+
+      - name: Verify admin token present
+        run: |
+          if [ -z "$MOLECULE_ADMIN_TOKEN" ]; then
+            # Schedule + push triggers must hard-fail when the token is
+            # missing — silent skip would mask infra rot. Manual dispatch
+            # gets the same hard-fail; an operator running this on a fork
+            # without secrets configured needs to know up-front.
+            echo "::error::MOLECULE_STAGING_ADMIN_TOKEN secret not set (Railway staging CP_ADMIN_API_TOKEN)"
+            exit 2
+          fi
+          echo "Admin token present ✓"
+
+      - name: CP staging health preflight
+        run: |
+          code=$(curl -sS -o /dev/null -w "%{http_code}" --max-time 10 "$MOLECULE_CP_URL/health")
+          if [ "$code" != "200" ]; then
+            echo "::error::Staging CP unhealthy (got HTTP $code). Skipping — not a workspace bug."
+            exit 1
+          fi
+          echo "Staging CP healthy ✓"
+
+      - name: Run external-runtime E2E
+        id: e2e
+        run: bash tests/e2e/test_staging_external_runtime.sh
+
+      # Mirror the e2e-staging-saas.yml safety net: if the runner is
+      # cancelled (e.g. concurrent staging push), the test script's
+      # EXIT trap may not fire, so we sweep e2e-ext-* slugs scoped to
+      # *this* run id.
+      - name: Teardown safety net (runs on cancel/failure)
+        if: always()
+        env:
+          ADMIN_TOKEN: ${{ secrets.MOLECULE_STAGING_ADMIN_TOKEN }}
+        run: |
+          set +e
+          orgs=$(curl -sS "$MOLECULE_CP_URL/cp/admin/orgs" \
+            -H "Authorization: Bearer $ADMIN_TOKEN" 2>/dev/null \
+            | python3 -c "
+          import json, sys, os, datetime
+          run_id = os.environ.get('GITHUB_RUN_ID', '')
+          d = json.load(sys.stdin)
+          # Scope STRICTLY to this run id (e2e-ext-YYYYMMDD-<runid>-...)
+          # so concurrent runs and unrelated dev probes are not touched.
+          # Sweep today AND yesterday so a midnight-crossing run still
+          # cleans up its own slug.
+          today = datetime.date.today()
+          yesterday = today - datetime.timedelta(days=1)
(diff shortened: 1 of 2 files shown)


HongmingWang-Rabbit (commented):

**Correctness.** Pins the four `awaiting_agent` write paths end-to-end on a real staging tenant:
1. `workspace.go:333` (create-no-URL) → asserts both response body AND DB row read back as `awaiting_agent` (the GET-after-POST is the proof, not the response body)
2. `registry.go:resolveDeliveryMode` (register) → asserts transition to `online` + `delivery_mode=poll` (PR #2382 default)
3. `registry/healthsweep.go:sweepStaleRemoteWorkspaces` (stale window, no heartbeat) → asserts row transitions `online` → `awaiting_agent`
4. `liveness.go` (re-register) → asserts recovery `awaiting_agent` → `online`

`set -euo pipefail`, `curl --fail-with-body`, and the slug-suffixed EXIT/INT/TERM trap mirror the `test_staging_full_saas.sh` pattern. Hibernation correctly carved out as a separate concern.

**Readability.** Each step is labeled `N/8` with intent + the pre-046 silent-failure mode it protects against. Helpers (`log`/`ok`/`fail`/`tenant_call`/`admin_call`) are minimal and consistent. Docstrings on both files explain *why* this exists, not just *what*.

**Architecture.** Right call to keep this as its own workflow rather than folding into `e2e-staging-saas.yml`:
- Different default runtime (external vs hermes); folding would double cold-start budget on every staging push
- Slug prefixes (`e2e-ext-...` vs `e2e-...`) make parallel runs safe
- Path filter targets exactly the source touching the enum-write surface (5 Go files + migrations + the script + the workflow itself) — minimal trigger surface, matches the drift class
- 07:30 UTC cron staggered 30min after the existing 07:00 saas cron — won't both stomp the EC2 quota window simultaneously

**Security.**
- `actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5` SHA-pinned to v4 — matches org convention
- `permissions: contents: read` minimal scope
- `MOLECULE_STAGING_ADMIN_TOKEN` verified-present step hard-fails on schedule + push triggers; aligns with `feedback_schedule_vs_dispatch_secrets_hardening`
- Bearer token passed via env to script, never echoed in logs
- DELETE-tenant request body includes `"confirm":"$slug"` — matches the destructive-op protection on the CP admin endpoint
- Safety-net teardown is **strictly** scoped to `e2e-ext-{date}-{run_id}-*` (today + yesterday for midnight-crossing); `GITHUB_RUN_ID` empty → bail-without-deleting. Won't blast unrelated tenants. Right defense per `feedback_canvas_e2e_state_file_first` (don't pattern-sweep by date prefix alone).

**Performance.** Total wall ≈ 15min provision + 90s stale window + 90s buffer + teardown ≈ 18-20min on a happy path. Workflow `timeout-minutes: 25` has reasonable headroom. `concurrency.group: e2e-staging-external` with `cancel-in-progress: false` serializes staging+main triggers — explicit choice to protect the EC2 quota window, not the silent-cancel pattern from `feedback_concurrency_group_per_sha` (which warns about `cancel-in-progress: true` on path-filtered E2E). Acceptable trade-off.

**One tiny observation, not a blocker:** `concurrency.group` is a literal string (`e2e-staging-external`), so a manual `workflow_dispatch` will queue behind a push run rather than scope to its own ref. If you want a manual run to bypass the queue, switch to `concurrency.group: e2e-staging-external-${{ github.ref }}-${{ github.event_name }}` (or just `${{ github.run_id }}` for fully concurrent dispatches). Not required — current shape is intentional EC2-quota protection.

LGTM. The manual `workflow_dispatch` on `staging` ref is the right next step to confirm green against the live tenant fleet (which has 046 applied as of 16:05Z per the body).

HongmingWang-Rabbit (commented):
Five-axis review on the test code itself **plus** root-cause diagnosis of the failing CI.

## TL;DR

The test is correct and is catching a real production bug. The bug is **not in this PR**: it's a 4-line gate in `cmd/server/main.go` that prevents `sweepStaleRemoteWorkspaces` from ever starting on SaaS tenant deployments.

## Five-axis on the test code

**Correctness ✓.** All eight steps are well-shaped:
1. Org create + provision-wait — mirrors `test_staging_full_saas.sh:210-211` exactly (case `running` break, diagnostic burst on `failed`).
2. Per-tenant admin token retrieval + 15-min TLS readiness loop.
3. External workspace create with NO URL — pins `workspace.go:333` (the original `awaiting_agent` write site).
4. GET assertion — proves the DB row actually got `awaiting_agent`, not just the response body.
5. Register with explicit `delivery_mode=poll` — the comment at lines 461-475 is exemplary, explaining why asserting on the implicit external→poll default makes the call fragile to image-tag drift.
6. **180s wait → assert `awaiting_agent`** — this is where it's failing on staging.
7. Re-register → assert `online` recovery (operator-recoverable contract).
8. Cleanup runs in EXIT trap with leak detection.

**Readability ✓.** Long-form comments at every non-trivial step explain the WHY (which silent-failure path is being pinned). Variable names are consistent with `test_staging_full_saas.sh`.

**Architecture ✓.** Lives in its own workflow as a separate file because (a) folding into the hermes E2E doubles cold-start budget, (b) external lifecycle has unique timing, (c) parallel-runnable with hermes via slug prefix isolation. Justifications are spelled out in the workflow header.

**Security ✓.** Admin token via secrets, hard-fail on missing token (per `feedback_schedule_vs_dispatch_secrets_hardening` memory), safety-net teardown scoped strictly to this run id (today + yesterday) — no risk of touching unrelated tenants.

**Performance ✓.** 25-min job timeout. Concurrency group `e2e-staging-external` serializes runs.

## Diagnosis of the step-6 failure

The failing log line:
```
[17:41:45] ❌ After 180s with no heartbeat, expected status=awaiting_agent (sweep transition),
            got online — migration 046 likely not applied OR sweep not running
```

**It's the sweep-not-running branch.** Tracing it:

1. The sweep that does `online → awaiting_agent` is `registry.sweepStaleRemoteWorkspaces` in `internal/registry/healthsweep.go:127-183`.
2. It's invoked from `StartHealthSweep` (same file, lines 50-71). The function's docstring at lines 46-49 explicitly says: *"If `checker` is nil we still run the remote-liveness path — a deployment without Docker (e.g. a pure SaaS front-door) is a valid configuration."*
3. **But** `cmd/server/main.go:228-232` gates the entire goroutine on `prov != nil`:
   ```go
   if prov != nil {
       go supervised.RunWithRecover(ctx, "health-sweep", func(c context.Context) {
           registry.StartHealthSweep(c, prov, 15*time.Second, onWorkspaceOffline)
       })
   }
   ```
4. `prov` initialization at `main.go:135-155`: in SaaS mode (`MOLECULE_ORG_ID != ""`), only `cpProv` is set — `prov` stays nil. The Docker provisioner is intentionally not initialized because tenants don't run their workspaces locally; CP does the EC2 dance.
5. **Net effect**: on every SaaS tenant, the entire health-sweep goroutine never starts. `sweepStaleRemoteWorkspaces` never runs. External workspaces stay `online` forever even after their agent crashes.

**Git blame on the gate:** `git log -L /StartHealthSweep/,/^}/:cmd/server/main.go` shows the `if prov != nil` has been there since the initial commit — predates the Phase 30.7 addition of `sweepStaleRemoteWorkspaces`. The function was correctly hardened to handle a nil checker (April 2026), but main.go's gate was never updated to match. Classic dead-letter-of-the-spec.

This bug has been live in production for the entire window since #2382 merged (2026-04-30 14:43Z). Every external-runtime workspace on SaaS staging is currently in a state where heartbeat-loss does NOT transition to awaiting_agent. The drift gate (#2388) caught the migration side; the silent-write bug fix (#2382's hibernation/awaiting_agent paths) caught some of the SQL surface — but this orchestration-side bug slipped through both because there was no E2E coverage on this lifecycle path. **#2392 is the gate that catches it.**

## Suggested fix (separate PR)

The minimal patch is at `cmd/server/main.go:228`:
```go
// Before:
if prov != nil {
    go supervised.RunWithRecover(ctx, "health-sweep", func(c context.Context) {
        registry.StartHealthSweep(c, prov, 15*time.Second, onWorkspaceOffline)
    })
}

// After (always start; healthsweep.go already handles nil checker):
go supervised.RunWithRecover(ctx, "health-sweep", func(c context.Context) {
    registry.StartHealthSweep(c, prov, 15*time.Second, onWorkspaceOffline)
})
```

Test coverage for the SaaS path already exists: `healthsweep_test.go::TestStartHealthSweep_NilCheckerRunsRemoteSweep` (lines 273-302) explicitly covers nil checker. So removing the gate is safe — the function-level test is already pinning the behavior.

## Verdict

The test code itself is approve-shaped. The CI failure is a true positive that surfaces a real production bug. **Action:** I'll open a follow-up PR removing the `prov != nil` gate; once that's merged + tenants redeploy, this test will pass and #2392 becomes mergeable.

(I can't approve my own PR — flagging for the next reviewer that the test is correct, and the failing red is informational pending the orchestration fix.)

HongmingWang-Rabbit (commented):
## Five-axis review

### Context understood
This is the staging E2E that pins the four `workspaces.status='awaiting_agent'` transitions on a real tenant — the test that should have existed before #2382 shipped. The 17:41Z CI failure was exactly the bug class it's testing (`got online` after 180s of no heartbeat = sweep silently failed). That bug was fixed in #2408 (merged 5 min ago at `b5c7b34`), so I've kicked off a fresh run via `workflow_dispatch` to verify green now that staging has the fix.

### Correctness ✓
- Each transition is double-asserted: response body status AND a follow-up GET against the DB row. Catches the "response lies" bug class directly.
- Idempotent teardown trap with leak-detection (counts non-purged rows, exit 4 if found) — proven pattern from `test_staging_full_saas.sh`.
- Step 5b cross-checks `delivery_mode=poll` from the **register** response rather than the workspace GET, with a clear inline note about why (the GET handler's column list pre-dates the `delivery_mode` column from #2339 PR 1). Decoupling from image-tag drift on the staging fleet is the right call.
- Hibernation explicitly out of scope, with rationale (different timing model). Future hibernation harness is filable as a separate issue.

### Readability ✓
- Each step's docstring cites which line of which file silently failed pre-046, plus what the post-fix expectation is. A future debugger lands on the right file in seconds.
- The "id vs workspace_id" pitfall comment at step 5 (lines 261-265) saves the next maintainer from a mystery 400. That's the kind of comment that earns its keep.
- Numbered progress (1/8 through 8/8) makes the workflow log readable mid-run.

### Architecture ✓
- Separating from `e2e-staging-saas.yml` is justified — hermes vs external have different cold-start budgets, and folding would double the EC2 lifecycle cost. Concurrency group prevents two staging pushes from competing for the same quota window.
- Path filter scope matches exactly the source touching the awaiting_agent surface — won't fire on unrelated pushes.
- Schedule trigger at 07:30 UTC, staggered 30 min after the hermes E2E's 07:00 cron — good neighbor.
- Workflow safety net mirrors the hermes harness pattern (run-id-scoped slug sweep), per the memory `feedback_canvas_e2e_state_file_first` style.

### Security ✓
- Hard-fail (`exit 2`) on missing `MOLECULE_STAGING_ADMIN_TOKEN` — no silent-skip on schedule/push, per `feedback_schedule_vs_dispatch_secrets_hardening`. Manual dispatch gets the same hard-fail (correct: a fork without secrets needs to know).
- Admin token only in `Authorization: Bearer` headers, not URLs — won't leak in CI logs.
- All curl calls bounded with `--max-time` and `--fail-with-body`. No hangs.
- Slug derivation prefixed with run id and date-scoped — sweep can never touch unrelated tenants.

### Performance ✓
- 25-min workflow timeout vs ~20-min test budget (15-min provisioning ceiling + 3-min stale wait + register/heartbeat). Reasonable.
- Single tenant per run; cleanup leaves no resources behind.
- Schedule + push triggers are path-filtered → minimal background load.

### Findings

| Severity | Item |
|---|---|
| **FYI** | Previous CI run failed at the exact transition #2408 fixes. New run kicked off now (#25187...01) — expect green. |
| **Nit** | Workflow timeout 25min vs test budget ~20min is tight. If staging EC2 cold-start ever balloons (memory says hermes cold-start is ~7min), we're 5min from a false-failure. **Consider** bumping to 30min for headroom. Not blocking. |
| **FYI** | Inline comments cite `workspace.go:333` and `healthsweep.go` line numbers — those will rot. Acceptable for an E2E harness; CI failures will surface the actual error from the live service rather than relying on the line numbers. |
| **Nit** | Step 5 sets `delivery_mode: "poll"` explicitly to dodge image-tag drift (well-reasoned, with full rationale on lines 268-276), and verifies the value via the register response echo. Adds a small dependency on the register-response shape, but it's the right trade-off vs gating on the implicit default. |

### Dependency review
No new dependencies. Uses bash + curl + python3, all already standard in CI.

### Verdict
**Approve** pending the in-flight workflow run going green. Five axes all clean, change is the right shape and size, and the previous failure mode is exactly what #2408 fixes — so the green should land on this run.

PR #2396, opened 2026-04-30
refactor(workspace-status): typed constants + AST-based drift gate

HongmingWang-Rabbit:
## Summary
Defense-in-depth follow-up to #2388 (migration 046, which added `awaiting_agent`+`hibernating` to the `workspace_status` enum after they had been silently failing in prod for ~5 days).

- New `internal/models/workspace_status.go` — `WorkspaceStatus` typed string alias + `AllWorkspaceStatuses` canonical slice + per-state `Status*` constants
- All 18 production status-write sites refactored to parameterized `$N` typed values (`SET status = $N` instead of `SET status = 'literal'`); status sourced from `models.Status*` constants
- Drift gate (`workspace_status_enum_drift_test.go`) now uses Go AST (`go/parser` + `go/ast`) + SQL parser instead of regex, asserting:
  - codebase `AllWorkspaceStatuses` ⊆ migration enum (catches the #2388-style failure)
  - every `Status*` const appears in the canonical slice
  - every slice entry resolves to a known const

## Why this matters
The 5-day silent enum-mismatch outage was invisible because sqlmock matches SQL by regex and never enforces live enum constraints. With status as a typed parameterized value:
1. A typo in a constant fails compilation, not at runtime.
2. Adding a new state requires updating both the slice and the migration — drift gate catches mismatches in CI.
3. Future enum drift surfaces in tests as `pq: invalid input value for enum`, not as silent prod 500s.

This is the type-safe + static-AST half of the layered-defense plan; PR #2392 (live-staging external-runtime regression E2E) is the runtime half.

## Test plan
- [x] `go test ./... -count=1 -race` — green
- [x] `go vet ./...` — clean
- [x] Drift gate (`TestWorkspaceStatusEnum_NoLiteralDrift`) — green
- [ ] Staging E2E once #2392 lands

### DIFF
--- a/workspace-server/internal/bundle/importer.go
+++ b/workspace-server/internal/bundle/importer.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/Molecule-AI/molecule-monorepo/platform/internal/db"
 	"github.com/Molecule-AI/molecule-monorepo/platform/internal/events"
+	"github.com/Molecule-AI/molecule-monorepo/platform/internal/models"
 	"github.com/Molecule-AI/molecule-monorepo/platform/internal/provisioner"
 	"github.com/google/uuid"
 )
@@ -131,7 +132,8 @@ func buildBundleConfigFiles(b *Bundle) map[string][]byte {
 
 func markFailed(ctx context.Context, wsID string, broadcaster *events.Broadcaster, err error) {
 	db.DB.ExecContext(ctx,
-		`UPDATE workspaces SET status = 'failed', updated_at = now() WHERE id = $1`, wsID)
+		`UPDATE workspaces SET status = $1, updated_at = now() WHERE id = $2`,
+		models.StatusFailed, wsID)
 	broadcaster.RecordAndBroadcast(ctx, "WORKSPACE_PROVISION_FAILED", wsID, map[string]interface{}{
 		"error": err.Error(),
 	})
--- a/workspace-server/internal/db/workspace_status_enum_drift_test.go
+++ b/workspace-server/internal/db/workspace_status_enum_drift_test.go
@@ -51,6 +51,7 @@ func TestWorkspaceStatusEnum_NoLiteralDrift(t *testing.T) {
 	repoRoot := findRepoRoot(t)
 	migrationsDir := filepath.Join(repoRoot, "workspace-server", "migrations")
 	statusFile := filepath.Join(repoRoot, "workspace-server", "internal", "models", "workspace_status.go")
+	srcRoot := filepath.Join(repoRoot, "workspace-server")
 
 	enum := loadWorkspaceStatusEnum(t, migrationsDir)
 	if len(enum) == 0 {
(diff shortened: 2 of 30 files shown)


HongmingWang-Rabbit (commented):
Five-axis pass. Substantive refactor (28 files, 18 production write sites + AST drift gate) → comment-only per loop policy.

**Correctness.** The typed-constants approach closes the 5-day silent-enum class at the right layer: the compiler now catches typos that previously sailed through sqlmock's regex match. The 10 declared constants (`StatusProvisioning`, `StatusOnline`, …, `StatusAwaitingAgent`, `StatusHibernating`) cover migrations 043 + 046. Healthsweep diff confirms the pattern — `UPDATE workspaces SET status = $1, updated_at = now() WHERE id = $2 …` with the value sourced from `models.Status*`. Underlying type `string` preserves database/sql + JSON marshalling.

**Readability.** The `models/workspace_status.go` docstring is exemplary — explains why typed (not just what), cites the 5-day outage incident, lists the three failure mechanisms (ad-hoc literals + sqlmock regex match + dropped errors). The drift gate's docstring includes explicit add-status / retire-status runbooks. The `String()` method spares callers from casting.

**Architecture.** Right placement (`internal/models`), single source of truth (`AllWorkspaceStatuses` slice). The shift from regex-grepping every literal to AST-parsing the const block + slice aligns with `feedback_behavior_based_ast_gates` ("behavior-based gates beat name-list gates"). Pairing with #2392's runtime E2E gives layered defense: this catches drift at compile + CI, #2392 catches drift in live staging behavior.

**Security.**
- No credentials, no auth-bypass; this PR's whole motion is removing string concatenation of status values
- No deleted production tests — actually expands test coverage with new constants used in `*_test.go` updates across 12 test files
- No `.github/workflows/*` changes
- Parameterized `$N` writes are strictly safer than literal-interpolation

**Performance.** Test runs go/parser + a sqlparser at CI time — milliseconds. Production paths: parameterized writes vs literal-string writes have identical wire cost; the typed constant adds zero overhead (string under the hood).

**One observation worth tracking, not a blocker:** the previous gate (now removed) scanned every `.go` file for raw `'X'` SET-status literals — that caught both write-side enum drift AND any future regression where a developer wrote `SET status = 'oops'` directly. The new gate only checks `AllWorkspaceStatuses ⊆ enum`, so it does NOT enforce "writes must use typed constants." In practice the parameterization convention is now well-established and a future regressor would be caught at code review, but if you ever want belt-and-suspenders, a small companion test that re-greps for `SET status =\s*'` on the prod tree and asserts zero matches would close the convention loop. (Healthsweep example shows read-side `status NOT IN ('removed', 'provisioning')` literals still exist — those don't risk silent-write failures, but a typo there would silently match zero rows.)

Body claim that "literals are forbidden / fails the build if a literal sneaks in" is slightly aspirational under the current gate — the gate now enforces the constants ⇄ enum bridge, but does not enforce the "no literals at write sites" half. Worth tightening the docstring on `workspace_status.go` if you want the comment to match the test.

LGTM. CI checks (Analyze Go/JS/Python + Platform-Go + E2E Staging SaaS) are still in progress; once green this is straightforwardly mergeable.

HongmingWang-Rabbit:
Pushed `188db337` to address the literal-write enforcement gap you flagged. The drift gate now has a second axis: it walks production `.go` AST, finds string literals containing `UPDATE workspaces … SET status = '<lit>'`, and fails with `file:line → SET status = '<lit>'`.

I verified it actually fires by injecting a sentinel `_drift_sanity.go` with a hard-coded `SET status = 'failed'` literal — gate flagged it correctly, then I removed the sentinel.

Caught one real miss in the original sweep: `workspace_bootstrap.go:62` (the `BootstrapFailed` admin handler) was still writing `'failed'` as a literal. Fixed in the same commit, with `bootstrap_failed_test.go` updated to match the parameterized shape.

Updated the `workspace_status.go` docstring to match what the gate now actually enforces.

Issue #2395, opened 2026-04-30
redeploy-fleet + tenant provision serve stale workspace-server images even with fresh :latest


HongmingWang-Rabbit:
## Symptom

Two independent observations, same root cause:

**(1) Production tenant — user-visible, 2026-04-30 ~17:30Z**

User \`hongmingwang.moleculesai.app\` tried to attach an image in canvas chat. Got:

\`\`\`
Upload failed: 503
{"detail":"Reprovisioning the workspace will mint the platform_inbound_secret it's missing.",
 "error":"workspace not yet enrolled in v2 upload (RFC #2312)"}
\`\`\`

The exact \`detail\` string only exists in commit \`e632a313\` (Apr 29, RFC #2312 PR-C). It was replaced in \`501a42d7\` (\`refactor(chat_files): extract resolveWorkspaceForwardCreds shared by Upload+Download\`) which **also added inline lazy-heal**. Both \`501a42d7\` (Apr 30) and \`30a569c7\` (helper extraction) are on \`origin/main\` HEAD \`26fecb90\`.

\`publish-workspace-server-image\` republished \`:latest\` on main at 16:56:56Z (run 25178314632, 3m20s, success). \`redeploy-tenants-on-main\` then fired at 17:00:53Z (run 25178465532) and reported \`{"slug":"hongmingwang","ssm_status":"Success"}\`. Yet 30 min later the deployed code path is still pre-\`501a42d7\`. **The redeploy claimed success but the running container did not actually swap.**

**(2) Fresh E2E tenant — CI-visible, same day**

\`E2E Staging External Runtime\` (PR #2392) test-run \`25179171433\` failed at step 5 with:

\`\`\`
{"error":"hostname \"example.invalid\" cannot be resolved (DNS error): lookup example.invalid on 127.0.0.53:53: no such host"}
HTTP_CODE=400
\`\`\`

Test sent \`/registry/register\` with no \`delivery_mode\`, expecting \`resolveDeliveryMode\`'s \`runtime=external\`→\`poll\` branch (PR #2382, on main since 2026-04-30) to skip URL validation. The DNS lookup is what \`validateAgentURL\` does for push mode. So the EC2 instance the freshly-provisioned tenant booted was **also running pre-#2382 workspace-server**, despite being provisioned from \`:latest\` minutes after publish.

Same class as (1): a current \`:latest\` GHCR tag, a deploy/provision step that reports success, and a host that's running last week's image.

## Hypothesis

[...]

PR #2399, opened 2026-04-30
feat(workspace): surface peer-discovery failure reason instead of "may be isolated"

### LINKED ISSUE #2397: list_peers + get_peers swallow non-200 + exceptions, leaving agent + user blind to root cause
## Symptom

User has two workspaces visible in canvas:

- \`Claude Code Agent\` (T2 Standard, runtime=claude-code, status=online)
- \`mac test\` (T4, REMOTE)

Asks the agent: "did you actually call MCP?". Agent calls \`mcp__a2a__list_peers\`, gets back:

> \"No peers available (this workspace may be isolated)\"

The agent then says: \"so the MCP call went through successfully, it just didn't find any peers to report.\" User pushes back: \"but I can see you have one peer named mac test\". Agent has no further diagnostic info to offer — by the time it sees the empty result, the actual reason is lost.

### 1. \`workspace/a2a_client.py:232\` — \`get_peers()\`

\`\`\`python
async def get_peers() -> list[dict]:
        try:
            resp = await client.get(
                f\"{PLATFORM_URL}/registry/{WORKSPACE_ID}/peers\",
                headers={...},
            )
            if resp.status_code == 200:
                return resp.json()
            return []          # ← swallows 401/403/404/5xx
        except Exception:
            return []          # ← swallows network errors, JSON decode errors
\`\`\`

### 2. \`workspace/builtin_tools/a2a_tools.py:16\` — \`list_peers()\`

Same pattern, second copy. Two-copy drift hazard.

### 3. \`workspace/a2a_tools.py:411\` — \`tool_list_peers()\`

\`\`\`python
async def tool_list_peers() -> str:
    peers = await get_peers()
    if not peers:
        return \"No peers available (this workspace may be isolated)\"
    ...
\`\`\`

The \"may be isolated\" hint is actually wrong half the time — \`peers == []\` covers four distinct conditions:

| Condition | What user/agent sees | What's actually true |
|---|---|---|
| Discovery 200 OK, empty result | \"may be isolated\" | True — no parent, no children, no siblings |
| Discovery 401/403 | \"may be isolated\" | Workspace token bad — operator action needed |
| Discovery 404 | \"may be isolated\" | Workspace ID not in registry — re-register needed |
| Discovery 5xx | \"may be isolated\" | Platform error — operator action needed |
| Network error | \"may be isolated\" | PLATFORM_URL unreachable — operator action needed |

All five collapse to the same message. The user can't tell whether they need to:
- Provision more workspaces (true isolation)
- Re-register (404)
- Restart the workspace (auth)
- Page on-call (platform 5xx)
- Check network (timeout)

## Why this is the right thing to fix now

User just reported this on their staging tenant. Same observability category as issue #2395 (deploy claimed success / nothing actually changed) — one is deploy-side, this is runtime-side, both have the property that operator + agent both say \"green\" while the user is stuck.

## Action

Two-prong fix in \`workspace/a2a_client.py\` and \`workspace/builtin_tools/a2a_tools.py\`:

1. **New helper that returns diagnostic context**:
   \`\`\`python
   async def get_peers_with_diagnostic() -> tuple[list[dict], str | None]:
       \"\"\"Returns (peers, error_summary). error_summary is None on success;
       a short human-readable string when peers is empty for a non-trivial reason.\"\"\"
   \`\`\`
   Existing \`get_peers()\` and \`list_peers()\` continue to return \`list[dict]\` so non-tool callers don't break.

2. **Update \`tool_list_peers()\`** to call the diagnostic helper and surface the actual reason:
   - True empty: \"You have no peers in the platform registry. (No parent, no children, no siblings registered.)\" — drop the \"may be isolated\" hint.
   - 401/403: \"Authentication to platform failed. The workspace may need restarting.\"
   - 404: \"This workspace's ID is not registered with the platform. Re-registration needed.\"
   - 5xx: \"Platform error: HTTP \<code\>.\"
   - Network: \"Cannot reach platform at \<url\>: \<error\>.\"

3. **Tests**: extend \`test_a2a_tools_module.py\` to cover each of the five branches above. The existing tests already mock 200 / 404 / exception — adding 401, 5xx, and timeout is one new test class.

## Constraints / non-goals

- Don't change the \`Peers\` SQL in \`discovery.go\` — that's separate. The discovery scope (parent + siblings + children, no org-wide visibility for unrelated roots) might or might not be the right behavior, but it's a design decision and orthogonal to surfacing-the-actual-failure-reason.
- Don't add platform-side logging that exposes peer IDs to other workspaces (existing C18 token check at \`discovery.go:222\` is the gate; honor it).

## Repro

Two workspaces on a tenant, in different parent_id trees (e.g. both root with no relation, or one is a child of a deleted parent). Call \`mcp__a2a__list_peers\` from the first. Today: \"may be isolated\". Expected after fix: \"You have no peers in the platform registry.\" (or whatever specific reason applies).

## Related

- #2395 — deploy-side observability (same shape, different layer)
- Memory: \`feedback_chase_verification_to_staging\` — the systemic version of \"don't stop at success-shaped output without verifying user-visible behavior\"

HongmingWang-Rabbit:
## Summary

Closes #2397. Today, every empty-peer condition collapses to one ambiguous string: *"No peers available (this workspace may be isolated)"*. The user has no way to tell five different conditions apart:

| Condition | What user/agent sees today | What's actually true |
|---|---|---|
| Discovery 200 + empty | "may be isolated" | True — no parent, no children, no siblings |
| Discovery 401/403 | "may be isolated" | Workspace token bad — restart |
| Discovery 404 | "may be isolated" | Workspace ID not in registry — re-register |
| Discovery 5xx | "may be isolated" | Platform error — page on-call |
| Network error | "may be isolated" | PLATFORM_URL unreachable — check network |

Five different operator actions, one ambiguous string. This is the runtime-side cousin of #2395 (deploy claimed success / nothing actually changed): success-shaped output that hides the actual failure.

## How

- New `get_peers_with_diagnostic() -> tuple[list[dict], str | None]` in `workspace/a2a_client.py`. Returns `(peers, None)` on 200; `(peers=[], diagnostic="…")` on every other branch with an actionable string.
- Existing `get_peers()` shims through the new helper, discarding the diagnostic — preserves the bare-list contract for non-tool callers (e.g. `get_peers_summary` for system-prompt bootstrap).
- `tool_list_peers()` in `workspace/a2a_tools.py` switches to the diagnostic helper and surfaces the actual reason. The misleading "may be isolated" string is gone; true empty now reads "no peers in the platform registry."

## Tests

`TestGetPeersWithDiagnostic` (new) covers all branches:
- 200 with peers
- 200 with empty list
- 401 / 403 (auth)
- 404 (not registered)
- 503 (platform error)
- network exception
- 200 with non-list body (regression guard)
- bare-list shim still returns `list[dict]`

`TestToolListPeers` (rewritten) asserts each diagnostic branch surfaces the reason AND that "may be isolated" never appears anywhere.

`go ./... clean`, 122 a2a tests pass, coverage 91.53% (floor 86%).

## Out of scope

Not changing `Peers` SQL in `discovery.go` — that's a separate design question (parent_id-scoped vs org-wide). This PR is strictly about surfacing the actual failure reason when discovery is reachable.

The duplicate `list_peers` in `workspace/builtin_tools/a2a_tools.py` is left alone — it has no active prod consumers (only set up as a stub in conftest.py), so updating it now would be churn for no benefit.

## Test plan

- [x] Unit tests pass locally (`pytest workspace/tests/`)
- [ ] Workspace runtime publishes via publish-runtime.yml on staging push
- [ ] On staging, ask the agent `mcp__a2a__list_peers` from a workspace whose registry returns 404 — expect "Re-registration needed" instead of "may be isolated"

### DIFF
--- a/workspace/a2a_client.py
+++ b/workspace/a2a_client.py
@@ -229,19 +229,61 @@ async def send_a2a_message(target_url: str, message: str) -> str:
     return _format_a2a_error(last_exc, target_url)
 
 
-async def get_peers() -> list[dict]:
-    """Get this workspace's peers from the platform registry."""
+async def get_peers_with_diagnostic() -> tuple[list[dict], str | None]:
+    """Get this workspace's peers, returning (peers, diagnostic).
+
+    diagnostic is None when the call succeeded (status 200, even if the list
+    is empty). When peers is [] for a non-trivial reason (auth failure,
+    workspace-id missing from registry, platform error, network error),
+    diagnostic is a short human-readable string explaining what went wrong
+    so callers can surface it instead of "may be isolated" — see #2397.
+
+    The legacy get_peers() shim below preserves the bare-list contract for
+    non-tool callers.
+    """
+    url = f"{PLATFORM_URL}/registry/{WORKSPACE_ID}/peers"
     async with httpx.AsyncClient(timeout=10.0) as client:
         try:
             resp = await client.get(
-                f"{PLATFORM_URL}/registry/{WORKSPACE_ID}/peers",
+                url,
                 headers={"X-Workspace-ID": WORKSPACE_ID, **auth_headers()},
             )
-            if resp.status_code == 200:
-                return resp.json()
-            return []
-        except Exception:
-            return []
+        except Exception as e:
+            return [], f"Cannot reach platform at {PLATFORM_URL}: {e}"
+
+        if resp.status_code == 200:
+            try:
+                data = resp.json()
+            except Exception as e:
+                return [], f"Platform returned 200 but body was not JSON: {e}"
+            if not isinstance(data, list):
+                return [], f"Platform returned 200 but body was not a list: {type(data).__name__}"
+            return data, None
+
+        if resp.status_code in (401, 403):
+            return [], (
+                f"Authentication to platform failed (HTTP {resp.status_code}). "
+                "The workspace bearer token may be invalid — restarting the workspace usually re-mints it."
+            )
+        if resp.status_code == 404:
+            return [], (
+                f"Workspace ID {WORKSPACE_ID} is not registered with the platform (HTTP 404). "
+                "Re-registration via the platform's /registry/register endpoint is needed."
+            )
+        if 500 <= resp.status_code < 600:
+            return [], f"Platform error: HTTP {resp.status_code}."
+        return [], f"Unexpected platform response: HTTP {resp.status_code}."
+
+
+async def get_peers() -> list[dict]:
+    """Get this workspace's peers from the platform registry.
+
+    Bare-list shim over get_peers_with_diagnostic() — discards the diagnostic
+    so callers that don't care about the failure reason (e.g. system-prompt
+    bootstrap formatters) get the same shape they always had.
+    """
+    peers, _ = await get_peers_with_diagnostic()
+    return peers
 
 
 async def get_workspace_info() -> dict:
--- a/workspace/a2a_tools.py
+++ b/workspace/a2a_tools.py
@@ -18,6 +18,7 @@ from a2a_client import (
     _peer_names,
     discover_peer,
     get_peers,
+    get_peers_with_diagnostic,
     get_workspace_info,
     send_a2a_message,
 )
@@ -410,9 +411,16 @@ async def tool_send_message_to_user(message: str, attachments: list[str] | None 
 
 async def tool_list_peers() -> str:
     """List all workspaces this agent can communicate with."""
-    peers = await get_peers()
+    peers, diagnostic = await get_peers_with_diagnostic()
     if not peers:
-        return "No peers available (this workspace may be isolated)"
+        if diagnostic is not None:
+            # Non-trivial empty: auth failure / 404 / 5xx / network — surface
+            # the actual reason so the user/agent doesn't have to guess. #2397.
+            return f"No peers found. {diagnostic}"
+        return (
+            "You have no peers in the platform registry. "
+            "(No parent, no children, no siblings registered.)"
+        )
     lines = []
     for p in peers:
         status = p.get("status", "unknown")
--- a/workspace/tests/test_a2a_client.py
+++ b/workspace/tests/test_a2a_client.py
@@ -578,6 +578,149 @@ class TestGetPeers:
 
 
 # ---------------------------------------------------------------------------
+# get_peers_with_diagnostic — issue #2397
+#
+# Pin: an empty peer list MUST come with an actionable diagnostic on every
+# non-200 + every transport failure. The bug was that get_peers swallowed
+# every failure mode behind `return []`, leaving the agent's tool wrapper
+# with no way to distinguish "you have no peers" from "auth broke" / "404
+# from registry" / "platform 5xx" / "network timeout". Each of these
+# requires a different operator action.
+# ---------------------------------------------------------------------------
+
+class TestGetPeersWithDiagnostic:
+
+    async def test_200_returns_peers_and_no_diagnostic(self):
+        """200 with valid list → (peers, None). diagnostic stays None on success."""
+        import a2a_client
+
+        peers = [{"id": "ws-1", "name": "Alpha"}]
+        resp = _make_response(200, peers)
+        mock_client = _make_mock_client(get_resp=resp)
+
+        with patch("a2a_client.httpx.AsyncClient", return_value=mock_client):
+            result, diag = await a2a_client.get_peers_with_diagnostic()
+
+        assert result == peers
+        assert diag is None
+
+    async def test_200_empty_list_returns_no_diagnostic(self):
+        """200 with [] → (peers=[], diag=None). Truly no peers is success, not error."""
+        import a2a_client
+
+        resp = _make_response(200, [])
+        mock_client = _make_mock_client(get_resp=resp)
+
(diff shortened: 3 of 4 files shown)


HongmingWang-Rabbit (commented):
Five-axis pass. Substantive logic change → comment-only per loop policy.

**Correctness.** Right idea cleanly executed: every previously-collapsed empty-peer condition now returns a distinct, actionable diagnostic instead of all flowing into "may be isolated." Branch coverage is thorough — 200, 200-with-non-list-body, 401/403, 404, 5xx, network exception, and catch-all. The 200-with-non-list-body regression guard is a nice touch (body-schema drift is exactly the kind of silent-success failure the PR is fighting). The shim approach for `get_peers()` preserves the bare-list contract for `get_peers_summary` (system-prompt bootstrap) without ripple changes.

**Readability.** Docstrings on both functions explain the contract AND link the user-visible motivation (#2397). Branch ordering reads top-down (network → 200 → auth → not-registered → 5xx → unknown), so a maintainer can scan it without thinking. Each diagnostic string includes the operator's next action ("restart usually re-mints," "re-registration needed," etc.) — good operator UX.

**Architecture.** Right shape: rename existing helper to a `_with_diagnostic` variant + slim shim for legacy callers. This avoids the alternative of either (a) breaking the bare-list signature on the system-prompt path, or (b) bolting an `Optional[diagnostic]` parameter onto the legacy function. The "out of scope" notes are honest — `discovery.go` Peers SQL and the duplicate `builtin_tools/a2a_tools.py:list_peers` are correctly punted.

**Security.**
- No credentials, no auth-bypass; `auth_headers()` continues to flow through `platform_auth` (the typed signature gate from #2387 guards the surface).
- 404 diagnostic includes `WORKSPACE_ID` — that's tenant-scoped, not secret, already in logs; no leak.
- No SQL string concat (this is HTTP-only).
- No deleted tests — actually expands `TestGetPeersWithDiagnostic` + rewritten `TestToolListPeers`.
- No `.github/workflows/*` changes.

**Performance.** Same single HTTP call as before; only the post-response branch logic changed. The `tuple[list[dict], str | None]` return adds ~80 bytes of allocation per call vs the old `list[dict]` — well under any conceivable hot path budget for a registry/peers lookup.

LGTM. CI checks (Analyze Go/JS/Python) still in progress; once green this is straightforwardly mergeable.

HongmingWang-Rabbit (commented):
Five-axis pass. Real fix for the success-shaped-failure-hides-the-actual-reason bug class — the runtime cousin of #2395.

**Correctness.** Walked all six branches in `a2a_client.py` against the new tests:
- 200 + valid list → `(peers, None)` ✓ (test: `test_200_returns_peers_and_no_diagnostic`)
- 200 + empty → `([], None)` — correctly classified as success, not failure ✓
- 200 + non-list → `([], "...not a list")` ✓ (regression guard for body-shape drift)
- 401/403 → auth diagnostic + restart hint ✓
- 404 → registration hint ✓
- 5xx → "Platform error" ✓
- Network exception → unreachable diagnostic with PLATFORM_URL + underlying error ✓

The split between `client.get()` (network) and `resp.json()` (parse) try/except blocks is exactly right — distinct error sources get distinct diagnostics instead of one catch-all.

**Branch coverage gap (Optional/Consider).** Production code has two branches without explicit tests:
1. `200 + invalid JSON` — handled by `except Exception as e: return [], f"...not JSON: {e}"` but no `test_200_with_invalid_json` (the `test_200_with_non_list_body` uses a valid JSON dict, not a JSON decode error).
2. The catch-all `return [], f"Unexpected platform response: HTTP {resp.status_code}."` — fires on 4xx other than 401/403/404 (e.g. 429, 451). No test pins this.

Both are belt-and-suspenders. The PR body claim "all branches" is ~9/11 by my count. Worth two more 4-line tests if you want clean 100%.

**Readability.** The `get_peers_with_diagnostic` docstring is exemplary — it explains the return shape, the relationship to the legacy `get_peers()` shim, AND cites the issue. Diagnostic strings are user-facing and contain the HTTP code + the actionable next step. The shim pattern (`get_peers()` discards the diagnostic) preserves the bare-list contract for the system-prompt formatters without breaking them.

**Architecture.** Clean layering: client returns `(peers, diag)`, tool surfaces. Slight crossover — the client embeds operator advice ("Restart the workspace usually re-mints it") in the diagnostic string, which arguably belongs at the tool layer. **Optional/FYI:** future refactor could push human-readable strings to `a2a_tools.py` and have the client return structured error info (e.g. `(peers, ErrorReason.AUTH_FAILED)` enum), so the client stays format-agnostic. For Phase 1 the current shape is fine.

**Security.** Diagnostic strings include HTTP status codes but NOT raw response bodies — prevents accidental leakage from a misbehaving platform. Network exception message includes `e` (could include URLs/IPs) but the runtime is sandboxed and this is workspace-internal output. Low risk.

**Performance.** Same single HTTP call, same 10s timeout. No regression.

**The `tool_list_peers` rewrite removes the misleading "may be isolated" string in every branch:**
- True empty → "You have no peers in the platform registry. (No parent, no children, no siblings registered.)"
- Auth/404/5xx/network → "No peers found. {diagnostic}"

`assert "may be isolated" not in result` appears in three test cases — locks the regression closed.

**Verdict: approve.** Approving via secondary account so this can land.

PR #2403, opened 2026-04-30
fix(ci): hard-fail when >50% of fleet unreachable post-redeploy

HongmingWang-Rabbit:
## Summary

Follow-up to #2402 addressing the **Optional/Consider** suggestion in the review on that PR — adds a sanity floor on top of the unreachable-soft-warn:

> if MORE than half the fleet is unreachable, this isn't a teardown race — it's a real outage (e.g. the new image crashes on startup). Hard-fail.

#2402 was merged before I got to this; opening as a separate small PR.

## Why

The unreachable-soft-warn from #2402 fixed the e2e-* teardown-race flake. But it left a residual gap: if the new image crashes on startup, every tenant ends up unreachable and the soft-warn alone would let that ship as a green deploy. Canary-verify catches it on the canary tenant first, but this guard is a fallback for canary-skip dispatches and same-batch races.

50% threshold is comfortably above the typical e2e-* teardown rate (5-10/hour, ~1 ephemeral tenant per batch) but below any plausible real-outage scenario.

## How

- Add a 50%-floor check after the soft-warn in both `redeploy-tenants-on-{staging,main}.yml`.
- Hard-fail with a clear `::error::` message that distinguishes "real outage" from "teardown race".

## Test plan

- [x] YAML validates (`python3 -c 'import yaml; yaml.safe_load(...)'`)
- [ ] Merge to staging
- [ ] Watch a real redeploy chain — expect identical behavior in the typical case (1 e2e-* unreachable, soft-warn pass)

### DIFF
--- a/.github/workflows/redeploy-tenants-on-main.yml
+++ b/.github/workflows/redeploy-tenants-on-main.yml
@@ -306,6 +306,16 @@ jobs:
           if [ $UNREACHABLE_COUNT -gt 0 ]; then
             echo "::warning::$UNREACHABLE_COUNT tenant(s) unreachable post-redeploy. Likely benign teardown race — CP healthz monitor catches real outages."
           fi
+
+          # Belt-and-suspenders sanity floor: same logic as the staging
+          # variant — if MORE than half the prod fleet is unreachable,
+          # this is a real outage, not a teardown race. Hard-fail.
+          TOTAL_VERIFIED=${#SLUGS[@]}
+          if [ $UNREACHABLE_COUNT -gt $((TOTAL_VERIFIED / 2)) ]; then
+            echo "::error::$UNREACHABLE_COUNT of $TOTAL_VERIFIED tenant(s) unreachable — exceeds 50% threshold. Likely real outage, not teardown race."
+            exit 1
+          fi
+
           if [ $STALE_COUNT -gt 0 ]; then
             echo "::error::$STALE_COUNT tenant(s) returned a stale SHA. ssm_status=Success was misleading — see job summary."
             exit 1
--- a/.github/workflows/redeploy-tenants-on-main.yml
+++ b/.github/workflows/redeploy-tenants-on-main.yml
@@ -308,11 +308,12 @@ jobs:
           fi
 
           # Belt-and-suspenders sanity floor: same logic as the staging
-          # variant — if MORE than half the prod fleet is unreachable,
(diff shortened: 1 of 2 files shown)


HongmingWang-Rabbit (commented):

**Correctness.** The 50%-floor reads as `UNREACHABLE_COUNT > TOTAL_VERIFIED / 2` (strict greater-than, integer division). Math:
- 2-tenant fleet: trips on 2/2 (`2 > 1`), passes on 1/2 (`1 > 1` false) ✓ teardown race
- 4-tenant fleet: trips on 3/4 (`3 > 2`), passes on 2/4 (`2 > 2` false) ✓
- 0-tenant fleet: `0 > 0` false → no false positive ✓
- 10-tenant prod fleet: trips on 6+, passes on 5 ✓

**One edge case worth flagging:** 1-tenant fleet with 1 unreachable. `TOTAL_VERIFIED=1`, threshold = `1/2=0`, `1 > 0` is true → hard-fail. This means a single-tenant fleet (or a redeploy batch reduced to 1 tenant after some have already been culled) where the lone tenant happens to be a torn-down `e2e-*` would now hard-fail instead of soft-warning. Almost certainly fine on prod (>1 tenant always), but on staging during quiet windows this could cause the same flake class #2402 just fixed. If you want to be extra safe, add `&& TOTAL_VERIFIED -ge 2` (or `-ge 4`) to the floor check — only enforce the threshold once the fleet is large enough that a teardown race becomes statistically implausible.

**Readability.** "Belt-and-suspenders sanity floor" plus "this isn't a teardown race — it's a real outage" inline comment makes the intent obvious. The `::error::` message distinguishes the two failure modes for on-call.

**Architecture.** Layered defense reads correctly:
1. Canary-verify is the primary signal (catches real outages on tenant 0 before fanout)
2. Per-tenant soft-warn from #2402 absorbs 1-of-N ephemeral teardown races
3. This 50%-floor catches the canary-skip + same-batch-race scenarios that the first two miss

The duplication between `redeploy-tenants-on-{main,staging}.yml` is fine at this size — extracting to a shared composite action would add more complexity than it saves for ~10 lines of logic.

**Security.** No credentials, no auth-bypass, no secrets handling. Pure threshold logic over already-collected results.

**Performance.** Trivial bash arithmetic; zero workflow runtime cost.

LGTM modulo the 1-tenant edge case observation. CodeQL-Analyze checks still in progress.

PR #2415, opened 2026-04-30
feat(workspace-runtime): inbox polling for standalone molecule-mcp

HongmingWang-Rabbit:
## Summary

Adds inbound A2A delivery to the universal MCP server so standalone-runtime agents (Claude Code, hermes, codex) can OBSERVE messages, not just send them.

Before: `molecule-mcp` was outbound-only — agents could `delegate_task`, `list_peers`, `commit_memory`, but never saw canvas-user or peer-agent messages addressed to them. The "constantly responding" gap from the canvas screenshots was a direct symptom.

After: a daemon thread polls `/workspaces/:id/activity?type=a2a_receive&since_id=<cursor>` every 5s, fills an in-memory queue, and exposes three new MCP tools:

- `wait_for_message(timeout_secs?)` — block until next message (capped 300s)
- `inbox_peek(limit?)` — list pending messages non-destructively
- `inbox_pop(activity_id)` — drop a handled message

Cursor persisted to `${CONFIGS_DIR}/.mcp_inbox_cursor`; on 410 (cursor pruned) we fall back to `since_secs=600` for bounded recovery.

## Scope discipline

In-container runtimes (push delivery via canvas WebSocket) skip activation — `inbox.get_state()` returns None and the tools surface `(inbox not enabled)` rather than double-deliver. Gate is `MOLECULE_MCP_DISABLE_INBOX` env, mirroring the existing `MOLECULE_MCP_DISABLE_HEARTBEAT` pattern.

For Claude Code specifically: the tools work but Claude doesn't auto-poll mid-turn — the channel plugin remains the right answer for true conversational delivery there. The inbox tools are the runtime-agnostic baseline that hermes/codex agents (with continuous loops) consume natively.

## Verification

Live tested against `hongmingwang.moleculesai.app` workspace `8dad3e29-c32a-4ec7-9ea7-94fe2d2d98ec`:

- ✅ `_poll_once` enqueued 5 a2a_receive rows from a 24h backlog, advanced cursor to newest
- ✅ Text extraction returned actual user-typed messages ("hi can you list your peers", "check again", "what do you mean Hermes Agent...")
- ✅ Caught a real bug: A2A SDK v1 uses `kind: "text"` not `type: "text"` for the part discriminator. My initial extractor (copied from molecule-mcp-claude-channel/server.ts:475) silently fell through to summary fallback. Now accepts both shapes — channel plugin should pick this up next sync.
- ✅ `tool_wait_for_message` returned head non-destructively, `tool_inbox_pop` removed it, queue went 5→4

## Test plan

- [x] `pytest tests/test_inbox.py` — 35 unit tests, all green (state transitions, cursor persist, 410 recovery, both `kind` + `type` envelope shapes, daemon thread)
- [x] `pytest tests/test_platform_tools.py tests/test_a2a_mcp_server.py tests/test_mcp_cli.py` — 79 tests still pass; snapshot regenerated for the 3 new tools
- [x] `scripts/wheel_smoke.py` updated to pin `inbox.activate / get_state / start_poller_thread` importability (matches the 0.1.16 main_sync regression-class precedent)
- [x] `TOP_LEVEL_MODULES` includes `inbox` so the rewriter package-prefixes imports cleanly
- [x] Live smoke against staging tenant (above)
- [ ] Watch CI through to staging merge + verify wheel auto-publishes (publish-runtime.yml triggers on workspace/** + scripts/build_runtime_package.py — both touched)

### DIFF
--- a/scripts/build_runtime_package.py
+++ b/scripts/build_runtime_package.py
@@ -64,6 +64,7 @@ TOP_LEVEL_MODULES = {
     "events",
     "executor_helpers",
     "heartbeat",
+    "inbox",
     "initial_prompt",
     "internal_chat_uploads",
     "internal_file_read",
--- a/scripts/wheel_smoke.py
+++ b/scripts/wheel_smoke.py
@@ -43,6 +43,21 @@ def smoke_imports_and_invariants() -> None:
     assert callable(cli_main), "a2a_mcp_server.cli_main must be callable"
     assert callable(mcp_cli_main), "mcp_cli.main must be callable"
 
+    # inbox.activate / get_state / start_poller_thread form the inbound
+    # delivery path for the standalone molecule-mcp wrapper. mcp_cli.main
+    # imports + activates these at startup; if a wheel ships without
+    # them, the standalone agent silently loses the wait_for_message /
+    # inbox_peek / inbox_pop tools and reverts to outbound-only.
+    from molecule_runtime.inbox import (  # noqa: F401
+        InboxState,
+        activate as inbox_activate,
+        get_state as inbox_get_state,
+        start_poller_thread as inbox_start_poller_thread,
+    )
+    assert callable(inbox_activate), "inbox.activate must be callable"
+    assert callable(inbox_get_state), "inbox.get_state must be callable"
+    assert callable(inbox_start_poller_thread), "inbox.start_poller_thread must be callable"
+
     assert a2a_client._A2A_ERROR_PREFIX, "a2a_client missing error sentinel"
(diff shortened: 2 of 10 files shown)


hongmingwang-moleculeai (approved):
Live-verified against staging tenant: poll → enqueue → wait_for_message → pop round-trip works against real workspace 8dad3e29. Caught and fixed an A2A-SDK-v0-vs-v1 envelope shape regression (kind vs type) that the channel plugin's extractText also has. Auto-merge engaged.

PR #2425, opened 2026-05-01
fix(mcp_cli): escalate heartbeat 401s with re-onboard guidance

HongmingWang-Rabbit:
## Summary

When a workspace is deleted server-side (DELETE /workspaces/:id), the platform revokes its auth tokens. The universal molecule-mcp wheel's heartbeat thread starts 401-ing — but previously just logged at WARNING per tick with no actionable signal. Operators had no clue their workspace was gone; their MCP tools just 401-ed silently.

This was hit live on the hongmingwang tenant 2026-04-30 — the channel bridge's stored .env pointed at a deleted workspace (\`a1771dba\`), heartbeat 401-ed forever, no escalation, no recovery instructions.

The register-time path at \`mcp_cli.py:104-111\` already does the right thing for 401 (sys.exit(3) with regenerate-from-Tokens-tab message). Extend the same pattern to the heartbeat — but escalate gracefully (not exit), since the MCP server is still running for in-flight tools.

### Behaviour

| Consecutive 401/403 | Log level |
|---|---|
| 1, 2 | WARNING ("auth failure 1/3 — token may be revoked") |
| 3 | **ERROR** with full re-onboard instructions, names the dead workspace_id |
| 23, 43, 63, ... (every 20 ticks ~7min) | re-log ERROR |

5xx, 429, and transport errors do NOT increment the auth counter — that would mislead the operator (a server blip would trigger "token revoked" when the token is fine).

### Tests

5 new cases:
1. Single 401 stays at WARNING (no premature alarm)
2. Three consecutive 401s escalate to ERROR with the right keywords (\`revoked\`/\`deleted\`, \`regenerate\`/\`tokens\`, names the workspace_id)
3. 403 Forbidden treated identically to 401
4. Recovery via 200 resets the counter — subsequent isolated 401 doesn't escalate
5. Three consecutive 500s do NOT trigger the auth path

## Test plan

- [x] \`pytest tests/test_mcp_cli.py\` — 31 pass (26 existing + 5 new)
- [x] Wider sweep: \`tests/test_mcp_cli.py tests/test_a2a_client.py tests/test_inbox.py\` — 112 pass
- [ ] CI green (auto)

### DIFF
--- a/workspace/mcp_cli.py
+++ b/workspace/mcp_cli.py
@@ -51,6 +51,16 @@ logger = logging.getLogger(__name__)
 # laptop sleep.
 HEARTBEAT_INTERVAL_SECONDS = 20.0
 
+# After this many consecutive 401/403 heartbeats, escalate from
+# WARNING to ERROR with re-onboard guidance. 3 ticks at 20s = ~1 minute
+# of sustained auth failure — enough to rule out a transient platform
+# blip but quick enough that an operator doesn't sit puzzled for 10
+# minutes wondering why their MCP tools 401. Same threshold used for
+# repeat-logging at 20-tick (~7 min) intervals so a long-running
+# session that missed the first ERROR still sees the message.
+_HEARTBEAT_AUTH_LOUD_THRESHOLD = 3
+_HEARTBEAT_AUTH_RELOG_INTERVAL = 20
+
 
 def _platform_register(platform_url: str, workspace_id: str, token: str) -> None:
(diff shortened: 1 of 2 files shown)


hongmingwang-moleculeai (approved):
Self-approve: small + focused. Five-axis pass — surfaces a real silent-fail UX gap caught live earlier today, narrow scope, tests cover the loud path + the refusal-to-escalate paths.

PR #2445, opened 2026-05-01
feat(terminal): add /workspaces/:id/terminal/diagnose endpoint

HongmingWang-Rabbit:
## Summary

Adds `GET /workspaces/:id/terminal/diagnose` — a non-interactive HTTP/JSON sibling of `/terminal` that runs the same EIC + ssh stages and reports per-stage status. Built specifically because the canvas terminal currently fails silently (\"Session ended\" with no error frame on hongmingwang's hermes workspace) and there's no remote-readable signal of which stage broke — the ssh client's stderr lives only in the workspace-server's stdout on the tenant CP EC2.

The endpoint splits the failure modes that all look identical from the user's side:
- **IAM broke** → `first_failure=\"send-ssh-public-key\"`, error contains `AccessDeniedException`
- **Tunnel/SG broke** → `first_failure=\"wait-for-port\"`, detail contains the `aws ec2-instance-connect open-tunnel` stderr
- **sshd auth broke** → `first_failure=\"ssh-probe\"`, detail contains `Permission denied`
- **Shell or network broke** → `first_failure=\"ssh-probe\"`, detail contains `Connection refused` or timeout

## Stages mirrored from `handleRemoteConnect`

1. `ssh-keygen` — ephemeral session keypair
2. `send-ssh-public-key` — AWS EIC API push (IAM-gated)
3. `pick-free-port` — local port for the tunnel
4. `open-tunnel` — `aws ec2-instance-connect open-tunnel` subprocess
5. `wait-for-port` — confirm tunnel actually binds (folds tunnel stderr into Detail when it doesn't)
6. `ssh-probe` — non-interactive \`ssh ... 'echo MOLECULE_TERMINAL_PROBE_OK'\` that proves auth + bash + round-trip via `CombinedOutput`

Local Docker workspaces get a smaller `container-found` + `container-running` probe, same response shape so callers don't need to branch.

## Auth

Rides on existing `WorkspaceAuth` middleware. Operators with the tenant `ADMIN_TOKEN` (fetched via `/cp/admin/orgs/:slug/admin-token`) can probe any workspace without per-workspace token — same admin path the canvas dashboard uses.

## Response shape

Always HTTP 200 (success or stage failure both in the JSON body — no need to branch on status):

\`\`\`json
{
  \"workspace_id\": \"...\",
  \"instance_id\": \"i-...\",
  \"remote\": true,
  \"ok\": false,
  \"first_failure\": \"ssh-probe\",
  \"steps\": [
    {\"name\": \"ssh-keygen\", \"ok\": true, \"duration_ms\": 12},
    {\"name\": \"send-ssh-public-key\", \"ok\": true, \"duration_ms\": 234},
    {\"name\": \"pick-free-port\", \"ok\": true, \"duration_ms\": 1, \"detail\": \"port=12345\"},
    {\"name\": \"open-tunnel\", \"ok\": true, \"duration_ms\": 456},
    {\"name\": \"wait-for-port\", \"ok\": true, \"duration_ms\": 789},
    {\"name\": \"ssh-probe\", \"ok\": false, \"duration_ms\": 11000,
     \"error\": \"exit status 255\",
     \"detail\": \"Permission denied (publickey).\"}
  ]
}
\`\`\`

## Test plan

- [x] \`go test ./internal/handlers/ -run \"TestHandleDiagnose|TestDiagnoseRemote\" -v -count=1\` — 3 new tests pass
- [x] \`go test ./internal/handlers/ -count=1\` — full suite passes (3.9s, no regressions)
- [x] \`go vet ./...\` clean
- [ ] After merge: deploy to hongmingwang tenant + curl \`/workspaces/{hermes-id}/terminal/diagnose\` to pin the actual hermes failure stage

## New tests

- \`TestHandleDiagnose_RoutesToRemote\` — instance_id present → remote path; \`first_failure=\"send-ssh-public-key\"\` when stubbed to return AWS error
- \`TestHandleDiagnose_RoutesToLocal\` — empty instance_id → local path; \`first_failure=\"docker-available\"\` with nil docker client
- \`TestDiagnoseRemote_StopsAtSSHProbe\` — full success through wait-for-port, probe stub returns \"Permission denied\"; confirms \`first_failure=\"ssh-probe\"\` and the probe step's Detail/Error carries ssh stderr verbatim (the whole point of the endpoint)

Tests stub \`sendSSHPublicKey\`, \`openTunnelCmd\`, \`sshProbeCmd\` via existing package-level vars (same pattern as \`TestSSHCommandCmd_*\`) so the suite stays hermetic — no AWS calls, no network.

## Related

- Resolves task #200
- Supports task #193 (root cause: workspace EC2 sshd unresponsive — without this endpoint we couldn't pin the failure stage from outside the tenant CP EC2)
- Complements PR #2434 (ConnectTimeout=10 symptom fix shipped 2026-04-30)

### DIFF
--- a/workspace-server/internal/handlers/terminal_diagnose.go
+++ b/workspace-server/internal/handlers/terminal_diagnose.go
@@ -1,18 +1,47 @@
 package handlers
 
 import (
+	"bytes"
 	"context"
 	"fmt"
 	"net/http"
 	"os"
 	"os/exec"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/Molecule-AI/molecule-monorepo/platform/internal/db"
 	"github.com/Molecule-AI/molecule-monorepo/platform/internal/provisioner"
+	"github.com/Molecule-AI/molecule-monorepo/platform/internal/wsauth"
 	"github.com/gin-gonic/gin"
 )
+
+// syncBuf is a goroutine-safe writer that wraps bytes.Buffer with a mutex.
+// Used to capture subprocess stderr without racing the os/exec stderr-copy
+// goroutine: ``cmd.Stderr = io.Writer`` spawns a background goroutine that
+// reads from the subprocess's stderr fd and calls Write on our writer, so
+// reading the buffer from another goroutine (e.g., on wait-for-port
+// timeout while the tunnel may still be writing) without synchronization
+// is a data race that ``go test -race`` would flag. ``strings.Builder``
+// and bare ``bytes.Buffer`` aren't goroutine-safe; this tiny shim is the
+// cheapest fix.
+type syncBuf struct {
+	mu sync.Mutex
+	b  bytes.Buffer
+}
+
+func (s *syncBuf) Write(p []byte) (int, error) {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	return s.b.Write(p)
+}
+
+func (s *syncBuf) String() string {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	return s.b.String()
+}
 
 // HandleDiagnose handles GET /workspaces/:id/terminal/diagnose. It runs the
 // same per-step pipeline as HandleConnect (ssh-keygen → EIC send-key → tunnel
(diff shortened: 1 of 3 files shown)


hongmingwang-moleculeai (commented):
Five-axis review (LGTM from a quality standpoint, flagged for human merge per substantive-logic policy):

- **Correctness** — Mirrors HandleConnect's EIC+ssh stages 1-for-1; `defer kill+wait` on the tunnel subprocess is right (no zombies); 30s outer timeout + 10s ssh ConnectTimeout + 10s waitForPort caps are coherent. Always-200 with failure in JSON body is the right contract — operator tools don't need branch-on-status. Local-path fallback (no instance_id) returns the same response shape — good for callers.
- **Readability** — Lead comment names *why* this exists ("canvas terminal fails silently — ssh stderr lives only in workspace-server stdout on tenant CP EC2"), and the stage→team routing hint ("send-ssh-public-key → IAM team; ssh-probe → SG/sshd") is operationally non-obvious and worth keeping in the source. Function-var stubs for `sshProbeCmd`/`openTunnelCmd`/`sendSSHPublicKey` follow the existing pattern.
- **Architecture** — Pure diagnostic, read-only, mounted on the same `wsAuth` middleware as `/terminal`. No new auth path. Stage list reuses `eicSSHOptions`, `sendSSHPublicKey`, `openTunnelCmd`, `pickFreePort`, `waitForPort` — single source of truth, drift-resistant.
- **Security** — Same auth gate as `/terminal`. SSH stderr surfaces to the caller — that's the *point* and the caller is already permitted to drive an interactive shell against the same workspace, so no new info disclosure. Env reads (`WORKSPACE_EC2_OS_USER`, `AWS_REGION`) are standard and not echoed in the response.
- **Performance** — Subprocess spawn cost matches existing `/terminal` open path, capped tighter. Each stage records DurationMs so latency triage is built in.

**Coverage check:** 3 dispatch tests — `RoutesToRemote` (instance_id non-empty → EIC path, fail at send-key), `RoutesToLocal` (empty instance_id → docker path, fail at docker-available), `StopsAtSSHProbe` (full happy path through send-key + tunnel + wait, then probe fails with stubbed "Permission denied"). Branches covered: remote-vs-local dispatch, early-failure surfacing, late-failure stderr surfacing — the operationally important paths. Stubbing `openTunnelCmd` with `nc -l` to make `waitForPort` real-but-hermetic is a clean trick.

**Loop policy note:** Substantive new endpoint (~551 LOC), so my session loop defers final approval to a human reviewer rather than auto-approving. CI green (21/21 SUCCESS), MERGEABLE, no .github/workflows changes, no security red flags. Ready for merge once a human signs off.

Issue #2275, opened 2026-04-29
[P1] Wheel smoke: invoke main() against stub config — catches a2a-sdk migration finds at publish time


HongmingWang-Rabbit:
## Background

Today we found 5 a2a-sdk v0→v1 migration regressions across template repos:

- **gemini-cli**: top-level import → boot-smoke caught it (3 cascade fires failed)
- **crewai / openclaw / autogen**: lazy import inside \`async def execute()\` → boot-smoke MISSED, images shipped, runtime ImportError on first A2A message
- **claude-code**: \`from a2a.types import FilePart\` (lazy) → boot-smoke MISSED, every workspace crashed on result delivery (production impact)

## Root cause

The current boot-smoke step in \`molecule-ci/.github/workflows/publish-template-image.yml\` does \`docker inspect\` and imports \`/app/*.py\` at module load. **Lazy imports inside functions are never evaluated** at smoke time — they crash at runtime instead.

## Fix

Extend the boot-smoke step to actually invoke a stub \`main()\` against minimal env so the runtime startup path executes. Specifically:

1. Add a stub-config workspace ID + minimal env (no real platform).
2. Run \`molecule-runtime --smoke-mode\` (new flag) which loads adapter, calls \`execute()\` with a no-op message, and exits.
3. Any ImportError or critical exception during this path = smoke failure.

Catches: every import path actually evaluated during a real workspace boot, including lazy imports.

## Why P1

Today's incident: 4 templates shipped broken images for 4+ hours before audit caught them. The cost of one image-rebuild gate (~30s extra in CI) eliminates a recurring class of production breakage. Direct empirical evidence: this single gate would have caught **all 5** of today's findings before the v0.1.36 publish.

## Implementation sketch

1. \`molecule_runtime\` exposes \`--smoke-mode\` flag — short-circuits to a fake EventQueue + stub RequestContext, calls \`execute()\` once, exits 0/1.
2. \`molecule-ci/.github/workflows/publish-template-image.yml\` adds a step:
   \`\`\`yaml
   - name: Invoke main() smoke (catches lazy imports)
     run: |
       docker run --rm \
         -e WORKSPACE_ID=fake -e MOLECULE_SMOKE_MODE=1 \
         "$IMAGE" molecule-runtime --smoke-mode
   \`\`\`
3. Failure = block publish (same gate as today's docker-inspect smoke).

## Linked

[...]

PR #2446, opened 2026-05-01
feat(wheel-smoke): exercise execute() to catch lazy imports (#2275)

HongmingWang-Rabbit:
## Summary

- New `workspace/smoke_mode.py`: `MOLECULE_SMOKE_MODE=1` short-circuit in `main.py` invokes `executor.execute(stub_ctx, stub_queue)` once with a 5s timeout, exit 1 on `ImportError`/`ModuleNotFoundError`, exit 0 on timeout / clean return / non-import errors.
- Closes the gap that issue #2275 names: existing `wheel_smoke.py` only IMPORTs `molecule_runtime.main` at module scope, missing lazy imports buried inside `async def execute(...)` bodies (e.g. `from a2a.types import FilePart`). The 2026-04-2x v0→v1 a2a-sdk migration shipped 5 such regressions that all looked fine at module-load smoke.
- Phase 2 (separate PR in molecule-ci) wires `MOLECULE_SMOKE_MODE=1` into `publish-template-image.yml` so the publish gate runs this against every template image before pushing.

## Why exit 0 on timeout?

If the import tree is healthy, `execute()` proceeds far enough to hit a network boundary (LLM call, etc.) and stalls — that's the expected outcome and a *pass*. If a lazy import is broken, the call raises before it gets that far. Other exception types (auth errors, validation) are caught by adapter-level tests, not this gate, so they also pass.

## What's stubbed vs real

- Stub `(RequestContext, EventQueue)` is built from the *real* a2a-sdk so SendMessageRequest/RequestContext constructor changes also surface as smoke failures — the regression class includes "SDK refactored mid-publish."
- The executor itself isn't stubbed; its full code path runs.
- conftest.py's `a2a.*` shims don't include `ServerCallContext` / `SendMessageRequest`, so two construction tests are gated on `_real_a2a_sdk_available()` and skip in CI's stubbed environment.

## Test plan

- [x] `pytest workspace/tests/test_smoke_mode.py` — 21 passed, 2 skipped (real-SDK-only construction tests, expected behavior with conftest stubs)
- [x] Rewriter tests still pass (17/17) after adding `smoke_mode` to `TOP_LEVEL_MODULES`
- [x] Wheel build succeeds; `smoke_mode.py` rewritten cleanly, importable as `from molecule_runtime.smoke_mode import is_smoke_mode, run_executor_smoke`
- [x] End-to-end validated against installed a2a-sdk: failing executor (FilePart import error) → exit 1; timeout executor → exit 0; clean executor → exit 0; auth-error executor → exit 0
- [ ] Phase 2 PR in `molecule-ci` wires this into `publish-template-image.yml`

### DIFF
--- a/scripts/build_runtime_package.py
+++ b/scripts/build_runtime_package.py
@@ -78,6 +78,7 @@ TOP_LEVEL_MODULES = {
     "prompt",
     "runtime_wedge",
     "shared_runtime",
+    "smoke_mode",
     "transcript_auth",
     "watcher",
 }
--- a/workspace/main.py
+++ b/workspace/main.py
@@ -136,6 +136,20 @@ async def main():  # pragma: no cover
         await adapter.setup(adapter_config)
         executor = await adapter.create_executor(adapter_config)
 
+        # 5a. Boot-smoke short-circuit (issue #2275): if MOLECULE_SMOKE_MODE
+        # is set, exercise the executor's full import tree by calling
+        # execute() once with stub deps + a short timeout. Skips platform
+        # registration + uvicorn entirely. Returns process exit code.
+        from smoke_mode import is_smoke_mode, run_executor_smoke
+        if is_smoke_mode():
+            exit_code = await run_executor_smoke(executor)
+            if hasattr(heartbeat, "stop"):
+                try:
+                    await heartbeat.stop()
+                except Exception:  # noqa: BLE001
+                    pass
+            raise SystemExit(exit_code)
+
         # 5b. Restore from pre-stop snapshot if one exists (GH#1391).
         # The snapshot is scrubbed before being written, so secrets are
         # already redacted — restore_state must not re-expose them.
--- a/workspace/smoke_mode.py
+++ b/workspace/smoke_mode.py
@@ -1,0 +1,140 @@
+"""Boot smoke mode — exercises the executor's full import tree without touching real platforms.
+
+Why this exists (issue #2275): the existing `wheel_smoke.py` only IMPORTS
+`molecule_runtime.main` at module scope. Lazy imports buried inside
+`async def execute(...)` bodies (e.g. `from a2a.types import FilePart`)
+NEVER evaluate at static-import time — they crash at first message
+delivery in production.
+
+The 2026-04-2x v0→v1 a2a-sdk migration shipped 5 such regressions in
+templates that all looked fine at module-load smoke. This module fills
+the gap by actually invoking `executor.execute(stub_ctx, stub_queue)`
+once with a short timeout. If the import-tree is healthy the call
+proceeds far enough to hit a network boundary (LLM call, etc.) and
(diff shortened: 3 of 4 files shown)


hongmingwang-moleculeai (commented):
Five-axis review (LGTM from a quality standpoint, deferring approval per loop policy until CI settles):

- **Correctness** — Exit-code semantics match the 4 regression classes described in the body: ImportError/ModuleNotFoundError → 1 (the gate's whole point); TimeoutError → 0 (imports past the gate, hit a network boundary — expected pass); other exceptions → 0 (downstream of imports, owned by adapter-level tests); stub-context build failure → 1 (broken a2a-sdk import path, same regression class). `heartbeat.stop()` cleanup before SystemExit is right. `asyncio.wait_for` cancellation is captured via CancelledError too — good.
- **Readability** — Docstrings name *why* the gate exists (2026-04-2x v0→v1 a2a-sdk migration shipped 5 lazy-import regressions; existing wheel_smoke.py only IMPORTs at module scope, missing `async def execute()` body imports like `from a2a.types import FilePart`). The PASS-on-timeout rationale is documented inline so future readers don't "helpfully" tighten the timeout. Test docstring also calls out the no-coverage-on-main reasoning + the conftest a2a shim caveat.
- **Architecture** — Function-scope a2a imports inside `_build_stub_context()` keep the wheel bootable without smoke mode active — wheel still imports cleanly even if the test-time a2a-sdk path is broken. `scripts/build_runtime_package.py` adds `smoke_mode` to the bundled module set so the gate ships with the runtime. Phase 2 (CI wiring) explicitly deferred to molecule-ci PR — clean split.
- **Security** — Env var read only (`MOLECULE_SMOKE_MODE`, `MOLECULE_SMOKE_TIMEOUT_SECS`). Stub message is hardcoded "smoke test". No platform registration, no uvicorn, no network egress (the timeout is *the boundary*). No new attack surface.
- **Performance** — Single 5s-timeout invocation per template image. Configurable via env if a slow adapter overlaps. Body explicitly cautions against making it too long since publish workflow multiplies across N templates — that operational note belongs in the source.

**Coverage:** `workspace/tests/test_smoke_mode.py` (190 LOC) exercises is_smoke_mode env truthy parsing, run_executor_smoke timeout-pass / import-error-fail / other-exception-pass / clean-completion-pass branches. End-to-end main.py path explicitly out-of-scope (covered by the publish-template-image.yml gate this exists for).

**Loop policy note:** 12/20 checks SUCCESS, 8 IN_PROGRESS (including the operationally relevant "PR-built wheel + import smoke" and "Python Lint & Test"). My session loop only auto-approves on all-green for non-dependabot authors, so deferring final approval to a human reviewer or the next tick once checks settle. Diff is +345/-0 additive; no security red flags; no .github/workflows changes; no deleted tests. This directly addresses the regression class the user memorized after the 2026-04-27 RuntimeCapabilities ImportError outage.

PR #2447, opened 2026-05-01
chore(smoke-mode): harden module-load + drop dead except clause

HongmingWang-Rabbit:
## Summary

Two follow-ups from the #2275 Phase 1 self-review (PR #2446):

1. **`_SMOKE_TIMEOUT_SECS` evaluated at module load** — `main.py` imports `smoke_mode` unconditionally (before the `is_smoke_mode()` check), so a malformed `MOLECULE_SMOKE_TIMEOUT_SECS` env value would SystemExit every workspace boot, not just smoke runs. Wrapped in try/except with a 5.0 fallback. Probability of a typo'd env value hitting production is low (CI-only knob), but the footgun is removed.
2. **Unreachable `AttributeError` in `_real_a2a_sdk_available()`** — `from X import Y` raises `ImportError` when `Y` is missing on `X`, never `AttributeError`. Dropped the unreachable branch.

## Test plan

- [x] New regression test (`test_smoke_timeout_falls_back_when_env_value_is_malformed`) reloads `smoke_mode` under a malformed env value and asserts the fallback. 22 pass, 2 skipped (real-SDK-only construction tests, expected).
- [x] No behavior change for the happy path.

### DIFF
--- a/workspace/smoke_mode.py
+++ b/workspace/smoke_mode.py
@@ -36,7 +36,13 @@ from typing import Any
 logger = logging.getLogger(__name__)
 
 
-_SMOKE_TIMEOUT_SECS = float(os.environ.get("MOLECULE_SMOKE_TIMEOUT_SECS", "5.0"))
+# Don't crash production boot if MOLECULE_SMOKE_TIMEOUT_SECS is malformed —
+# main.py imports smoke_mode unconditionally (before the is_smoke_mode()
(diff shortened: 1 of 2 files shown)


hongmingwang-moleculeai (approved):
LGTM — module-load harden + dead except cleanup, with regression test. Both findings from PR #2446 self-review.

PR #2452, opened 2026-05-01
fix(workspace-server): null removed_at when timestamp fetch fails (#2429 review polish)

HongmingWang-Rabbit:
Review finding from #2429 self-review (post-merge of #2449).

## Problem

The 410-Gone path in [\`workspace.go\`](https://github.com/Molecule-AI/molecule-core/blob/staging/workspace-server/internal/handlers/workspace.go#L661) issues a follow-up \`SELECT updated_at\` after detecting \`status='removed'\`. If that query fails (workspace row deleted between the two queries, transient DB error, etc.), \`removedAt\` stays as Go's zero time and the JSON body emits:

\`\`\`json
{
  \"error\": \"workspace removed\",
  \"id\": \"...\",
  \"removed_at\": \"0001-01-01T00:00:00Z\",
  \"hint\": \"...\"
}
\`\`\`

The client has to know to ignore that misleading sentinel. Better to emit \`null\` so callers can distinguish \"timestamp not available\" from \"removed at year zero\".

## Fix

Branch on \`removedAt.IsZero()\` and emit \`null\` for the failed-fetch path. Actionable signal (410 + hint) unchanged.

## Tests

- New \`TestWorkspaceGet_RemovedReturns410WithNullRemovedAtOnTimestampFetchFailure\` simulates the row vanishing via \`sqlmock\`'s \`WillReturnError(sql.ErrNoRows)\`.
- Existing \`_RemovedReturns410\` tightened to assert the happy-path timestamp is a non-null value (was previously just checking the key existed).

### DIFF
--- a/workspace-server/internal/handlers/workspace.go
+++ b/workspace-server/internal/handlers/workspace.go
@@ -660,16 +660,28 @@ func (h *WorkspaceHandler) Get(c *gin.Context) {
 	// to content negotiation.
 	if status, _ := ws["status"].(string); status == string(models.StatusRemoved) {
 		if c.Query("include_removed") != "true" {
+			// Best-effort fetch of the removal timestamp. If the row was
+			// deleted (or some transient DB error fired) between the
+			// scanWorkspaceRow above and this follow-up SELECT,
+			// removedAt stays as Go's zero time. Emit `null` in that
+			// case rather than the misleading `0001-01-01T00:00:00Z`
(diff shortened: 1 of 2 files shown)


hongmingwang-moleculeai (approved):
Self-review polish from #2449 retro. Test pins the regression class with sqlmock; happy-path test tightened from key-exists to value-non-null.

PR #2454, opened 2026-05-01
feat(canvas+workspace-server): data-driven Provider dropdown (#199)

HongmingWang-Rabbit:
## Summary

Option B PR-5 — final piece of the explicit-provider feature. Canvas Config tab now ships a Provider override input that's **adapter-driven**: the dropdown's suggestion list comes from each runtime's own template, not a hardcoded canvas-side enum.

Companion PR: [molecule-ai-workspace-template-hermes#31](https://github.com/Molecule-AI/molecule-ai-workspace-template-hermes/pull/31) — declares the providers list in hermes's \`config.yaml\`.

## Architecture

\`\`\`
template/config.yaml (runtime_config.providers: [...])
  └→ /templates JSON (templateSummary.Providers)
       └→ canvas RuntimeOption.providers
            └→ <datalist> options
\`\`\`

A new runtime can ship its own provider taxonomy without **any** canvas-side change. This is the \"Native + pluggable runtime\" invariant.

## What's in this PR

**workspace-server (`internal/handlers/templates.go`)**
- Add `Providers []string` to `templateSummary` JSON output
- Parse `runtime_config.providers` from each template's `config.yaml`
- 2 new tests (`TestTemplatesList_SurfacesProviders`, `TestTemplatesList_OmitsProvidersWhenAbsent`)

**canvas (`ConfigTab.tsx`)**
- Add `provider`/`originalProvider` state + dirty-tracking
- Wire `RuntimeOption.providers` from `/templates` response
- `<input list=…>` combobox surface — `<datalist>` populated from runtime data
- `deriveProvidersFromModels()` fallback for templates not yet migrated to explicit `providers:`
- Save flow: `PUT /workspaces/:id/provider` when dirty; suppress auto-restart from model handler to avoid double-restart
- Remove hardcoded `PROVIDER_SUGGESTIONS` constant
- 8 new tests (`ConfigTab.provider.test.tsx`) covering load/save/dirty/datalist-from-runtime/datalist-from-fallback paths

## Tests

- [x] \`go test ./internal/handlers/... -run Templates\` — 9 pass (2 new)
- [x] \`npx vitest run src/components/tabs/__tests__/\` — 15 pass (8 new)
- [x] \`npx tsc --noEmit\` clean
- [ ] E2E (after merge): hermes workspace on staging — Provider dropdown surfaces 21 hermes providers from \`/templates\`

## Notes

[...]

PR #2457, opened 2026-05-01
feat(canvas): dynamic provider dropdown in CreateWorkspaceDialog

HongmingWang-Rabbit:
## Summary

Make the provider dropdown in **CreateWorkspaceDialog** dynamic — same data-driven pattern PR #2454 set in ConfigTab.

The current dropdown lists all 16 entries of the hardcoded \`HERMES_PROVIDERS\` constant regardless of which template is selected. After PR #2454, ConfigTab reads \`runtime_config.providers\` from \`/templates\` and filters; the modal should match.

## Behavior

| Template state | Dropdown contents |
|---|---|
| Template declares \`runtime_config.providers: [\"anthropic\",\"minimax\"]\` | Only those 2 |
| Template has no \`providers\` field (older template) | Full \`HERMES_PROVIDERS\` (back-compat) |
| Template declares an unknown provider id | Full \`HERMES_PROVIDERS\` (defensive — never lock the operator out) |
| Currently-selected provider falls out of new list | Snaps to first available |

## Why

User feedback: \`selecting provider in the template modal should be dropdown and dynamic as well\` — three sources of truth for the provider list (canvas constant + template config.yaml + ConfigTab fetch) was always going to drift. This collapses the modal onto the same source ConfigTab uses.

## Test plan

- [x] 3 new tests pin filter, missing-field fallback, unknown-providers fallback
- [x] All 27 CreateWorkspaceDialog tests pass
- [x] All 488 canvas tests pass
- [x] tsc clean
- [ ] Once shipped: open New Workspace → type \`hermes\` → dropdown shows only providers the hermes template declares

### DIFF
--- a/canvas/src/components/CreateWorkspaceDialog.tsx
+++ b/canvas/src/components/CreateWorkspaceDialog.tsx
@@ -12,6 +12,19 @@ interface WorkspaceOption {
   tier: number;
 }
 
+// Subset of the /templates row used here. Mirrors the shape ConfigTab
+// reads. `providers` is the per-template declarative list of supported
+// LLM providers — sourced from the template's
+// runtime_config.providers (config.yaml). When present, it filters
+// the modal's provider <select> so an operator can only pick a
+// provider the template actually supports.
+interface TemplateSpec {
+  id: string;
+  name?: string;
+  runtime?: string;
+  providers?: string[];
+}
+
 interface HermesProvider {
   id: string;
   label: string;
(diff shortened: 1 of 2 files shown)


HongmingWang-Rabbit (commented):
**Five-axis review — comment-only (substantive UI logic + pending Canvas tabs E2E; companion to #2454).**

✅ **Correctness**
- `selectedTemplateSpec` matches by id/name/runtime case-insensitively — flexible against whatever the user types.
- `availableProviders` filter has the right defensive fallback: empty declared list → full catalog (older templates), AND filtered-list-zero (declared providers we have no metadata for) → full catalog. Prevents the lockout case where the operator can't submit because the `<select>` is empty.
- `useEffect` snaps `hermesProvider` to the first available when the filter shrinks the list — without this, the API key envVar lookup downstream would dereference a stale id.
- `/templates` fetch failure is swallowed → state stays empty → `availableProviders` falls back to HERMES_PROVIDERS. Self-hosted setups without /templates keep working.

✅ **Readability**
- Comments lead with the why: "same data source ConfigTab uses (PR #2454)", "operator can't be locked out", the static-catalog vs declared-list relationship.
- Three tests, each pins one failure mode with a docstring (filtered subset, missing-providers fallback, unknown-providers fallback).

✅ **Architecture**
- Reads from `/templates` — same source as ConfigTab's Provider override. Native + pluggable runtime invariant: each runtime owns its provider taxonomy in its own config.yaml; canvas surfaces it without canvas-side enums.
- Static `HERMES_PROVIDERS` catalog stays as the metadata source (label, envVar, defaultModel) — declared list filters it. Clean separation.

✅ **Security**
- No leaked credentials, no SQL surface, no auth bypass, no `.github/workflows/*` touched, no test deletions (adds 85 LOC of vitest cases).

✅ **Performance**
- `useMemo` on filter; `useEffect` only writes when the selected provider falls out of the available set (idempotent).
- One additional `/templates` fetch on dialog open — negligible.

⏳ **Status**: BLOCKED on review + Canvas tabs E2E pending. Once both clear, this is mergeable. Loop criteria flag substantive UI logic for human approval — comment-only here.

hongmingwang-moleculeai (approved):
LGTM — mirrors the ConfigTab pattern from PR #2454, three good fallback layers (declared / missing-field / unknown-id), tests pin all three.

PR #2461, opened 2026-05-01
feat(mcp): declare experimental.claude/channel capability for push UX

### LINKED ISSUE #2444: Follow-up: promote 'PR-built wheel + import smoke' to required check + live push-UX verification
Two loose ends from the 2026-04-30 push-UX session that need an explicit human action.

## 1. Promote \`PR-built wheel + import smoke\` to a required check on staging

PR #2439 refactored \`runtime-prbuild-compat.yml\` so the check always runs (with per-step \`if:\` gating on paths) — it's now eligible to be a required check without deadlocking unrelated PRs. PR #2433 is the cautionary precedent: a broken wheel landed on staging because this gate ran but didn't block.

Once the workflow has run cleanly on staging for a few PRs after #2439's merge, run:

\`\`\`bash
gh api -X PUT repos/Molecule-AI/molecule-core/branches/staging/protection/required_status_checks/contexts \\
  -f 'contexts[]=Analyze (go)' \\
  -f 'contexts[]=Analyze (javascript-typescript)' \\
  -f 'contexts[]=Analyze (python)' \\
  -f 'contexts[]=Detect changes' \\
  -f 'contexts[]=Shellcheck (E2E scripts)' \\
  -f 'contexts[]=Platform (Go)' \\
  -f 'contexts[]=Canvas (Next.js)' \\
  -f 'contexts[]=Python Lint & Test' \\
  -f 'contexts[]=PR-built wheel + import smoke'
\`\`\`

Same promotion likely also wanted on \`main\` if branch protection differs.

**Acceptance:** A PR opened with a deliberately-broken \`workspace/\` change (e.g. a syntax error in \`a2a_mcp_server.py\`) cannot reach staging — the check blocks it.

## 2. Live verify push-UX notifications on the published wheel

Wheel \`molecule-ai-workspace-runtime==0.1.57\` shipped with the \`notifications/claude/channel\` push seam (PRs #2433, #2436). Unit tests pin the wire shape but only a live run proves the bridge actually fires inside a notification-capable host.

**Repro steps:**

1. Fully restart Claude Code (not \`/mcp\` — that re-uses the cached session). The running session is bound to whatever wheel was installed at session start.
2. \`pip3 show molecule-ai-workspace-runtime\` → confirm 0.1.57 (or later)
3. From inside Claude Code:
   - \`list_peers()\` → pick a peer
   - From a separate workspace, send an A2A message TO this workspace
4. **Expected:** Claude Code interrupts the agent turn with the inbound message text, no \`wait_for_message\` call needed.
5. **Failure modes to catch:**
   - Agent has to call \`wait_for_message\` to see the message — push didn't fire (callback wiring bug)
   - Notification arrives but Claude Code doesn't surface it — host doesn't recognize the method (then it's a Claude Code change, not ours)
   - Wheel never receives the inbound — heartbeat / inbox poller broken (separate from push-UX)

**Acceptance:** A peer A2A message lands as an interruption in the live Claude Code session inside ~5 seconds of being sent, without the agent calling \`wait_for_message\`.

---

Both items are independent. Either can land first.

HongmingWang-Rabbit:
## Summary

Closes the live-verification gap in #2444 §2.

PR #2433 added the `notifications/claude/channel` push seam, but the initialize handshake never declared the matching `experimental.claude/channel` capability. Result: every notification fires correctly over the wire, then gets silently dropped by Claude Code's MCP client because the host doesn't know we want this method routed as a conversation interrupt.

Confirmed live in a fresh Claude Code session today:
- canvas chat "hi" arrived at the workspace runtime
- inbox poller queued it (visible via `inbox_peek`)
- no `<channel>` tag ever surfaced in the agent turn
- agent only saw the message after a manual `inbox_peek` call

The bun channel bridge at `molecule-mcp-claude-channel/server.ts:374` already declares this capability; we mirror that contract.

## Change

- `a2a_mcp_server.py`: extract `_build_initialize_result()` pure helper; declare `experimental.claude/channel` in the capabilities dict; replace the inline initialize handshake literal with the helper call. No behavioral change beyond the new capability flag.
- `tests/test_a2a_mcp_server.py`: 3 new pins — capability presence, `tools` capability still present, protocol version stable. The capability pin's failure message names the symptom so a future refactor can't strip the flag without a loud diff.

## Why a separate helper

The initialize result is the contract that flips Claude Code's notification routing on. Inlining it as a literal in `main()` made it testable only via integration tests, which is the regression class that hid this gap in the first place. The pure function is a 12-line cost; the catch is "silent push-UX regression on every refactor of `main()`."

## Verification

- ✅ `WORKSPACE_ID=test python3 -m pytest tests/test_a2a_mcp_server.py` — 21 passed
- ✅ `python3 scripts/build_runtime_package.py` — wheel build smoke clean
- ⏳ Live verification: requires the published wheel to land + a Claude Code restart in this session. Will report back once 0.1.66 ships.

## Test plan

- [x] Unit tests pass locally (21/21)
- [x] Wheel build smoke passes locally
- [ ] CI green (CodeQL, Python Lint, PR-built wheel + import smoke)
- [ ] Live: after the new wheel ships, canvas chat lands as `<channel>` tag without `inbox_peek` — closes #2444 §2

### DIFF
--- a/workspace/a2a_mcp_server.py
+++ b/workspace/a2a_mcp_server.py
@@ -149,6 +149,30 @@ async def handle_tool_call(name: str, arguments: dict) -> str:
 _CHANNEL_NOTIFICATION_METHOD = "notifications/claude/channel"
 
 
+def _build_initialize_result() -> dict:
+    """MCP initialize handshake result.
+
+    The ``experimental.claude/channel`` capability declaration is what
+    tells Claude Code's MCP client to route our
+    ``notifications/claude/channel`` emissions as conversation
+    interrupts (push UX). Without it the notification arrives over the
+    wire but is silently dropped instead of becoming a ``<channel>``
+    tag in the next agent turn — matching the
+    "Notification arrives but Claude Code doesn't surface it" failure
+    mode anticipated in molecule-core#2444. Mirrors the contract
+    declared by the molecule-mcp-claude-channel bun bridge
+    (server.ts:374).
+    """
+    return {
+        "protocolVersion": "2024-11-05",
+        "capabilities": {
+            "tools": {"listChanged": False},
+            "experimental": {"claude/channel": {}},
+        },
+        "serverInfo": {"name": "a2a-delegation", "version": "1.0.0"},
+    }
+
+
 def _build_channel_notification(msg: dict) -> dict:
     """Transform an ``InboxMessage.to_dict()`` into the MCP notification
     envelope expected by Claude Code's channel-bridge contract.
@@ -246,11 +270,7 @@ async def main():  # pragma: no cover
                     await write_response({
                         "jsonrpc": "2.0",
                         "id": req_id,
-                        "result": {
-                            "protocolVersion": "2024-11-05",
-                            "capabilities": {"tools": {"listChanged": False}},
-                            "serverInfo": {"name": "a2a-delegation", "version": "1.0.0"},
-                        },
+                        "result": _build_initialize_result(),
                     })
 
                 elif method == "notifications/initialized":
--- a/workspace/tests/test_a2a_mcp_server.py
+++ b/workspace/tests/test_a2a_mcp_server.py
@@ -237,3 +237,44 @@ def test_build_channel_notification_handles_missing_fields_gracefully():
     assert meta["activity_id"] == ""
     assert meta["peer_id"] == ""
     assert meta["kind"] == ""
+
+
+# ============== initialize handshake — capability declaration ==============
+# Without `experimental.claude/channel`, Claude Code's MCP client drops
+# our notifications/claude/channel emissions instead of routing them as
+# inline conversation interrupts. Anticipated as a failure mode in
+# molecule-core#2444 ("notification arrives but Claude Code doesn't
+# surface it"). Pin the declaration here so a refactor of
+# _build_initialize_result can't silently strip the flag.
+
+
+def test_initialize_declares_experimental_claude_channel_capability():
+    """Without this capability the push-UX bridge ships, the
+    notifications fire, and nothing happens in the host — silent. This
+    is the contract that flips Claude Code's routing on."""
+    from a2a_mcp_server import _build_initialize_result
+
+    result = _build_initialize_result()
+    experimental = result["capabilities"].get("experimental", {})
+
+    assert "claude/channel" in experimental, (
+        "experimental.claude/channel capability is required for Claude "
+        "Code to surface our notifications/claude/channel emissions as "
+        "conversation interrupts (issue #2444 §2). Removing this would "
+        "regress live push UX while leaving every unit test green."
+    )
+
+
+def test_initialize_keeps_tools_capability():
+    """Pin the tools capability too — losing it would break tools/list."""
+    from a2a_mcp_server import _build_initialize_result
+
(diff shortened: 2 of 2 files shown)


HongmingWang-Rabbit:
**Honesty addendum on this PR's verification claim.**

The PR body says this "closes #2444 §2." That's overstated. It closes the *capability-declaration candidate* — one of the failure modes #2444 explicitly enumerated:

> Notification arrives but Claude Code doesn't surface it — host doesn't recognize the method (then it's a Claude Code change, not ours)

The fix is based on observed pattern (telegram MCP declares `experimental.claude/channel` and works; molecule didn't and doesn't surface) but **not on a confirmed read of Claude Code's MCP client routing logic**. If Claude Code routes by server-name allowlist, by method-name without capability gating, or by some flag we haven't named yet, this PR adds a useful pin and a regression test but ships zero observable behavior change.

Live verification still required to actually close #2444 §2:

1. Wait for 0.1.66 (or whatever the next published wheel number is)
2. `pip3 install --upgrade molecule-ai-workspace-runtime`
3. **Fully restart Claude Code** (per #2444 step 1 — `/mcp` reuses the cached session)
4. Send a canvas chat message to the workspace
5. Expected: `<channel>` interrupt arrives in the agent turn within ~5s, no `inbox_peek` call needed

If step 5 fires, #2444 §2 is genuinely closed.
If it doesn't, the gap is in Claude Code's routing layer, not the MCP server — and we should refile against the host repo with a reproducer.

Will report back live verification result after the wheel ships.

HongmingWang-Rabbit:
**E2E verification result on the published wheel (0.1.66).**

Just installed 0.1.66 from PyPI and ran a real stdio JSON-RPC handshake against the binary:

\`\`\`
$ pip3 install --force-reinstall --no-deps "molecule-ai-workspace-runtime==0.1.66"
Successfully installed molecule-ai-workspace-runtime-0.1.66

$ WORKSPACE_ID=… PLATFORM_URL=… MOLECULE_WORKSPACE_TOKEN=fake MOLECULE_MCP_DISABLE_INBOX=1 \
    bash -c '(echo "{\"jsonrpc\":\"2.0\",\"id\":1,\"method\":\"initialize\",\"params\":{\"protocolVersion\":\"2024-11-05\",\"capabilities\":{},\"clientInfo\":{\"name\":\"e2e\",\"version\":\"0\"}}}"; sleep 2) | molecule-mcp'

{"jsonrpc": "2.0", "id": 1, "result": {"protocolVersion": "2024-11-05", "capabilities": {"tools": {"listChanged": false}, "experimental": {"claude/channel": {}}}, "serverInfo": {"name": "a2a-delegation", "version": "1.0.0"}}}
\`\`\`

The \`experimental.claude/channel\` capability is on the wire. Confirmed end-to-end against the actually-deployed wheel, not just the unit test of the helper.

This closes the molecule-mcp-side verification — wheel deploys correctly, handshake is right. Whether Claude Code's MCP client actually surfaces \`<channel>\` interrupts from this declaration is host-side behavior I can't measure from this end (and per follow-up research, also requires the \`instructions\` field which lands in #2463 → 0.1.67).

#2444 §2 closure still pending: live test in a Claude Code session running 0.1.67 with \`--dangerously-load-development-channels\`.

PR #2462, opened 2026-05-01
docs(mcp): correct server.ts reference + flag verification gap

HongmingWang-Rabbit:
## Summary

Follow-up to PR #2461 (commit `0a87dec5`), which was merged before live verification. Two doc-only corrections to the `_build_initialize_result()` docstring in `workspace/a2a_mcp_server.py`:

- **Wrong line reference.** The original claimed "mirrors molecule-mcp-claude-channel server.ts:374". Line 374 is unrelated poll-init code. The actual capability declaration site is server.ts:475 — and crucially, the bun bridge there declares only `{ capabilities: { tools: {} } }`. No `experimental.claude/channel`. The bridge is reported to deliver `notifications/claude/channel` successfully in Claude Code anyway, which directly contradicts the "capability is required for push UX" hypothesis.
- **MCP SDK doesn't gate custom notification methods.** `@modelcontextprotocol/sdk`'s `assertNotificationCapability` has no case for `notifications/claude/channel`, so server-side, declaring the capability is almost certainly a no-op.

The capability declaration stays in place (additive, the new tests pin its presence) but the docstring is downgraded from "this is what fixes push UX" to "hypothesized, unverified, and the more likely root causes are X / Y / Z."

## Why now

The next person to debug "Claude Code receives our notification but doesn't surface it" needs to know the experimental capability is *probably not the answer*, so they pursue:
- `writer.drain()` swallowing exceptions on a closed pipe
- inbox-thread → `asyncio.run_coroutine_threadsafe` race during init
- MCP transport not yet attached when the first inbox event fires

Live verification (#2444 §2 — fresh Claude Code session on this wheel + peer A2A message, observe interrupt) remains the open hard-gate.

## Test plan

- [x] Docstring-only change; no behavior touched
- [x] Existing tests still pass (`workspace/tests/test_a2a_mcp_server.py` — 21/21)
- [ ] Required check: runtime-pin-compat green
- [ ] Required check: ci.yml green
- [ ] Required check: codeql green

### DIFF
--- a/workspace/a2a_mcp_server.py
+++ b/workspace/a2a_mcp_server.py
@@ -152,16 +152,28 @@ _CHANNEL_NOTIFICATION_METHOD = "notifications/claude/channel"
 def _build_initialize_result() -> dict:
     """MCP initialize handshake result.
 
-    The ``experimental.claude/channel`` capability declaration is what
-    tells Claude Code's MCP client to route our
-    ``notifications/claude/channel`` emissions as conversation
-    interrupts (push UX). Without it the notification arrives over the
-    wire but is silently dropped instead of becoming a ``<channel>``
-    tag in the next agent turn — matching the
-    "Notification arrives but Claude Code doesn't surface it" failure
-    mode anticipated in molecule-core#2444. Mirrors the contract
-    declared by the molecule-mcp-claude-channel bun bridge
-    (server.ts:374).
+    Declares ``experimental.claude/channel`` as a *hypothesized*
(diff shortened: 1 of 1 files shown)


hongmingwang-moleculeai (approved):
Doc-only correction. Verified: bun bridge declares only `{ tools: {} }` at server.ts:475 and MCP SDK doesn't gate `notifications/claude/channel` on capability. Original commit's claim was inaccurate; this downgrade is appropriate.

Issue #2469, opened 2026-05-02
RFC: rename agent self-notify activity_type to eliminate inbox echo loop


HongmingWang-Rabbit:
## Summary

The workspace-server's `notify` handler (handlers/activity.go:475-480) writes the agent's outbound `send_message_to_user` POSTs to `activity_logs` with `activity_type='a2a_receive'` so the canvas chat-history loader can restore them after page reload. The molecule-mcp inbox poller (workspace/inbox.py:_poll_once) queries the same activity API filtered to `type=a2a_receive`, which means the agent's own outbound surfaces in its own inbox and is re-pushed via the channel notification path. Live-confirmed 2026-05-01: agent observed its own \"Hi! What can I help you with today?\" reply arriving as `← molecule: Agent message: ...` and explicitly recognized it as an echo (\"That's an echo of my own outgoing message — no new action needed\").

The current shape papers over the schema mismatch with two implicit discriminators:
- `method='notify'` (vs the real `message/send` / `tasks/send` for inbound)
- `source_id=NULL` (no peer = self)

## Why the proper fix lives at the data model

Calling an agent's own outbound `a2a_receive` is semantically wrong — it claims the workspace *received* something it *sent*. Every consumer of `activity_type` (chat-history loader, audit dashboards, the inbox poller, future analytics) has to remember the implicit `method='notify'` filter to interpret the row correctly. That's the kind of gap that grows new bugs over time.

Long-term proper: introduce `activity_type='agent_outbound'` so the row tells the truth about what happened. The inbox poller filters on a single type instead of `(type=a2a_receive AND method!='notify')`, the chat-history loader joins both types, audits get accurate counts, and future consumers don't have to learn the legacy quirk.

## Proposed change

**Migration** (workspace-server/migrations):
- Add new value to the activity_type enum check: `'agent_outbound'`
- Forward-only — no backfill needed; the existing `a2a_receive` rows for self-notify stay (chat-loader already handles them, no consumer pain)

**Handler** (handlers/activity.go):
- `Notify` handler emits `activity_type='agent_outbound'` instead of `a2a_receive` for the persistence row
- Summary stays `\"Agent message: \"+preview` (audit operators read this)
- response_body shape unchanged

**Chat-history loader** (handlers/activity.go: query that backs the canvas chat reload):
[...]

PR #2470, opened 2026-05-02
fix(inbox): skip self-notify rows to break echo loop

HongmingWang-Rabbit:
## Summary

The workspace-server's `/notify` handler writes the agent's own `send_message_to_user` POSTs to `activity_logs` with `activity_type='a2a_receive'`, `method='notify'`, `source_id=NULL` so the canvas chat-history loader can restore those bubbles after page reload (`workspace-server/internal/handlers/activity.go:425-480`). The activity API serves the row to `/workspaces/:id/activity?type=a2a_receive`, so the inbox poller picks it up and pushes the agent's own outbound back as an inbound `← molecule: Agent message: ...`.

Confirmed live 2026-05-01: agent observed its own `Hi! What can I help you with today?` reply arriving as a channel push and explicitly recognized it as an echo (\"That's an echo of my own outgoing message — no new action needed\").

## Fix

Add `_is_self_notify_row` predicate matched on `(method='notify' AND no source_id)` and call it from `_poll_once` before enqueue. The predicate combines BOTH discriminators so a future caller using `method='notify'` with a real `peer_id` still passes through. Cursor advances past skipped rows so we don't re-poll the same self-notify on every iteration.

This is **belt-and-braces**. The long-term proper fix lives upstream at the workspace-server: rename the misclassified `activity_type` from `'a2a_receive'` to `'agent_outbound'` so the row tells the truth about what happened — RFC at #2469. This guard stays regardless because it only excludes rows we never want anyway, so removing it after #2469 lands would just be churn.

## Test plan

- [x] Predicate true/false matrix:
  - method=notify + source_id=None → True
  - method=notify + source_id=\"\" → True
  - method=notify + source_id absent → True
  - method=message/send + source_id=None (real canvas) → False
  - method=tasks/send + source_id=ws-uuid (real peer) → False
  - method=notify + source_id=ws-peer-uuid (future repurpose) → False
- [x] Integrated `_poll_once` behavior:
  - Self-notify row in payload: not enqueued, count=0
  - Self-notify row: cursor still advances past it (no re-poll loop)
  - Self-notify row: notification callback NOT fired (no push to claude-channel)
- [x] Mutation-verified: reverting `inbox.py` to prior shape fails 7/7; applied state passes 48/48
- [ ] Live verify on staging workspace once published: `send_message_to_user` outbound no longer surfaces as inbound, agent does not see `Agent message: ...` echo

### DIFF
--- a/workspace/inbox.py
+++ b/workspace/inbox.py
@@ -364,6 +364,23 @@ def _extract_text(request_body: Any, summary: str | None) -> str:
     return summary or "(empty A2A message)"
 
 
+def _is_self_notify_row(row: dict[str, Any]) -> bool:
+    """Return True if ``row`` is the agent's own send_message_to_user
+    POST surfacing back through the activity API.
+
+    The shape (workspace-server handlers/activity.go, ``Notify`` writer):
+        method='notify' AND no peer (source_id is None or '')
+
+    Matched on both fields together so a future caller using
+    ``method='notify'`` for a different purpose with a real peer_id
+    still passes through.
+    """
+    if row.get("method") != "notify":
+        return False
+    source_id = row.get("source_id")
+    return source_id is None or source_id == ""
+
+
 def message_from_activity(row: dict[str, Any]) -> InboxMessage:
     """Convert one /activity row into an InboxMessage."""
     request_body = row.get("request_body")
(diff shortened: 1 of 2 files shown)


hongmingwang-moleculeai (approved):
Echo-loop guard. Predicate is well-scoped (matches BOTH method='notify' AND no source_id, so future method='notify'-with-peer doesn't accidentally get filtered). Cursor advancement preserved on skipped rows so the next poll doesn't keep re-fetching the self-notify. Notification callback also gated. 7 tests covering predicate truth table + 3 integrated _poll_once paths. Live-verified 2026-05-01.

PR #2471, opened 2026-05-02
feat(a2a-mcp): enrich channel envelope with peer name/role/agent_card_url

HongmingWang-Rabbit:
## Summary

The bare channel envelope only carried `peer_id` for `peer_agent` inbound, so a receiving agent had to round-trip to `/registry` to find out who's talking. This PR surfaces the sender's display name, role, and an agent-card URL alongside the routing fields so the agent can render \"ops-agent (sre): ping\" in one shot.

Wire shape change (additive — existing fields unchanged):

```diff
 \"meta\": {
   \"source\": \"molecule\",
   \"kind\": \"peer_agent\",
   \"peer_id\": \"<uuid>\",
   \"method\": \"message/send\",
   \"activity_id\": \"...\",
   \"ts\": \"...\",
+  \"peer_name\": \"ops-agent\",
+  \"peer_role\": \"sre\",
+  \"agent_card_url\": \"<PLATFORM_URL>/registry/discover/<uuid>\"
 }
```

`canvas_user` envelopes are unchanged (no peer to enrich).

## Implementation

- **a2a_client.py**: new `_peer_metadata` cache keyed by `peer_id` with `(fetched_at, record)` tuples. `enrich_peer_metadata(peer_id)` is sync, hits the cache first, falls back to `GET /registry/discover/<peer_id>` with a tight 2s timeout, returns None on validation/network/non-200 so callers can degrade gracefully. TTL is 5 min — long enough that a busy multi-peer chat doesn't hit registry on every push, short enough that role/name renames propagate within a session. `_agent_card_url_for(peer_id)` is deterministic from `peer_id` alone.

- **a2a_mcp_server.py**: `_build_channel_notification` calls `enrich_peer_metadata` when `peer_id` is non-empty; meta carries `peer_name`, `peer_role`, `agent_card_url` alongside the existing routing fields. `agent_card_url` surfaces unconditionally (constructable from peer_id alone); `peer_name`/`peer_role` only when registry lookup succeeds — never blocks the push on a registry stall.

## Test plan

- [x] 6 new branches in `test_a2a_mcp_server.py`:
  - `canvas_user` (peer_id='') → no enrichment fields, no registry GET
  - cache hit → no GET, fields populated from cache
  - cache miss → GET fires once, second push for same peer reuses cache
  - registry returns 500 → graceful degrade, peer_name/role absent, agent_card_url still present
  - cached entry past TTL → re-fetch fires, fresh fields surface
  - invalid peer_id (non-UUID) → no GET, fields absent (UUID validation guards URL construction)
- [x] All 33 existing `test_a2a_mcp_server.py` tests still pass (39/39 total)
- [x] Mutation-verified: reverting prod code fails 6/6 enrichment tests
- [ ] Live verify on staging workspace once published: peer-agent inbound surfaces with `peer_name=\"<sender-display-name>\"` in the channel tag

## Related

- molecule-core#2469 — RFC for the upstream `activity_type` rename (workspace-server)
- molecule-core#2470 — PR-A: defensive inbox poller filter for self-notify rows
- This is independent of both — pure envelope-shape addition.


hongmingwang-moleculeai (approved):
Channel envelope enrichment with peer name/role/agent_card_url. Wire-shape change is purely additive. TTL-cached registry lookup with 5-min freshness window matches the delegation-routing cache. Fail-open path returns the bare envelope (with deterministic agent_card_url for retry) when the registry is unreachable. Initial test had a monotonic-clock baseline bug — fixed and CI now green across all 19 checks.

PR #2472, opened 2026-05-02
feat(activity): add peer_id filter to /workspaces/:id/activity

HongmingWang-Rabbit:
## Summary

Surfaces conversation history with one specific peer for the wheel-side `chat_history` MCP tool (companion PR forthcoming). The filter joins `(source_id = \$X OR target_id = \$X)` so both inbound (peer was sender) and outbound (peer was recipient) turns appear in the same view, ordered by `created_at`, and composes cleanly with existing `type` / `source` / `since_secs` / `since_id` / `limit` filters.

Wire shape change (additive — existing callers unaffected):

```
GET /workspaces/:id/activity?peer_id=<uuid>
GET /workspaces/:id/activity?peer_id=<uuid>&type=a2a_receive&source=agent&limit=20
```

## Why a UUID guard?

`peer_id` is validated as a UUID at the trust boundary even though the args are bound parameters. Two reasons:
1. Cleaner 400 signal for the wheel-side caller — distinguishes \"you typed it wrong\" from \"empty result\" without round-tripping a no-rows response.
2. Defense in depth against any future code path that might interpolate the value into a URL or another query — the validator runs before the value is ever used outside SQL bind args.

## Test plan

- [x] 3 new branches in `activity_test.go`:
  - peer_id alone → query has `(source_id = \$X OR target_id = \$X)`, args bind correctly
  - peer_id + type + source compose into a single AND-chain in correct arg order
  - 5 malformed peer_ids (non-UUID, URL-encoded SQL injection, truncated, overlong, non-hex) all reject with 400
- [x] Existing `TestActivityList_*` suite still passes (no regression in the builder)
- [x] Mutation-verified: reverting activity.go fails all peer_id tests
- [ ] Live verify on staging: agent calls `chat_history(peer_id=...)` via the wheel and gets back the prior turns, ordered, paginated

## Related

- molecule-core#2469 — RFC for the upstream activity_type rename (workspace-server)
- molecule-core#2470 — PR-A: defensive inbox poller filter (lives in same PR sequence as this)
- molecule-core#2471 — PR-B: channel envelope enrichment (peer_name, peer_role, agent_card_url)
- Companion PR: `chat_history` MCP tool wrapping this route — opening next.

### DIFF
--- a/workspace-server/internal/handlers/activity.go
+++ b/workspace-server/internal/handlers/activity.go
@@ -15,6 +15,7 @@ import (
 	"github.com/Molecule-AI/molecule-monorepo/platform/internal/db"
 	"github.com/Molecule-AI/molecule-monorepo/platform/internal/events"
 	"github.com/gin-gonic/gin"
+	"github.com/google/uuid"
 )
 
 type ActivityHandler struct {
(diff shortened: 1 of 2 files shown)


hongmingwang-moleculeai (approved):
UUID parse at trust boundary (good — bound args are SQL-safe but malformed input deserves a clear 400). Filter binds source_id OR target_id with a single arg position so the chat_history MCP tool sees the full peer-conversation view. argIdx accounting + composition tests cover the builder. All 21 checks SUCCESS.

Issue #2486, opened 2026-05-02
Batch claude-code workspace provision silently drops — no log, no error, sweeper-only failure


HongmingWang-Rabbit:
## Symptom

User deployed a Director Pattern fleet (1 parent + 6 children, all `runtime=claude-code`) on the `hongming` prod tenant. **All 7 stuck in `provisioning` for >10min**, then sweeper marked them failed:

```
2026/05/02 01:34:22 Provision-timeout sweep: 0c5257dd-1c26-4414-8b32-6dc7165e80a7 (runtime="claude-code") stuck in provisioning > 10m0s — marked failed
2026/05/02 01:34:22 Provision-timeout sweep: 79d61bef-... — marked failed
2026/05/02 01:34:22 Provision-timeout sweep: cb8ef64b-... — marked failed
2026/05/02 01:34:22 Provision-timeout sweep: f3bc9c46-... — marked failed
2026/05/02 01:34:22 Provision-timeout sweep: a5798519-... — marked failed
2026/05/02 01:34:22 Provision-timeout sweep: 5fc4f325-... — marked failed
2026/05/02 01:34:22 Provision-timeout sweep: ed286836-... — marked failed
```

`workspaces.last_sample_error`: `"provisioning timed out — container started but never called /registry/register. Check container logs and network connectivity to the platform."`

## What's missing in the logs

Successful claude-code workspaces from the same tenant earlier in the day produced these lines:
```
CP provisioner: workspace dd40faf8-... → EC2 instance i-0948ab8866ddcc1ff (pending)
CPProvisioner: workspace dd40faf8-... started as machine i-0948ab8866ddcc1ff via control plane
```

**The 7 failed workspaces produced none of these lines.** Specifically — none of the four known exit paths in `workspace-server/internal/handlers/workspace_provision.go:provisionWorkspaceCP` logged anything:
- L648 `"CPProvisioner: prepare failed for X: msg"` — no
- L663 `"CPProvisioner: workspace start failed for X: err"` — no
- L677 `"CPProvisioner: persist instance_id failed for X: err"` — no
- L680 `"CPProvisioner: workspace X started as machine Y via control plane"` — no

So `provisionWorkspaceCP` either was never invoked, or returned without hitting any of its logged exits — both indistinguishable from "silent drop" at the operator level.

## Reproduction

1. Tenant on `hongming.moleculesai.app` (prod, EC2 `i-04e5197e96adb888f` us-east-2)
[...]

PR #2487, opened 2026-05-02
fix(provision): entry log + panic recovery on provision goroutines (#2486)

HongmingWang-Rabbit:
## Summary
- Add entry log at the top of \`provisionWorkspaceOpts\` + \`provisionWorkspaceCP\` so future stuck-in-provisioning incidents can distinguish "goroutine never started" from "started but exited via unlogged path."
- Add \`logProvisionPanic\` deferred recover at the same site — keeps a per-workspace panic from crashing the whole workspace-server (every other tenant workspace would die with it) and persists \`status='failed'\` so the canvas surfaces something instead of leaving the row stuck for 10min.

## Why this exists

Closes the visibility gap in #2486. Diagnosis:

> 7 claude-code workspaces (1 parent + 6 children Director Pattern) stuck in \`provisioning\` for >10min on prod tenant \`hongming\`. Sweeper marked them failed. Successful claude-code workspaces from earlier the same day produced \`CP provisioner: workspace X → EC2 instance Y\` logs. The 7 failed workspaces produced **none** of the four documented exit-path log lines in \`provisionWorkspaceCP\`. Operator can't tell if the goroutine even ran.

This PR doesn't root-cause the silent drop — that needs reproduction. It just makes the next repro debuggable + closes one branch ("could it have been a panic?") cleanly.

## Test plan
- [x] \`go test ./internal/handlers/ -run TestLogProvisionPanic\` — 3 new tests pass
- [x] Regression-injected: neuter the recover body → all 3 tests fail (verified)
- [x] Clean build: \`go build ./...\`
- [ ] After merge: redeploy hongming tenant, repro the 7-workspace burst, confirm entry log appears for each goroutine

## Related
- Issue #2486 (this PR addresses observability; root cause TBD)
- Task #124 (in_progress) — broader claude-code provisioning investigation

### DIFF
--- a/workspace-server/internal/handlers/workspace_provision.go
+++ b/workspace-server/internal/handlers/workspace_provision.go
@@ -6,7 +6,9 @@ import (
 	"log"
 	"os"
 	"path/filepath"
+	"runtime/debug"
 	"strings"
+	"time"
 
 	"github.com/Molecule-AI/molecule-monorepo/platform/internal/crypto"
 	"github.com/Molecule-AI/molecule-monorepo/platform/internal/db"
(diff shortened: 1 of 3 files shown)


hongmingwang-moleculeai (approved):
Closes the observability gap in #2486 cleanly. Entry log + panic recovery is no-regret. Tests + regression injection confirm the recover/log/persist contract.

hongmingwang-moleculeai (approved):
Re-approval after revision: routes panic recovery through markProvisionFailed (canvas broadcast + drift-gate-clean), captureLog helper restores prev writer correctly. Drift-gate failure root-caused. Regression injection still trips both panic tests.

PR #2491, opened 2026-05-02
test(provision): harden panic tests with re-raise guard + broadcast count

HongmingWang-Rabbit:
Post-merge follow-up to #2487 review feedback. Three tightenings to the panic-test suite landed in #2487:

1. **`guardAgainstReraise(fn)` shared helper** wraps every panic-test exercise. Original `RecoversAndMarksFailed` had its own outer `recover()`; the other two didn't, so a future regression that re-raised from `logProvisionPanic` would have crashed the test process instead of reporting a clean failure.

2. **Assert `bcast.count == 7`** in the concurrent repro. The `concurrentSafeBroadcaster.count` field was added during the -race fix but not consumed. Cross-checks the recorder-set assertion from a different angle — pins that every goroutine that reaches `cpProv.Start` also broadcasts its failure (canvas-broadcast contract).

3. **Comment on `captureLog`** noting `log.SetOutput` is process-global and incompatible with `t.Parallel()` — preempts a future footgun.

## Test plan
- [x] `go test -race ./internal/handlers/ -run 'TestProvisionWorkspaceCP_ConcurrentBurst|TestLogProvisionPanic' -v` — all 4 pass
- [x] `go test -race ./internal/handlers/ ./internal/db/` — full packages green

### DIFF
--- a/workspace-server/internal/handlers/workspace_provision_concurrent_repro_test.go
+++ b/workspace-server/internal/handlers/workspace_provision_concurrent_repro_test.go
@@ -189,6 +189,22 @@ func TestProvisionWorkspaceCP_ConcurrentBurst_NoSilentDrop(t *testing.T) {
 		}
 	}
 
+	// Assertion 4: every goroutine's failure path called RecordAndBroadcast
+	// exactly once (via h.markProvisionFailed inside provisionWorkspaceCP's
+	// "start failed" arm). Cross-checks Assertion 2 from a different angle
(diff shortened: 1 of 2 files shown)


hongmingwang-moleculeai (approved):
Defensive test hardening — guardAgainstReraise stops a future regression from aborting the test process; bcast.count assertion actually uses the field added in the race fix. Both 2-line wins.

PR #2493, opened 2026-05-02
harness(phase-2): multi-tenant compose + cross-tenant isolation replays

HongmingWang-Rabbit:
## Summary
- **Multi-tenant harness** — second `tenant-beta` + `postgres-beta` alongside alpha, distinct `MOLECULE_ORG_ID` + admin token per tenant. Mirrors production's per-tenant EC2+DB isolation.
- **cf-proxy routes by Host header** — `map \$host \$tenant_upstream` + explicit `resolver 127.0.0.11` (variables in `proxy_pass` need per-request DNS), exactly matching prod CF tunnel routing.
- **Two new replays** — `tenant-isolation.sh` (13 assertions: TenantGuard 404s on org mismatch, no body leak, allowlist sanity) + `per-tenant-independence.sh` (12 assertions: parallel A2A both tenants + concurrent INSERT race that catches lib/pq prepared-statement cache collision).
- **Bug fix carried with the rename** — `down.sh` was silently exiting non-zero on `SECRETS_ENCRYPTION_KEY` validation, leaking workspaces+activity_logs into subsequent boots. Caught when `tenant-isolation.sh` F1/F2 saw 3× duplicate alpha-parent rows. Same fix applied to the workflow's dump-logs step.
- **Workflow plumbing** — pin `molecule-ai-workspace-runtime>=0.1.78` so `channel-envelope-trust-boundary.sh` (which imports from `molecule_runtime.*` to catch wheel-build drift) actually has the wheel; rename `tenant`/`postgres` log dumps to `tenant-alpha`/`tenant-beta`/`postgres-alpha`/`postgres-beta`; drop the now-unnecessary `/etc/hosts` step (Phase 0 already migrated bash scripts to header-based routing).

## What this catches that single-tenant didn't
The four prior replays exercise the SaaS request path against one tenant. They cannot prove that:
- TenantGuard rejects a misrouted request (production CF tunnel + AWS LB are the failure surface — TenantGuard allowlist drift shipped #2398).
- Two tenants doing legitimate work in parallel keep their `activity_logs` / `workspaces` / connection-pool state partitioned (lib/pq prepared-statement cache collision is documented as an org-wide hazard).

Both replays target those failure modes against the production `Dockerfile.tenant` image — not a unit-test stub.

## Test plan
- [x] `./run-all-replays.sh` from a clean state — 6/6 passed (buildinfo-stale-image, channel-envelope-trust-boundary, chat-history, peer-discovery-404, per-tenant-independence, tenant-isolation)
- [x] `down.sh` cleans volumes (no \"SECRETS_ENCRYPTION_KEY missing\" error)
- [x] YAML lint passes on the workflow
- [ ] `Harness Replays` CI check goes green on this PR

### DIFF
--- a/.github/workflows/harness-replays.yml
+++ b/.github/workflows/harness-replays.yml
@@ -106,16 +106,6 @@ jobs:
           path: molecule-ai-plugin-github-app-auth
           token: ${{ secrets.PLUGIN_REPO_PAT || secrets.GITHUB_TOKEN }}
 
-      - name: Add /etc/hosts entry for harness-tenant.localhost
-        # ubuntu-latest doesn't auto-resolve *.localhost the way macOS
-        # sometimes does. seed.sh + replay scripts curl
-        # http://harness-tenant.localhost:8080 — without the entry
-        # they'd fail with getaddrinfo ENOTFOUND.
-        if: needs.detect-changes.outputs.run == 'true'
-        run: |
-          echo "127.0.0.1 harness-tenant.localhost" | sudo tee -a /etc/hosts >/dev/null
-          getent hosts harness-tenant.localhost
-
       - name: Install Python deps for replays
         # peer-discovery-404 (and future replays) eval Python against the
         # running tenant — importing workspace/a2a_client.py pulls in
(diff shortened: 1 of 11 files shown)


hongmingwang-moleculeai (approved):
Phase 2 brings real production-shape isolation testing into the harness — TenantGuard 404 enforcement + per-tenant DB partition + concurrent-INSERT race that catches the lib/pq cache-collision class. cf-proxy nginx is correct (variable-in-proxy_pass needs explicit resolver — the comment captures this). down.sh fix is the highest-leverage change in the PR: it was silently leaking volumes across runs. All 6 replays pass clean. LGTM.

Issue #2332, opened 2026-04-30
Workspace awareness: agents don't surface platform capabilities up front; CLI-runtime adapters can't message the user


HongmingWang-Rabbit:
Follow-up from the dogfooding session — \"workspaces seem to be not knowing what they can do natively with platform.\" Investigation result is below.

## What's actually wired

8 MCP tools exposed to MCP-capable adapters (claude-code, hermes, langchain, crewai), all sourced from \`workspace/platform_tools/registry.py\`:

\`delegate_task\`, \`delegate_task_async\`, \`check_task_status\`, \`list_peers\`, \`get_workspace_info\`, \`send_message_to_user\`, \`commit_memory\`, \`recall_memory\`

System prompt sections (in order):

1. Platform Instructions
2. Prompt files
3. MEMORY.md / USER.md snapshots
4. Parent context
5. Platform Rules
6. Platform Guidelines
7. Skills
8. Inter-Agent Communication (tools + when_to_use)
9. Hierarchical Memory (tools + when_to_use)
10. Peer capabilities
11. Delegation failure handling

### 1. No top-level Platform Capabilities summary

Agents read top-down. Section 8 is where they learn about A2A — by then the model has already started reasoning about what it'll do. A 6-line block at section 1.5 — \"## Platform Capabilities — you have native access to: peer collaboration (delegate_task/async/check_status, list_peers, get_workspace_info), user messaging (send_message_to_user), persistent memory (commit_memory, recall_memory)\" — would make the toolkit visible upfront.

**Fix:** add a generated \"Platform Capabilities\" preamble in \`build_system_prompt\` immediately after Platform Instructions, generated from the registry (so it auto-updates when tools change).

### 2. CLI-runtime adapters can't message the user

\`workspace/executor_helpers.py:_CLI_A2A_COMMAND_KEYWORDS\` explicitly maps \`send_message_to_user\` to \`None\` — meaning agents on CLI-runtime adapters (ollama, custom subprocess) **cannot surface anything to the user** because the \`a2a_cli\` subprocess interface has no \`say\` / \`message\` subcommand.

\`\`\`python
\"send_message_to_user\": None,  # not exposed via CLI subprocess
\`\`\`

This makes CLI agents silently fail instead of asking for help. Likely contributes to the \"agent goes idle when stuck\" pattern observed during dogfooding.

[...]

PR #2507, opened 2026-05-02
fix(canary): reframe smoke prompt to give GPT-4o explicit permission to echo

HongmingWang-Rabbit:
## Summary

Canary smoke test went from ~100% pass to **30% flake** starting 2026-05-01 22:11 with model-refusal replies:

> *\"I'm unable to do that.\"*
> *\"I'm unable to fulfill that request. Can I assist you with anything else?\"*
> *\"I'm unable to reply with responses that don't allow me to fulfill tasks…\"*

3 fails / 10 recent runs.

## Root cause

The Platform Capabilities preamble (#2332, 2026-04-30) added \`\"Use them proactively\"\` to the top of every system prompt. Combined with the heavy A2A + HMA tool docs further down, GPT-4o reads the canary's contrived bare-echo prompt (\`\"Reply with exactly: PONG\"\`) as out-of-role and intermittently refuses.

**Real user prompts don't hit this** — only the synthetic smoke prompt does. So the platform's prompt is correct (we *want* agents primed toward tool use); the fix belongs in the canary.

## Change

Two strings: parent + child smoke prompts now read

> \"This is the platform smoke test verifying agent wiring. No tools or memory are needed — please respond with exactly the single token: PONG\"

Plus a load-bearing comment so a future reader doesn't trim back to the brittle phrasing.

## Test plan

- [x] \`bash -n\` syntax check passes
- [x] python json round-trip on the new payload validates
- [ ] CI green
- [ ] Next 2-3 canary runs go green (the actual validation — current 30% flake means seeing 3 in a row gives ~95% confidence)

## Out of scope (separate concerns)

- Demo readiness: real demo prompts won't trigger this refusal pattern. Canary is the only victim.
- Long-term: if other smoke tests adopt the same bare-echo phrasing, they'll need the same treatment. Not addressing those preemptively — apply only when they break.

### DIFF
--- a/tests/e2e/test_staging_full_saas.sh
+++ b/tests/e2e/test_staging_full_saas.sh
@@ -433,6 +433,19 @@ done
 
 # ─── 8. A2A round-trip on parent ───────────────────────────────────────
 log "8/11 Sending A2A message to parent — expecting agent response..."
+# Smoke prompt phrasing — DO NOT trim back to the bare "Reply with exactly: PONG"
+# version that ran here pre-2026-05-02. After the Platform Capabilities preamble
+# (#2332, 2026-04-30) landed in the system prompt, GPT-4o began intermittently
+# refusing the bare echo prompt with messages like:
+#   - "I'm unable to do that."
+#   - "I'm unable to fulfill that request. Can I assist you with anything else?"
+#   - "I'm unable to reply with responses that don't allow me to fulfill tasks…"
+# 3 fails / 10 runs ≈ 30% flake. Root cause: the preamble primes the model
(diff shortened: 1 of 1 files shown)


hongmingwang-moleculeai (approved):
LGTM. Diagnosis is well-supported by the timeline correlation (#2332 landed 2026-04-30 → flakes start 2026-05-01 22:11) and three same-shape refusals. Locating the fix in the canary rather than the platform prompt is the right call — we want the prompt priming agents toward tool use; the canary's contrived task is the outlier. Comment will save the next person from naively shortening the phrasing.

PR #2509, opened 2026-05-02
fix(a2a): send v1 file Part shape; tolerate v1 server-side

HongmingWang-Rabbit:
## Summary

Image-only chats surface `Error: message contained no text content.` because canvas posts the v0 file-part shape `{kind:"file", file:{uri,name,mimeType}}` that the workspace runtime's a2a-sdk v1 protobuf parser silently drops. v1 `Part` has fields `[text, raw, url, data, metadata, filename, media_type]` — there is no `kind`, no nested `file`. With `ignore_unknown_fields=True`, `ParseDict` drops both legacy keys, producing a fully-empty Part. No text, no extracted attachments → the executor's `"no text content"` guard fires.

Three coordinated changes:

1. **canvas/ChatTab.tsx** — outbound file parts now use the v1 flat shape `{url, filename, mediaType}` so the v1 parser populates Part fields instead of dropping them.
2. **workspace/executor_helpers.py** — `extract_attached_files` learns a v1 detection branch (non-empty `part.url` + `filename` + `media_type`) alongside the existing v0 RootModel and flat-file shapes. Defends every runtime mounting the OSS wheel against any client still on the v0 wire.
3. **canvas/message-parser.ts** — `extractFilesFromTask` tolerates the v1 shape on incoming agent responses, so file chips render in chat history regardless of which Part shape the runtime emits.

Repro:
\`\`\`python
from a2a.types import a2a_pb2
from google.protobuf.json_format import ParseDict
v0 = {"kind":"file","file":{"uri":"workspace:/x.png","name":"x.png","mimeType":"image/png"}}
p = ParseDict(v0, a2a_pb2.Part(), ignore_unknown_fields=True)
# All fields empty → silently dropped by extract_attached_files (pre-fix).
\`\`\`

## Test plan

- [x] \`workspace/tests/test_executor_helpers.py\` — new v1-shape test + empty-Part defense (`86 passed`)
- [x] \`canvas/.../message-parser.test.ts\` — new v1 protobuf flat-shape extraction test + URL-basename fallback (`91 passed`)
- [x] \`tsc --noEmit\` clean across canvas
- [ ] After merge + cascade: send image-only chat in hongming Claude Code Agent workspace; expect agent to receive the image (manifest line in prompt) instead of `"no text content"` error

Surfaced 2026-05-01 in hongming workspace after the `/workspace` chown fix unblocked uploads.

### DIFF
--- a/canvas/src/components/tabs/ChatTab.tsx
+++ b/canvas/src/components/tabs/ChatTab.tsx
@@ -32,10 +32,23 @@ interface A2AFileRef {
   bytes?: string;
   size?: number;
 }
+// A2A Part shape — covers both v0 (Pydantic discriminated union with
+// `kind: "text" | "file"`) and v1 (a2a-sdk protobuf with flat fields:
+// `text` for text parts; `url` + `filename` + `mediaType` for file
+// parts; no `kind` discriminator at all). Outbound we now send v1 for
+// file parts because the v1 protobuf parser drops the v0 keys via
+// `ignore_unknown_fields=True`, surfacing as the user-visible
+// "Error: message contained no text content" on image-only chats
+// (2026-05-01 hongming incident). Text parts stay accident-compatible
+// across v0/v1 because the field name is `text` in both.
 interface A2APart {
-  kind: string;
+  kind?: string;
   text?: string;
   file?: A2AFileRef;
+  // v1 file-part fields (flat — no nested `file` object):
+  url?: string;
+  filename?: string;
+  mediaType?: string;
 }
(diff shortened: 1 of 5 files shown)


hongmingwang-moleculeai (approved):
Reviewed: a2a v0/v1 Part shape coordination is the right shape. Canvas outbound v1, runtime tolerant of v0+v1, parser tolerant of v0+v1 on incoming. Tests pin both shapes on both sides. Fix matches root cause (json_format ignore_unknown_fields silent-drop).

PR #2510, opened 2026-05-02
fix(canvas): revert v1 outbound file part shape — JSON-RPC layer rejects it

HongmingWang-Rabbit:
## Summary

Hotfix on top of PR #2509. Live test showed a2a-sdk's JSON-RPC layer validates against the **v0 Pydantic discriminated union** (`TextPart | FilePart | DataPart`), so the v1 flat shape `{url, filename, mediaType}` is rejected with:

\`\`\`
Invalid Request:
  params.message.parts.0.TextPart.text — Field required
  params.message.parts.0.FilePart.file — Field required
  params.message.parts.0.DataPart.data — Field required
\`\`\`

That means every chat message with attachments would 400 against the workspace once Vercel deploys the canvas with the v1 outbound shape — regression.

The actual root cause of the original user-visible "Error: message contained no text content" was the missing `/workspace` chown that already shipped via CP PR #381 (`workspace_dir mkdir+chown`) + test pin #382. Verified end-to-end by sending an image-only A2A message with v0 shape against the freshly-provisioned hongming Claude Code Agent — agent received the file, read its bytes, and replied normally:

> I can see you've attached an image (v0v1-test.png), but it appears to be blank/empty… the file arrived at \`/workspace/.molecule/chat-uploads/…\` and I can read it.

## Scope of revert

Only canvas outbound. Defensive v1-tolerance stays in:
- \`workspace/executor_helpers.py\` — \`extract_attached_files\` still accepts v1 protobuf parts (future-proof, harmless on v0 hot path)
- \`canvas/message-parser.ts\` — \`extractFilesFromTask\` still tolerates v1 shape on incoming agent responses

## Test plan

- [x] Canvas tests still green: 91/91
- [x] Workspace tests still green: 86/86
- [x] Live test against hongming with v0 shape post-PR #381 + restart — agent receives image, no "no text content" error
- [ ] Confirm Vercel canvas deploy from main does NOT include the v1 outbound shape (this PR cherry-flips back before any user action)

### DIFF
--- a/canvas/src/components/tabs/ChatTab.tsx
+++ b/canvas/src/components/tabs/ChatTab.tsx
@@ -32,23 +32,19 @@ interface A2AFileRef {
   bytes?: string;
   size?: number;
 }
-// A2A Part shape — covers both v0 (Pydantic discriminated union with
-// `kind: "text" | "file"`) and v1 (a2a-sdk protobuf with flat fields:
-// `text` for text parts; `url` + `filename` + `mediaType` for file
-// parts; no `kind` discriminator at all). Outbound we now send v1 for
-// file parts because the v1 protobuf parser drops the v0 keys via
-// `ignore_unknown_fields=True`, surfacing as the user-visible
-// "Error: message contained no text content" on image-only chats
-// (2026-05-01 hongming incident). Text parts stay accident-compatible
-// across v0/v1 because the field name is `text` in both.
+// A2A Part — outbound matches the v0 Pydantic discriminated-union
+// shape that a2a-sdk's JSON-RPC layer validates against (TextPart |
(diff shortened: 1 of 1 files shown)


hongmingwang-moleculeai (approved):
Reverting v1 outbound back to v0. Live test against hongming with v0 + freshly chowned workspace confirms agent receives image-only message correctly. v1 outbound was a wrong hypothesis — JSON-RPC layer validates against v0 Pydantic union and rejects v1 with 'Field required'.

hongmingwang-moleculeai (approved):
Reverting v1 outbound. Live test confirmed JSON-RPC layer rejects v1 flat shape.

PR #2535, opened 2026-05-03
Persist canvas-selected model+provider on first deploy

HongmingWang-Rabbit:
## Summary

Fixes the load-bearing bug where hermes workspaces fail provisioning with `container started but never called /registry/register` whenever the canvas selects a non-default model.

## The bug chain

1. Canvas POSTs `/workspaces` with `{model: "minimax/MiniMax-M2.7-highspeed"}`.
2. The workspace row is created and the FIRST provision succeeds — `payload.Model` is plumbed through to `applyRuntimeModelEnv` and the container boots with `HERMES_DEFAULT_MODEL=minimax/...`.
3. **But neither `MODEL_PROVIDER` nor `LLM_PROVIDER` is ever persisted to `workspace_secrets`.**
4. On the next Restart, the payload is rebuilt from the `workspaces` row (no `model` column), so `payload.Model` is empty.
5. `applyRuntimeModelEnv` falls back to `envVars["MODEL_PROVIDER"]` — also empty (never written).
6. `HERMES_DEFAULT_MODEL` is never exported, the container's `start.sh` falls through to its hardcoded default (`nousresearch/hermes-4-70b`), `derive-provider.sh` produces the wrong provider, the hermes gateway 401s, `/health` poll fails, `molecule-runtime` never registers.

Symptom: failed-workspace `95ed3ff2` (2026-05-02).

Compounding: even on the first provision, `LLM_PROVIDER` is never written (canvas doesn't send `provider`), so CP user-data omits the `provider:` field from `/configs/config.yaml` and `derive-provider.sh` falls through to `PROVIDER="auto"` for every custom-prefix slug — pre-empting the runtime detection that would have picked the right provider from a Molecule-managed key.

## The fix

After the workspace row commits in `WorkspaceHandler.Create`:

1. Persist `MODEL_PROVIDER = payload.Model` (when non-empty) to `workspace_secrets`. This is the row `applyRuntimeModelEnv`'s fallback chain reads on Restart.
2. Persist `LLM_PROVIDER = deriveProviderFromModelSlug(payload.Model)` to `workspace_secrets` (only when the prefix is recognized — unknown prefixes return `""` so `derive-provider.sh`'s `*=auto` branch keeps the final say at runtime).

`deriveProviderFromModelSlug` is a Go translation of the case statement in `workspace-configs-templates/hermes/scripts/derive-provider.sh` covering all the prefixes the shell script handles. Two intentional differences:

- `nousresearch/*` and `openai/*` both default to `openrouter` here — the shell script's `HERMES_API_KEY`/`OPENAI_API_KEY` checks need runtime env we may not have at provision time. `derive-provider.sh` can still upgrade to `nous`/`custom` at boot when the keys are present.
- Unknown prefixes return `""` (skip the write) instead of `"auto"` — persisting `auto` would block a future `Save+Restart` with a known prefix from re-deriving.

`derive-provider.sh` remains the source of truth at runtime; this is strictly a *gating* hint that survives restarts.

## Implementation

- `secrets.go`: extracted `setModelSecret` and `setProviderSecret` (lowercase) helpers from the existing `SetModel`/`SetProvider` gin handlers. Both now reuse the same encryption + upsert logic — no SQL is inlined in the create path.
- `workspace_provision.go`: added `deriveProviderFromModelSlug(model string) string` directly above `applyRuntimeModelEnv`. `applyRuntimeModelEnv` itself is unchanged.
- `workspace.go`: added the two-line persistence call in `WorkspaceHandler.Create` right after `tx.Commit()`. Failures are logged and non-fatal — the workspace row stays consistent and a later `Save+Restart` via the canvas can re-persist.

## Test plan

- [x] `go build ./...` clean
- [x] `go vet ./...` clean
- [x] `go test ./internal/handlers/... -run "DeriveProvider|FirstDeploy|MintModelProvider" -v` — 4 new test functions, 31 sub-cases pass
- [x] Existing `SetModel`/`SetProvider` tests still pass (refactor invariance)
- [x] Full handlers package test suite still passes
- [ ] Staging E2E: provision a hermes workspace with `model=minimax/MiniMax-M2.7`, restart, confirm it stays on minimax (does not regress to `nousresearch/hermes-4-70b`)

New tests added (`workspace_provision_shared_test.go`):

- `TestDeriveProviderFromModelSlug` — table-driven, 28 cases including all known prefixes + unknown + edge cases (empty / no-slash / leading-slash)
- `TestWorkspaceCreate_FirstDeploy_PersistsModelAndProvider` — happy path with `minimax/MiniMax-M2.7`, asserts both rows are written
- `TestWorkspaceCreate_FirstDeploy_NoModel_NoSecretWritten` — empty `payload.Model` writes neither row
- `TestWorkspaceCreate_FirstDeploy_UnknownModel_OnlyMintModelProvider` — unknown prefix writes `MODEL_PROVIDER` but skips `LLM_PROVIDER`

### DIFF
--- a/workspace-server/internal/handlers/secrets.go
+++ b/workspace-server/internal/handlers/secrets.go
@@ -467,6 +467,35 @@ func (h *SecretsHandler) GetModel(c *gin.Context) {
 	c.JSON(http.StatusOK, gin.H{"model": string(decrypted), "source": "workspace_secrets"})
 }
 
+// setModelSecret writes (or clears, when value=="") the MODEL_PROVIDER
+// workspace secret. Extracted from SetModel so non-handler call sites
+// (notably WorkspaceHandler.Create — first-deploy path that persists the
+// canvas-selected model so applyRuntimeModelEnv's restart fallback finds
+// it) can reuse the encryption + upsert logic without inlining the SQL.
+//
+// Returns nil on success. Caller is responsible for any restart trigger;
+// the gin handler re-adds that after a successful write.
+func setModelSecret(ctx context.Context, workspaceID, model string) error {
+	if model == "" {
+		_, err := db.DB.ExecContext(ctx,
+			`DELETE FROM workspace_secrets WHERE workspace_id = $1 AND key = 'MODEL_PROVIDER'`,
+			workspaceID)
+		return err
+	}
+	encrypted, err := crypto.Encrypt([]byte(model))
+	if err != nil {
+		return err
+	}
+	version := crypto.CurrentEncryptionVersion()
+	_, err = db.DB.ExecContext(ctx, `
+		INSERT INTO workspace_secrets (workspace_id, key, encrypted_value, encryption_version)
+		VALUES ($1, 'MODEL_PROVIDER', $2, $3)
+		ON CONFLICT (workspace_id, key) DO UPDATE
+			SET encrypted_value = $2, encryption_version = $3, updated_at = now()
+	`, workspaceID, encrypted, version)
+	return err
+}
+
 // SetModel handles PUT /workspaces/:id/model — writes the model slug
 // into workspace_secrets as MODEL_PROVIDER (the key GetModel reads).
 // For hermes, the value is a hermes-native slug like "minimax/MiniMax-M2.7";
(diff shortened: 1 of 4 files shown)


hongmingwang-moleculeai (approved):
LGTM. Reuses extracted setModelSecret/setProviderSecret helpers to avoid SQL drift, the deriveProviderFromModelSlug mapping matches derive-provider.sh, and the unknown-prefix returns empty decision keeps the runtime check in derive-provider.sh as the source of truth. The 4 new tests pin all three branches (happy path, empty model, unknown prefix) plus the 28-case slug to provider table.

Issue #246, opened 2026-04-15
template: Security Auditor needs notification channel for critical findings


HongmingWang-Rabbit:
## Summary

The Security Auditor workspace has no notification channel configured. Critical security findings have no push path to human operators — they exist only in memory/logs until someone manually checks.

## Current state
`org-templates/molecule-dev/org.yaml` — Security Auditor workspace:
- `channels:` — **not set**
- Only PM has a channel (Telegram)

## Why this matters now
- Issue #241 was filed today: YAML injection + path traversal in `workspace_provision.go` (MEDIUM severity)
- Issues #221, #226 are prior instances of the same vulnerability class
- If Security Auditor's 12h cron had fired and found #241 at night, there would be no push notification to the team

## Proposed fix
Add an email (or Slack) channel to Security Auditor, gated to `severity >= HIGH`:
```yaml
channels:
  - type: email
    config:
      to: ${SECURITY_ALERT_EMAIL}
    enabled: true
    filter: critical_findings_only
```

## Also consider
DevOps Engineer should have a build-break notification channel (separate issue filed).

_Filed by template-fitness audit 2026-04-15T19:39Z_

HongmingWang-Rabbit:
Triage note: this is the template-side consumer for either #243 (Slack) or #245 (Telegram). Blocked until one of those adapter decisions lands. Once an adapter ships, the Security Auditor template needs 3 additions:
1. \`channels:\` block in \`defaults.yaml\` pointing at the chosen adapter
2. New system-prompt line telling the Security Auditor to route CRITICAL findings via the channel instead of only filing issues
3. A pre-fire check in the cron prompt so we don't spam the channel with duplicate findings across ticks

_triage comment from overnight agent tick_

PR #2538, opened 2026-05-03
fix(runtime): canvas-picked model wins universally + per-model required_env

HongmingWang-Rabbit:
## Summary

Two surgical edits to `workspace/` that fix Bug B (canvas-picked model silently dropped for templated workspaces) and Bug D (preflight rejects valid auth for non-default models). Both surfaced 2026-05-02 during the MiniMax E2E.

- **Bug B** (`workspace/config.py`): `load_config` now resolves `runtime_config.model` as `os.environ.get(\"MODEL_PROVIDER\") or runtime_raw.get(\"model\") or model`. Centralising in the parser means **every adapter** (claude-code, hermes, codex, langgraph, future ones) gets canvas-picked-model passthrough for free — no per-adapter env-reading code.
- **Bug D** (`workspace/preflight.py`): when `runtime_config.models[]` declares per-entry `required_env`, preflight matches the picked model id (case-insensitive) and uses that entry's `required_env` outright. **REPLACE semantics, not union** — different models have different auth paths (OAuth token vs API key vs third-party provider key); unioning would re-introduce the very crash-loop this fix closes.
- Surface (`workspace/config.py`): `RuntimeConfig` now carries `models: list[dict]` so the canvas Model dropdown source flows through to preflight without forcing the parser schema to grow.

## What this fixes

Picking MiniMax-M2.7 in canvas → workspace-server plumbed `MODEL_PROVIDER=MiniMax-M2.7` correctly → workspace booted with **sonnet** because the template's verbatim \`runtime_config.model: sonnet\` always won. Then preflight failed with \`missing CLAUDE_CODE_OAUTH_TOKEN\` despite the user having set MINIMAX_API_KEY. After this PR, both paths honour the picked model.

## Test plan

- [x] \`pytest workspace/tests/test_preflight.py workspace/tests/test_config.py\` — 75/75 pass locally
- [x] 9 new preflight tests covering: per-model match (case-insensitive), REPLACE not union, fallback to top-level when no \`models[]\` or no match, multi-entry, malformed entries dropped, picked-model unset, picked-model not in list
- [x] Existing config parser tests untouched and green
- [ ] CI must verify full coverage gate (≥86%)
- [ ] Sibling PR against \`molecule-ai-workspace-template-claude-code\` lands the per-template defensive fixes + boot debug logging that complement this universal fix

Closes part of #246.

### DIFF
--- a/workspace/config.py
+++ b/workspace/config.py
@@ -100,6 +100,16 @@ class RuntimeConfig:
                                # "minimax"). Falls back to the top-level resolved
                                # provider when empty. Adapters (hermes, claude-code,
                                # codex) prefer this over slug-parsing the model name.
+    # Per-model entries surfaced in the canvas Model dropdown. Each entry is a
+    # raw dict with at least ``id``; ``required_env`` is the per-model auth
+    # list (e.g. ``{"id": "MiniMax-M2.7", "required_env": ["MINIMAX_API_KEY"]}``).
+    # Preflight prefers an entry's ``required_env`` over the top-level
+    # ``required_env`` when the picked ``model`` matches an entry's ``id``
+    # (case-insensitive). The top-level list remains the fallback so single-
+    # model templates need not migrate. Surfaced 2026-05-02 after a user
+    # picked MiniMax in canvas, set MINIMAX_API_KEY, and still got booted
+    # into a CLAUDE_CODE_OAUTH_TOKEN preflight failure.
+    models: list[dict] = field(default_factory=list)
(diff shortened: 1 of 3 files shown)


hongmingwang-moleculeai (approved):
Approving — universal MODEL_PROVIDER fix in workspace/config.py is the right shape (parser-level, every adapter benefits). Per-model required_env REPLACE semantics in preflight is correct (different models have different auth paths, union would re-introduce the crash-loop). 9 new preflight tests cover the matrix. CI green.

PR #2543, opened 2026-05-03
fix(workspace-server): surface structured provider registry on /templates (#235)

HongmingWang-Rabbit:
## Summary

Add `provider_registry` field to `/templates` response, sourced from the template's TOP-LEVEL `providers:` block. Each entry carries `name`, `auth_mode`, `model_prefixes`, `model_aliases`, `base_url`, `auth_env`. Existing `providers` slug list (from `runtime_config.providers`) is unchanged — both shapes coexist and are independent.

## Why

Audit #253 (closed-at-code-merged failure mode audit) flagged task #235 as a HIGH-severity contract drift: marked completed, but `templates.go` only ever emitted the slug list — the structured ProviderEntry shape the description promised was never plumbed.

In production today:
- `claude-code` ships a top-level `providers:` block with 6+ structured entries (anthropic-oauth, anthropic-api, xiaomi-mimo, minimax, zai, moonshot…)
- `hermes` still uses `runtime_config.providers: [nous, openrouter, …]` slug list
- canvas's `ProviderModelSelector` derives provider taxonomy from per-model `required_env` because the structured payload was never surfaced

The canvas comment already anticipates this fix: *"Vendor is inferred via prefix rules below. Templates that ship explicit vendor metadata (future) should override the heuristic."* This PR makes that "future" the present — the canvas can optionally drop its prefix-inference fallback for registry-shipping templates in a separate PR.

## Backwards compat

Purely additive. The new field uses `omitempty` so templates without the top-level block emit nothing (not `null` — that would break Array.isArray on the canvas). Existing slug list (`providers`) is untouched, all 5 existing tests in `templates_test.go` still pass.

## Test plan

- [x] `TestTemplatesList_SurfacesProviderRegistry` — claude-code-shaped fixture, asserts:
  - 2 entries in declaration order
  - field plumbing on first (oauth) and second (third-party) entries
  - wire-shape gate (`provider_registry` JSON tag) to catch struct-tag renames
- [x] `TestTemplatesList_OmitsProviderRegistryWhenAbsent` — hermes-shaped fixture, asserts:
  - new field absent (omitempty)
  - existing `providers` slug list still present
- [x] All 9 `TestTemplatesList_*` tests pass locally
- [ ] CI: `Platform (Go)`

Closes #254 / reconciles #235.

### DIFF
--- a/workspace-server/internal/handlers/templates.go
+++ b/workspace-server/internal/handlers/templates.go
@@ -45,6 +45,28 @@ type modelSpec struct {
 	RequiredEnv []string `json:"required_env,omitempty" yaml:"required_env"`
 }
 
+// providerRegistryEntry mirrors a row from a template's top-level
+// `providers:` registry block (claude-code, hermes, etc.). Each entry
+// fully describes one provider: its name, auth flow, the model id
+// prefixes/aliases that route to it, an optional base_url override, and
+// the env vars required to authenticate.
+//
+// This is the structured taxonomy the canvas's ProviderModelSelector
+// comment anticipates ("Templates that ship explicit vendor metadata
+// (future) should override the heuristic.") — surfacing it here lets
+// the canvas drop its prefix-inference fallback for templates that ship
+// an explicit registry. Templates without the block omit the field
+// (omitempty); the canvas falls back to its current per-model
+// required_env derivation.
+type providerRegistryEntry struct {
+	Name          string   `json:"name" yaml:"name"`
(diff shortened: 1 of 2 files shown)


hongmingwang-moleculeai (approved):
Closes the #235 contract drift caught by audit #253. Pure additive backend change: new `provider_registry` field reading top-level `providers:` block, slug list unchanged, omitempty so legacy templates emit nothing. 3 new tests including order preservation + wire-shape gate.

PR #2558, opened 2026-05-03
fix(a2a): enqueue Task before TaskStatusUpdateEvent for v1 SDK contract

HongmingWang-Rabbit:
## Summary

`a2a-sdk` ≥ 1.0 raises `InvalidAgentResponseError` when an executor publishes a `TaskStatusUpdateEvent` (e.g. via `TaskUpdater.start_work`) before any `Task` event for fresh requests. The framework only auto-creates the Task on continuation messages (existing `task_id` resolves via `task_manager.get_task`); new requests leave `_task_created` unset and the SDK validation at `a2a/server/agent_execution/active_task.py` rejects the first status update.

PR #2170 migrated the executor surface to v1 but missed this contract. The synthetic E2E gate has caught it on every staging run for ~7 hours+ (3 consecutive runs against 3 different commits, all failing identically):

```
{"jsonrpc":"2.0","id":"e2e-msg-1","error":{"code":-32603,
 "message":"Agent should enqueue Task before TaskStatusUpdateEvent
 event","data":null}}
```

## Fix

In `_core_execute()`, enqueue a `Task(state=SUBMITTED)` before the `TaskUpdater` is constructed, gated on `context.current_task is None` so continuation messages don't double-enqueue (which the SDK logs about but doesn't reject).

## Tests

- `test_first_event_is_task_for_new_request` — pins the new-request path: first enqueue must be a `Task` with the expected id/context_id
- `test_no_task_enqueue_on_continuation` — pins the continuation path: when `context.current_task` is set, the executor must NOT re-enqueue Task
- `conftest.py` extended with stubs for `Task` / `TaskStatus` / `TaskState` so the import inside the executor resolves under unit tests

All 45 tests pass (43 existing + 2 new).

## Scope notes

- google-adk adapter does not have this bug — its `execute()` only emits `Message` events. Its `cancel()` does emit one TaskStatusUpdateEvent, but cancel is rarely-invoked and out of scope here.
- Unit tests can't fully replicate the SDK's framework-level validation (mocks bypass it), so the synthetic E2E gate remains the live verification — next firing after merge → publish-runtime cascade → next synth-E2E should go green at step "8/11 Sending A2A message to parent — expecting agent response".

## Verification path

1. Merge → publish-runtime cascade kicks (workspace/** path-filter trigger)
2. Wheel uploads to PyPI → wheel smoke verifies imports
3. Template image rebuild on the 4 manifest entries (claude-code, hermes, openclaw, codex)
4. Next synth-E2E firing on staging (cron every 20 min, on :00 :20 :40) provisions a fresh langgraph workspace and exercises the A2A path

## Related

- Bug introduced by #2170 (a2a-sdk v0→v1 migration)
- Surfaced by the synth-E2E gate (#2342) doing exactly what it was built for
- Memory note will be updated to track v1-contract pattern alongside renamed-symbols list

### DIFF
--- a/workspace/a2a_executor.py
+++ b/workspace/a2a_executor.py
@@ -268,6 +268,26 @@ class LangGraphA2AExecutor(AgentExecutor):
             task_id = context.task_id or str(uuid.uuid4())
             context_id = context.context_id or str(uuid.uuid4())
 
+            # A2A v1 contract (a2a-sdk ≥ 1.0): enqueue a Task event before any
+            # TaskStatusUpdateEvent. The framework only auto-creates the Task
+            # on continuation messages (existing task_id resolves via
+            # task_manager.get_task()). For fresh requests get_task() returns
+            # None and the SDK rejects the first status update with
+            # InvalidAgentResponseError("Agent should enqueue Task before
+            # TaskStatusUpdateEvent event") — see a2a/server/agent_execution/
+            # active_task.py for the validation site. PR #2170 migrated the
+            # surface to v1 but missed this contract; the synth-E2E gate
+            # surfaced it on every run after staging deploy.
+            if getattr(context, "current_task", None) is None:
+                from a2a.types import Task, TaskState, TaskStatus
+                await event_queue.enqueue_event(
+                    Task(
+                        id=task_id,
+                        context_id=context_id,
+                        status=TaskStatus(state=TaskState.TASK_STATE_SUBMITTED),
+                    )
+                )
+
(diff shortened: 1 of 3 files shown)


hongmingwang-moleculeai (approved):
Bug class is well-understood (a2a-sdk v1 contract that #2170 missed). Fix is correctly gated on context.current_task to avoid double-enqueue on continuations. Tests pin both branches: new request → first event Task; continuation → no Task re-enqueue. Live verification via next synth-E2E firing after publish-runtime cascade.

PR #2604, opened 2026-05-03
fix(canvas): chat bubble + sub-tab contrast in light theme

HongmingWang-Rabbit:
## Summary

User reported pale-purple chat bubbles with washed-out text in canvas (screenshot). Root causes:

- **Chat user bubble** used `bg-accent-strong/30 text-blue-100` — translucent dark-blue overlay on warm-paper renders as pale lavender; light-blue text on top of that is borderline invisible. Switched to solid `bg-accent text-white`.
- **System/error bubble** had the same pattern with red. Migrated to `bg-bad/10 text-bad` so dark-mode tracks automatically.
- **Agent bubble** had `/80 + /30` alpha modifiers — solid `bg-surface-card text-ink border-line` gives WCAG AA contrast in both themes.
- **`prose-invert`** was unconditional. On agent/system bubbles in light mode the markdown body rendered as light text on a light surface. Now only applied on the user bubble (solid blue).
- **Timestamp** used `text-ink-soft` (3.5:1 on warm-paper). Bumped to `text-ink-mid` (7:1) for agent/system, `white/70` for user.

Sub-tab bar in `SidePanel.tsx`:
- Right-edge scroll-fade was hardcoded `from-zinc-950`. In light mode that paints a dark vertical strip — switched to `from-surface` so the gradient matches whichever theme is active.
- Inactive tab text was `text-ink-soft` (failing AA). Bumped to `text-ink-mid`.
- Active tab dropped the `/40` opacity on its bg-card so selection is unambiguous.

All semantic tokens already exist; no globals.css changes.

## Test plan
- [x] `pnpm build` clean
- [x] `npx tsc --noEmit` clean
- [ ] Browser-verify on staging tenant in both light + dark mode after deploy

### DIFF
--- a/canvas/src/components/SidePanel.tsx
+++ b/canvas/src/components/SidePanel.tsx
@@ -202,7 +202,7 @@ export function SidePanel() {
       {/* Tabs — relative wrapper lets the fade gradient position against the scroll container */}
       <div className="relative border-b border-line/40">
         {/* Right-edge fade: signals more tabs are hidden off-screen when the bar overflows */}
-        <div className="pointer-events-none absolute inset-y-0 right-0 w-8 bg-gradient-to-l from-zinc-950 to-transparent z-10" aria-hidden="true" />
+        <div className="pointer-events-none absolute inset-y-0 right-0 w-8 bg-gradient-to-l from-surface to-transparent z-10" aria-hidden="true" />
       <div
         role="tablist"
         aria-label="Workspace panel tabs"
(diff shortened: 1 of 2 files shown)


hongmingwang-moleculeai (approved):
Light-mode contrast fix on canvas chat. Translucent overlays + hardcoded blue/red text classes were the root cause; switching to solid accent + semantic tokens is the right call. prose-invert conditioning closes the markdown body bug. All tokens already exist.

PR #2629, opened 2026-05-03
feat(canvas): add Hermes/Codex/OpenClaw tabs to ExternalConnectModal + default to Universal MCP

HongmingWang-Rabbit:
## Summary

External Connect modal previously had tabs for Python SDK / curl / Claude Code channel / Universal MCP. Operators using hermes / codex / openclaw as their external runtime had no copy-paste — they pieced together WORKSPACE_ID + PLATFORM_URL + auth_token into config files by reading docs.

Adds three runtime-specific snippets stamped server-side, gated on the platform supplying the field (older platform builds without it don't surface empty tabs).

## Hermes tab

Installs `molecule-ai-workspace-runtime` + the `hermes-channel-molecule` plugin, exports the 4 env vars, writes the `gateway.plugin_platforms.molecule` block into `~/.hermes/config.yaml`. Same long-poll-based push semantics the Claude Code channel tab delivers (push parity with the in-tree `template-hermes` adapter).

## Codex tab

Wires the molecule_runtime A2A MCP server into `~/.codex/config.toml` (`[mcp_servers.molecule]` block with env_vars passthrough + literal env values). **Outbound-tools-only today** — codex's MCP client doesn't route arbitrary `notifications/*` (verified by reading `codex-rs/codex-mcp/src/connection_manager.rs`); push parity on external codex would need a separate bridge daemon (future work). Snippet calls this out so operators know to pair with Python SDK if they need inbound delivery.

## OpenClaw tab

Installs openclaw + onboards, wires the molecule MCP server via `openclaw mcp set`, starts the gateway on loopback. Same outbound-tools-only caveat as codex. The in-tree `template-openclaw` adapter implements the full `sessions.steer` push path, but an external setup would need the same bridge daemon to translate platform inbox events into `sessions.steer` calls (future work).

## Default tab change

Pre-2026-05-03 the modal defaulted to Claude Code. Operators using non-Claude runtimes opened to a tab they had to skip past. Changed to Universal MCP — runtime-agnostic, works as a starting point for any operator. Runtime-specific tabs are still one click away.

## Tab order

Universal MCP → Python SDK → Claude Code → Hermes → Codex → OpenClaw → curl → Fields

## Test plan

- [x] `go build ./internal/handlers/...` clean
- [x] Canvas TypeScript clean (`tsc --noEmit`)
- [ ] Manual: provision a `runtime=external` workspace on staging, verify modal shows all 6 runtime tabs + Universal MCP opens by default + token auto-stamping works on each new tab
- [ ] Each new tab's copy-paste runs end-to-end against a real workspace (paste, follow steps, verify the runtime can call `list_peers` / receive A2A)

### DIFF
--- a/canvas/src/components/ExternalConnectModal.tsx
+++ b/canvas/src/components/ExternalConnectModal.tsx
@@ -40,6 +40,22 @@ export interface ExternalConnectionInfo {
   // + inbound. Optional for backward compat with platforms that
   // haven't shipped PR #2413 yet.
   universal_mcp_snippet?: string;
+  // Hermes channel snippet — for operators whose external agent IS a
+  // hermes-agent session. Routes A2A traffic into the hermes gateway
+  // via the molecule-channel plugin (Molecule-AI/hermes-channel-molecule).
+  // Long-poll based (no tunnel) — same UX shape as the Claude Code
+  // channel tab. Gives hermes true push parity. Optional for backward
+  // compat with platforms that haven't shipped this PR yet.
+  hermes_channel_snippet?: string;
+  // Codex MCP config snippet — wires the molecule MCP server into
+  // ~/.codex/config.toml so codex agents can call platform tools.
+  // Outbound-tools-only today (codex's MCP client doesn't route
+  // notifications/*); push parity would need a separate bridge daemon.
+  codex_snippet?: string;
+  // OpenClaw MCP config snippet — wires molecule MCP + starts the
+  // openclaw gateway on loopback. Outbound-tools-only today; push
+  // parity on an external openclaw needs a sessions.steer bridge.
+  openclaw_snippet?: string;
 }
 
 interface Props {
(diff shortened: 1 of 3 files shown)


HongmingWang-Rabbit:
## Self-review (5-axis) — found 2 Critical + 2 Required + 2 Optional

Caught these reviewing my own PR before merge.

### Critical

**1. Hermes snippet — `cat >> ~/.hermes/config.yaml` corrupts existing configs.**

Most existing hermes installs have a top-level `gateway:` block. The current snippet appends a SECOND `gateway:` block. YAML rejects duplicate top-level keys → on next gateway restart, hermes-cli's config loader either errors out or silently uses one of the blocks (impl-defined). Footgun for any operator who's run hermes before on the target machine.

The in-tree codex_mcp_config.sh handles the analogous case by stripping the prior block before re-appending. The external snippet doesn't. Fix shape: provide an explicit "add under your existing gateway: block" instruction OR ship a Python one-liner that does proper YAML merge. Going with the explicit-instruction approach in the fix — auto-merge attempts on a YAML config are fragile and platform-side errors here would be invisible.

**2. Codex snippet — `cat >> ~/.codex/config.toml` corrupts on re-run.**

Same class. TOML rejects duplicate `[mcp_servers.molecule]` tables. The in-tree codex_mcp_config.sh strips before re-appending; this snippet doesn't. Operators re-running the snippet (token rotation, second workspace) will hit a parse error and codex won't start.

### Required

**3. OpenClaw `openclaw onboard --non-interactive` missing provider/model.**

The in-tree template-openclaw adapter passes `--provider`, `--model`, `--api-key`, etc. explicitly. Without them, `--non-interactive` either fails (no defaults) or applies a stub default the user doesn't want. Fix: either provide placeholders for those args OR remove the onboard step and tell the user to run it themselves first.

**4. OpenClaw gateway started with `&` dies when terminal closes.**

`openclaw gateway --dev --port 18789 --bind loopback &` is fine for a quick local test but the workspace stops accepting A2A as soon as the operator closes the terminal. Either `nohup ... &> ~/.openclaw/gateway.log &`, or split into "open this in a second terminal," or systemd. Current shape sets up a workspace that silently dies overnight.

### Optional

**5. Codex `env_vars = [\"MOLECULE_INBOUND_SECRET\", \"PYTHONPATH\"]`** is copy-pasted from the in-container template. Both values are container-side only — harmless on an external setup (env_vars only forwards env vars that exist) but adds noise to a snippet operators will read carefully. Consider stripping to `[]` or removing the field entirely.

**6. No tests added** for the 3 new response fields. Existing modal also has no tests, but a server-side test asserting the response shape would catch regressions when external_connection.go is next touched. Not blocking; filing as followup if not addressed in this PR.

### Verdict
**Request changes** — must address 1 + 2 before merge. 3 + 4 are required-by-merge but can be terse. Optional findings can be deferred.

Shipping fix commit shortly.

HongmingWang-Rabbit:
Pushed ff0d4da addressing all 4 required findings. Optional 5 + 6 deferred. Re-queuing auto-merge.

Issue #2632, opened 2026-05-03
test: Canvas tabs E2E intermittently fails on staging (~50% pass rate)


HongmingWang-Rabbit:
## Symptom

\`Canvas tabs E2E\` (e2e-staging-canvas.yml) is failing intermittently across recent PRs and direct staging pushes. Last 12 runs across all branches show ~50% failure rate.

Recent sample (2026-05-03):
| Time | SHA | Result |
|------|-----|--------|
| 20:33Z (×2) | f7d663d1 | both **fail** |
| 20:20Z | 6f203c56 | success |
| 20:17Z | 01bbf4c8 | success |
| 20:16Z | 1ec7e4af | success |
| 20:16Z | ae4739f3 | **fail** |
| 20:13Z | ff0d4dae | success |
| 20:12Z | e89dd892 | **fail** |

## Failure mode

Two distinct symptoms in the same test:

1. \`Error: provision failed: e2e-canvas-...\` — CP returns \`instance_status=failed\` during tenant create polling.
2. \`Error: Workspace failed: \` (empty \`last_sample_error\`) — tenant provisions OK, but workspace agent boot reports \`status=failed\` with no detail.

## Impact

- Auto-promote-staging gates on this workflow. When it fails, promote PRs (e.g. #2625) sit blocked until rerun, triggering a re-rebase loop with sibling PRs that dismisses approvals each time.
- Workaround: \`gh run rerun <id> --failed\` and re-approve the promote PR. Sometimes takes 2–3 attempts.

## Likely root causes (need investigation)

- **EC2 capacity / quota** during burst tenant creation
- **CP race** in the provisioning state machine on freshly-created tenants
- **OpenAI quota cascade** (#2578): if the staging key is exhausted, workspace boot may report empty \`last_sample_error\` instead of a clear "no provider key" error
- **Hermes cold-boot timeout** under load (the test uses default tier 1 / hermes runtime per workspace)

## Suggested next steps

1. Add provisioning retry-on-transient-failure in the test harness (1 retry on \`instance_status=failed\` before bailing)
2. Surface \`last_sample_error\` in the workspace failure path so we can stop guessing
3. Tag the test as \`flaky\` so it doesn't gate auto-promote — at the cost of weakened canary signal
4. Investigate whether OpenAI quota exhaustion (#2578) is causing the empty-error workspace boot failures

## Related

- #2578 — OpenAI quota exhausted (canary failures)
[...]

Issue #2578, opened 2026-05-03
Synth E2E (staging): MOLECULE_STAGING_OPENAI_KEY account is over quota — needs billing top-up


HongmingWang-Rabbit:
## Symptom

Continuous synth E2E (staging) has been red since 2026-05-01 21:29Z (>40h). Now that the upstream A2A/auth fixes (#2558, #2563, #2567, #2571) are in, the underlying root cause is visible:

```
Agent error: Error code: 429 - {'error': {
  'message': 'You exceeded your current quota, please check your plan and billing details.',
  'type': 'insufficient_quota',
  'code': 'insufficient_quota'
}}
```

(from synth E2E run 25279775715, model = `openai:gpt-4o`, key from repo secret `MOLECULE_STAGING_OPENAI_KEY`)

## Why it matters

- Synth E2E is the canary for staging health. With the API key over quota, **every** scheduled run fails — turning the canary into noise.
- Adjacent E2E suites that may also use this key (`e2e-staging-saas.yml`, `canary-staging.yml` per the comment in `.github/workflows/continuous-synth-e2e.yml:92-98`) are likely affected too.

## Action required

Operator action — not fixable in code:
1. Log into the OpenAI org that owns this key
2. Either top up the billing balance, increase the rate limit, or rotate the key to a higher-quota org
3. If rotating: update the `MOLECULE_STAGING_OPENAI_KEY` repo secret at Settings → Secrets and Variables → Actions

## Verify resolution

Manually dispatch one synth E2E run after the top-up:
```
gh workflow run continuous-synth-e2e.yml --repo Molecule-AI/molecule-monorepo --ref staging -f runtime=langgraph
```

Expect a green run within ~5 min — the workflow should reach step 11/11 with an actual agent text response.

## Related

- #2566 (closed) — the A2A/auth chain that masked this for 36+ hours
- PR #2571 — final code-side fix (model_slug colon-form for langgraph)

HongmingWang-Rabbit:
## Update — 2026-05-04: blast radius now consolidated

PR #2710 (merged 07:27Z) migrated `canary-staging.yml` and `e2e-staging-saas.yml` to the same `claude-code + MiniMax` path that `continuous-synth-e2e.yml` switched to on 2026-05-03. **All three** auto-running E2E gates now share one operator-action dependency instead of three independent ones.

### Verified after merge

Manual dispatch run #25306722939 (canary, ref=staging) fails fast at the new "Verify LLM key present" step:

[...]

PR #2634, opened 2026-05-03
canvas/e2e: surface admin-orgs row + workspace body on failure

HongmingWang-Rabbit:
## Summary

Two diagnostic upgrades to the Playwright staging-setup harness — zero behavior change.

1. **provision-failed throw** now includes the full admin-orgs row (boot stage, last error, terraform/SSM state) instead of just the slug. Every \`provision failed: <slug>\` in CI history was followed by a manual repro to find out WHY.

2. **workspace-failed throw** dumps the full \`/workspaces/{id}\` body when \`last_sample_error\` is empty. Boot crashes, image-pull errors, missing PYTHONPATH, and OpenAI-quota-at-startup all surface as a bare \`Workspace failed:\` today (see #2632). Now they carry whatever \`boot_stage\` / \`image\` / \`last_error\` fields the API row exposes.

## Why now

#2632 documents the flakiness. The underlying causes (#2632 CP race, #2578 OpenAI quota) need separate fixes — but right now both look identical in the CI log, which is the problem. After this lands, the next flake reveals its own subsystem.

## Test plan

- [x] \`tsc --noEmit\` clean
- [ ] Verify failure log on next E2E flake carries the admin-orgs row OR the workspace body

### DIFF
--- a/canvas/e2e/staging-setup.ts
+++ b/canvas/e2e/staging-setup.ts
@@ -169,7 +169,17 @@ export default async function globalSetup(_config: FullConfig): Promise<void> {
         orgID = row.id;
         return true;
       }
-      if (row.instance_status === "failed") throw new Error(`provision failed: ${slug}`);
+      if (row.instance_status === "failed") {
+        // Dump every diagnostic field the admin row carries — boot stage,
+        // last error, terraform/SSM state, etc. The bare slug message used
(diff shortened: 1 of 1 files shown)


hongmingwang-moleculeai (approved):
Pure observability win — strictly more info on failure, no behavior change. LGTM.

PR #2639, opened 2026-05-03
canvas/SearchDialog: auto-highlight first match + semantic placeholder

HongmingWang-Rabbit:
## Summary

Two small UIUX fixes for Cmd+K search.

- **Auto-highlight first match** while typing. Before, Enter on a non-empty query was a no-op (\`focusedIndex\` stayed at -1 until the user pressed ↓). Now the top match is pinned so Enter just works — standard search-palette UX. Empty query still keeps -1 so opening the dialog (showing ALL workspaces) doesn't arbitrarily pin one row.
- **\`placeholder-zinc-400\` → \`placeholder-ink-soft\`**. The hardcoded zinc broke the semantic-token pattern other inputs use; placeholder now flips with theme correctly.

## Test plan

- [x] \`tsc --noEmit\` clean
- [x] Full canvas suite: 1220/1220 (15 SearchDialog tests including 2 new)
- [ ] Cmd+K → type a workspace name → press Enter → that workspace selects (no ↓ needed)
- [ ] Cmd+K → no query → Enter → no-op (existing behavior)
- [ ] Toggle theme → placeholder color flips

### DIFF
--- a/canvas/src/components/SearchDialog.tsx
+++ b/canvas/src/components/SearchDialog.tsx
@@ -36,11 +36,6 @@ export function SearchDialog() {
     }
   }, [open]);
 
-  // Reset focused index when query changes
-  useEffect(() => {
-    setFocusedIndex(-1);
-  }, [query]);
-
   const filtered = nodes.filter((n) => {
     if (!query) return true;
     const q = query.toLowerCase();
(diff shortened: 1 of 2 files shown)


hongmingwang-moleculeai (approved):
Auto-highlight is the standard search-palette pattern; the empty-query carve-out is the right call. 2 new tests, 13 existing ones still cover the keyboard rollup. LGTM.

PR #2640, opened 2026-05-03
test(e2e): canary classifies provider-quota 429 as operator-action, not platform regression

HongmingWang-Rabbit:
## Summary

The staging canary's A2A step has a ladder of specific regression classifiers (\`hermes-agent error 401\`, \`model_not_found\`, \`Encrypted content is not supported\`, \`Unknown provider\`, \`hermes-agent unreachable\`, \`Invalid API key\`) followed by a generic \`error|exception\` catch-all. **Provider-side OpenAI 429 quota errors fall through to the catch-all** — the canary issue body and CI log just say \"A2A returned an error-shaped response,\" which is true but obscures the actual operator action.

This adds a 7th classifier above the catch-all that matches \`exceeded your current quota\` and \`insufficient_quota\` (both terms appear in OpenAI's quota-exhaustion 429). When matched, the failure names the operator action directly:

> A2A — PROVIDER QUOTA EXHAUSTED (NOT a platform regression). Operator action: top up \`MOLECULE_STAGING_OPENAI_KEY\` billing or rotate to a higher-quota org at Settings → Secrets and Variables → Actions. Tracked in #2578.

## Why this is correct, not \"lowering the bar\"

- Steps 0–7 cover **full platform health**: CP reachable, tenant provisioned, DNS+TLS up, workspace booted, A2A delivered end-to-end.
- Reaching step 8 with a provider-side 429 means the platform **is** healthy — the failure is downstream of every platform invariant.
- The canary still **exits 1** (CI stays red, threshold-3 alarm still fires); **only the failure message changes**.
- All 6 existing specific classifiers run BEFORE the new one, so any real platform regression is still caught with its specific message.

## Verification

- Regex tested against the actual 429 string from canary run 25291517608:
  > \`API call failed after 3 retries: HTTP 429: You exceeded your current quota, please check your plan and billing details.\`
  → matches ✅
- Negative tests: \`PONG\` (benign reply), \`hermes-agent unreachable\` (a real platform regression that has its own classifier) → no match ✅
- \`bash -n\` syntax check clean
- \`shellcheck -S error\` clean

## Tracking

- #2593 — \"🔴 Canary failing: staging SaaS smoke\" (the alert)
- #2578 — \"Synth E2E (staging): MOLECULE_STAGING_OPENAI_KEY account is over quota — needs billing top-up\" (root cause)

Both stay open until the operator tops up the key — this PR doesn't close either.

### DIFF
--- a/tests/e2e/test_staging_full_saas.sh
+++ b/tests/e2e/test_staging_full_saas.sh
@@ -517,6 +517,7 @@ fi
 #   "Encrypted content is not supported" → hermes codex_responses API misroute (#14)
 #   "Unknown provider"               → bridge misconfigured PROVIDER= (regression of #13 fix)
 #   "hermes-agent unreachable"       → gateway process died
+#   "exceeded your current quota"    → MOLECULE_STAGING_OPENAI_KEY billing (NOT a platform regression — #2578)
 #
 # Fail LOUD with the specific pattern so CI log + alert channel makes the
 # regression unambiguous.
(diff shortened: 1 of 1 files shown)


hongmingwang-moleculeai (approved):
Approved — additive classifier above the existing catch-all, regex verified against real 429 string, no behavior change to alarm logic.

PR #2644, opened 2026-05-03
canvas/CookieConsent: stop claiming aria-modal without a focus trap

HongmingWang-Rabbit:
## Summary

Three a11y fixes for the cookie banner.

- **role=\"dialog\" aria-modal=\"true\" → \`<section role=\"region\">\`**. The banner doesn't trap focus, doesn't block the page, and the user can keep using the canvas while it's up — none of which are modal semantics. Claiming \`aria-modal=true\` without a trap actively harms screen-reader users: they're told the rest of the page is inert, jump in, and then can't escape. Region semantics let AT navigate around it normally. Forcing a modal cookie banner would also be a dark pattern under GDPR.
- **Privacy link**: \`hover:text-accent\` (no-op, same color) → \`hover:text-accent-strong\` + focus-visible ring + \`underline-offset-2\`.
- **Both buttons**: \`focus-visible:ring-2 ring-offset-surface\` so keyboard users see focus.

## Test plan

- [x] \`tsc --noEmit\` clean
- [x] Full canvas suite: 1220/1220 (14 CookieConsent tests, including a new regression guard that explicitly asserts NO \`aria-modal\` and NO dialog role)
- [ ] On a fresh SaaS hostname (no localStorage entry): banner appears at bottom; Tab can leave it; screen reader navigates around it
- [ ] Both buttons + link visibly ring on Tab focus

### DIFF
--- a/canvas/src/components/CookieConsent.tsx
+++ b/canvas/src/components/CookieConsent.tsx
@@ -98,9 +98,17 @@ export function CookieConsent() {
   };
 
   return (
-    <div
-      role="dialog"
-      aria-modal="true"
+    // role="region" + aria-label, NOT role="dialog" + aria-modal. The
+    // banner is informational — it never blocks the page, never traps
+    // focus, and the user can keep using the canvas while it's up.
+    // Claiming aria-modal="true" without a focus trap is genuinely
+    // harmful for screen-reader users: they get told the rest of the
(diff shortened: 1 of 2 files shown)


hongmingwang-moleculeai (approved):
Removing the false aria-modal claim is the right call — region semantics + the new explicit guard test prevent regression. Link/button focus rings round it out. LGTM.

PR #2647, opened 2026-05-03
canvas/Legend: focus rings + 24x24 close-button touch target

HongmingWang-Rabbit:
## Summary

Two small a11y fixes for the floating legend.

- **Focus-visible ring** on both buttons (open pill + close ×). Keyboard users had no way to see where focus landed.
- **Close button hit area**: was ~10×16px (well under WCAG 2.5.5's 24×24 minimum). Bumped to \`w-6 h-6\` with negative margin so the visible × stays put but the hit area + focus ring are larger; hover bg makes the area visible.

## Test plan

- [x] \`tsc --noEmit\` clean
- [x] Full canvas suite: 1220/1220
- [ ] Tab into the canvas → reach the Legend pill (when collapsed) and the × (when expanded) — accent ring visible

### DIFF
--- a/canvas/src/components/Legend.tsx
+++ b/canvas/src/components/Legend.tsx
@@ -77,7 +77,7 @@ export function Legend() {
         onClick={openLegend}
         aria-label="Show legend"
         title="Show legend"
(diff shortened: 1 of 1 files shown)


hongmingwang-moleculeai (approved):
Same accent-ring pattern as the rest; the negative-margin hit-area-bump trick keeps the visual position fixed. LGTM.

PR #2650, opened 2026-05-03
ci: port DELETE-verify pattern to remaining staging e2e workflows

HongmingWang-Rabbit:
## Summary

Follow-up to #2648. Same \`>/dev/null || true\` swallow-on-error pattern existed in 4 other staging e2e workflows. Ports the verification fix mechanically:

- \`e2e-staging-canvas.yml\` (single-slug)
- \`e2e-staging-saas.yml\` (loop)
- \`e2e-staging-sanity.yml\` (loop)
- \`e2e-staging-external.yml\` (loop, was \`>/dev/null 2>&1\` variant)

Each now captures the HTTP code, logs success per-slug, and emits a workflow warning on non-2xx naming the slug + body excerpt. Loop bodies tally + summarise total leaks at the end.

## Why this is non-fail-flag

Exit semantics unchanged: a single cleanup miss still doesn't fail-flag the test (sweep-stale-e2e-orgs is the safety net within ~45 min). The behavior change is purely surfacing — failures that were silent are now visible on the workflow run page.

## Pairs with #2648

#2648 tightened the sweep cadence (15 min) + threshold (30 min). This PR makes per-run cleanup failures visible. Together: the worst-case leak window is ~45 min AND we know which workflow leaked when it happens.

## Test plan

- [x] \`python3 yaml.safe_load\` clean on all four files.
- [x] No exit-code change — \`exit 0\` preserved across the board.
- [x] Same shape as the canary teardown landed in #2648 (just adapted for single-slug vs loop bodies).

### DIFF
--- a/.github/workflows/e2e-staging-canvas.yml
+++ b/.github/workflows/e2e-staging-canvas.yml
@@ -184,8 +184,23 @@ jobs:
             exit 0
           fi
           echo "Deleting orphan tenant: $slug"
-          curl -sS -X DELETE "$MOLECULE_CP_URL/cp/admin/tenants/$slug" \
+          # Verify HTTP 2xx instead of `>/dev/null || true` swallowing
+          # failures. A 5xx or timeout previously looked identical to
+          # success, leaving the tenant alive for up to ~45 min until
+          # sweep-stale-e2e-orgs caught it. Surface failures as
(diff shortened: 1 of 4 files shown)


hongmingwang-moleculeai (approved):
Self-review: mechanical port of #2648 pattern, YAML validated, exit semantics preserved.

PR #2705, opened 2026-05-04
fix(canvas): A2ATopologyOverlay re-fetch storm hammering /activity → 429

HongmingWang-Rabbit:
## Summary

User reported (browser console screenshot): canvas hammering \`/workspaces/<id>/activity?type=delegation&limit=500&source=agent\` with 429s repeating across every visible workspace ID and an infinite \`uE → ux → uE → ux\` render trace.

Root cause: selector instability. \`A2ATopologyOverlay\` selected the entire \`nodes\` array from Zustand, then derived \`visibleIds\` via useMemo on \`[nodes]\`. The store mutates \`nodes\` reference on every update (status flips, position drags, peer-discovery writes, heartbeats — typically ~5/sec) even when the visible-ID set is unchanged. Each mutation:
1. \`visibleIds\` recreates (new array reference)
2. \`fetchAndUpdate\` (useCallback'd on \`[visibleIds]\`) recreates
3. The useEffect (deps include \`fetchAndUpdate\`) re-fires
4. Tears down the 60s setInterval, fires immediate fan-out fetch, sets new interval
5. Result: ~5×N requests/sec where N = visible workspaces → 429

## Fix

Select a **stable string key** (sorted CSV of visible IDs) from Zustand. The selector's shallow-equal short-circuit prevents re-renders when the actual visible-ID set is unchanged. \`visibleIds\` reference stays stable → \`fetchAndUpdate\` keeps identity → useEffect only re-fires when the visible-ID-set genuinely changes.

\`\`\`tsx
// Before
const nodes = useCanvasStore((s) => s.nodes);
const visibleIds = useMemo(() => nodes.filter(...).map(...), [nodes]);
// nodes reference recreates → visibleIds recreates → re-fetch storm

// After
const visibleIdsKey = useCanvasStore((s) =>
  s.nodes.filter((n) => !n.hidden).map((n) => n.id).sort().join(",")
);
const visibleIds = useMemo(
  () => visibleIdsKey ? visibleIdsKey.split(",") : [],
  [visibleIdsKey]
);
\`\`\`

## Test plan

- [x] New regression test: \`does not re-fetch when nodes[] reference changes but visible IDs are the same\`
- [x] Discipline check: reverted to broken pre-fix code → test fails (got 4 fetches, expected 2)
- [x] Restored fix → all 27 tests pass
- [x] Companion test \`re-fetches when the visible ID set actually changes\` pins the desired behavior so a future "stabilization" doesn't suppress legitimate updates

## Operator action

None — once this lands, refresh the canvas tab and the request flood stops.

### DIFF
--- a/canvas/src/components/A2ATopologyOverlay.tsx
+++ b/canvas/src/components/A2ATopologyOverlay.tsx
@@ -138,14 +138,37 @@ export function A2ATopologyOverlay() {
   // Stable Zustand action reference — safe to call inside effects
   const setA2AEdges = useCanvasStore((s) => s.setA2AEdges);
 
-  // Read the nodes array as a primitive ref; derive visible IDs outside the selector
-  const nodes = useCanvasStore((s) => s.nodes);
+  // Subscribe to a STABLE STRING KEY of visible workspace IDs, not the
+  // nodes array itself. Zustand returns a new array reference on every
+  // store update (status flips, position drags, peer-discovery writes,
+  // workspace-tab opens, etc.) — even when the set of visible IDs is
+  // unchanged. Selecting a sorted-CSV string makes Zustand's default
+  // shallow-equal short-circuit the re-render unless the actual ID set
+  // changes.
+  //
+  // Why this matters: previously visibleIds was useMemo'd on `nodes`, so
+  // the array reference recreated on every store mutation. fetchAndUpdate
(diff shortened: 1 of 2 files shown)


hongmingwang-moleculeai (approved):
approve: render-loop fix, regression test discipline-verified

PR #2804, opened 2026-05-05
fix(external-templates): drop MOLECULE_ORG_ID from codex/openclaw/hermes snippets

HongmingWang-Rabbit:
## Summary

- The codex / openclaw / hermes-channel external-tab snippets shipped with `MOLECULE_ORG_ID = \"<your org id>\"` as a required-looking env var. The molecule_runtime MCP subprocess that those snippets spawn **never reads** `MOLECULE_ORG_ID` — it's a workspace-server-side TenantGuard env (set by the control plane via user-data on the tenant box). External calls pass TenantGuard via the `isSameOriginCanvas` path (Origin matches Host) plus Bearer token + `X-Workspace-ID`.
- The reference is the universal_mcp snippet, which calls into the same molecule_runtime and has always (correctly) omitted `MOLECULE_ORG_ID`. This PR brings codex / openclaw / hermes-channel into line.
- Pinned with a structural test (`TestExternalTemplates_NoMoleculeOrgIDPlaceholder`) so the placeholder can't drift back into any of the six external templates.

## Symptom that caught it

An external codex CLI session, after pasting the codex tab snippet, flagged `MOLECULE_ORG_ID is still set to \"<your org id>\"` as an unresolved setup blocker. Agent did exactly the right thing — operator just had no value to fill, because the platform doesn't surface one and it isn't actually needed.

## Test plan

- [x] `go test ./internal/handlers/ -run TestExternalTemplates_NoMoleculeOrgIDPlaceholder -v` — PASS on fix
- [x] Verified the test FAILS on every pre-fix template (stash + re-run reproduced 3 expected failures: hermes / codex / openclaw — exactly the templates touched here)
- [x] `go test ./internal/handlers/` full suite — green
- [ ] After merge, click through the canvas External Connect modal codex tab to confirm the rendered snippet no longer contains the org id line


hongmingwang-moleculeai (approved):
Audited — operator-facing placeholder is dead-weight, structural test verified to fail on each pre-fix template. LGTM.

PR #2811, opened 2026-05-05
fix(provision): consolidate org-import gate + Auto self-marks-failed

HongmingWang-Rabbit:
## Summary

Closes the silent-drop bug class around workspace provisioning by making `provisionWorkspaceAuto` the **single source of truth** — both for backend routing AND for the no-backend failure path.

## The bug

Repro 2026-05-05 01:14 (production hongming tenant, prod):

```
01:14:10 org_import: provision concurrency cap=3
01:14:22 Org import: UX A/B Lab — 7 workspaces created
01:24:31 Provision-timeout sweep: ... — marked failed (×7)
```

7 claude-code workspaces sat in `provisioning` for 10 minutes — no provision goroutine ever fired (zero `CPProvisioner: goroutine entered` log lines). The sweeper marked them failed with the misleading `"container started but never called /registry/register"`.

PR #2798 was supposed to fix this exact bug class by routing org-import through `provisionWorkspaceAuto`. The fix was in the binary (verified via `strings /platform`). But it wasn't reachable on SaaS.

## Root cause

`org_import.go:178`:

```go
} else if h.provisioner != nil {   // ← Docker-only check
    // ... 220 lines of template/secret/plugin prep ...
    h.workspace.provisionWorkspaceAuto(...)  // ← line 395, PR #2798's fix
}
```

On SaaS, `h.provisioner` (Docker) is nil and `h.cpProv` (CP) is set. The outer gate evaluates false, the entire prep block is skipped — including the Auto call. Workspace row created, no goroutine, sweep marks failed.

`Create()` (single-workspace path) didn't have this gate, which is why **single-workspace creation worked but org-template didn't** even though they should be identical.

## Fix

**1. Replace the bare-Docker gate with a symmetric check:**

```go
} else if h.workspace.HasProvisioner() {  // either backend counts
```

New `WorkspaceHandler.HasProvisioner()` returns `cpProv != nil || provisioner != nil`.

**2. Make `provisionWorkspaceAuto` self-mark-failed when no backend wired** (defense in depth):

```go
func (h *WorkspaceHandler) provisionWorkspaceAuto(...) bool {
    if h.cpProv != nil { go h.provisionWorkspaceCP(...); return true }
    if h.provisioner != nil { go h.provisionWorkspace(...); return true }
    // No backend — mark failed instead of returning silently. Even if a
    // future caller bypasses HasProvisioner gating or ignores the bool
    // return, the workspace ends in a clean failed state rather than
    // sitting stuck until the 10-min sweeper fires.
    h.markProvisionFailed(...)
    return false
}
```

This collapses Create's 16-line `if !Auto { ...mark-failed... }` block — Auto handles it now. Create keeps only its UI-specific `workspace_config` UPSERT (so the Config tab renders runtime/model even on failed workspaces).

## Why this is the right shape

- **Single source of truth.** Auto owns both backend routing AND the no-backend failure path. No caller has to reason about either.
- **Defense in depth.** Two independent layers prevent silent-drop: the `HasProvisioner` gate skips wasted prep work, and Auto's self-mark catches anything that gets past the gate.
- **Symmetric to `Create`'s existing shape.** Single-workspace and bulk-import now go through the same dispatcher with the same guarantees.

## Tests

- `TestProvisionWorkspaceAuto_NoBackendMarksFailed` — pins new contract (Auto fires `markProvisionFailed`)
- `TestHasProvisioner_TrueOnCPOnly` — catches the SaaS-only blind spot that broke org-import
- `TestHasProvisioner_TrueOnDockerOnly` — preserves self-hosted shape
- `TestHasProvisioner_FalseWhenNeitherWired` — pins the skip-prep path
- `TestOrgImportGate_UsesHasProvisionerNotBareField` — source-level pin against the buggy `h.provisioner != nil` shape

Verified: the source-level test FAILS against the buggy gate shape, PASSES with the fix.

## Test plan

- [x] `go build ./...` clean
- [x] `go vet ./...` clean
- [x] All handler tests pass (`go test ./internal/handlers/...`)
- [x] New test FAILS against the buggy gate shape (verified by reverting just the gate change)
- [ ] Staging deploy → org-import on a SaaS tenant produces 7 EC2s (currently broken in production)
- [ ] Watch for `CPProvisioner: goroutine entered` log lines on the tenant CP after deploy

### DIFF
--- a/workspace-server/internal/handlers/a2a_proxy_helpers.go
+++ b/workspace-server/internal/handlers/a2a_proxy_helpers.go
@@ -163,7 +163,7 @@ func (h *WorkspaceHandler) maybeMarkContainerDead(ctx context.Context, workspace
 	if wsRuntime == "external" {
 		return false
 	}
-	if h.provisioner == nil && h.cpProv == nil {
+	if !h.HasProvisioner() {
 		return false
 	}
 
--- a/workspace-server/internal/handlers/org_import.go
+++ b/workspace-server/internal/handlers/org_import.go
@@ -175,8 +175,18 @@ func (h *OrgHandler) createWorkspaceTree(ws OrgWorkspace, parentID *string, absX
 		h.broadcaster.RecordAndBroadcast(ctx, "WORKSPACE_ONLINE", id, map[string]interface{}{
 			"name": ws.Name, "external": true,
 		})
-	} else if h.provisioner != nil {
-		// Provision container
+	} else if h.workspace.HasProvisioner() {
+		// Provision container — either backend (CP for SaaS, local Docker
+		// for self-hosted) is fine. Pre-2026-05-05 this gate was
+		// `h.provisioner != nil`, which only checked the Docker pointer
+		// and silently dropped every workspace on a SaaS tenant: the prep
+		// block was skipped, no Auto call ever fired, and the row sat in
+		// 'provisioning' until the 600s sweeper marked it failed with the
+		// misleading "container started but never called /registry/register"
+		// (incident: hongming tenant org-import 2026-05-05 01:14, 7-of-7
+		// claude-code workspaces stuck). Routing to the right backend
+		// happens inside provisionWorkspaceAuto — this gate just decides
+		// whether to do prep at all.
 		payload := models.CreateWorkspacePayload{
 			Name: ws.Name, Tier: tier, Runtime: runtime, Model: model,
 			WorkspaceDir:    ws.WorkspaceDir,
--- a/workspace-server/internal/handlers/workspace.go
+++ b/workspace-server/internal/handlers/workspace.go
@@ -112,21 +112,43 @@ func (h *WorkspaceHandler) SetCPProvisioner(cp provisioner.CPProvisionerAPI) {
 	h.cpProv = cp
 }
 
+// HasProvisioner reports whether either backend (CP or local Docker) is
+// wired. Callers that gate prep-work on "do we have something that can
+// provision a container?" should use this rather than direct field access
+// to either provisioner — those individual checks miss the SaaS path
+// (cpProv set, provisioner nil) or the self-hosted path (provisioner set,
+// cpProv nil) symmetrically. Org-import + future bulk paths gate their
+// template/config/secret prep on this so the work isn't wasted on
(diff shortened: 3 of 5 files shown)


HongmingWang-Rabbit:
Multi-model second-pass review (independent agent, no priors) flagged two issues my first review missed:

**Required (test gap):** `TestProvisionWorkspaceAuto_NoBackendMarksFailed` asserts the UPDATE fires via sqlmock but **does not assert the broadcast fires**. The new code path inside Auto explicitly broadcasts `WORKSPACE_PROVISION_FAILED`, but only the UPDATE is pinned. Without a broadcast assertion, a future refactor that drops the broadcast inside `markProvisionFailed` ships green. `concurrentSafeBroadcaster` already tracks calls — adding `bcast.events()` (or equivalent) check + assert seen `WORKSPACE_PROVISION_FAILED` for the workspace_id would close this. Filing follow-up.

**Required (architectural framing — not a bug, doc clarity):** The gate `else if h.workspace.HasProvisioner()` in `org_import.go` skips the prep block AND Auto's call entirely on no-backend deployments. Auto's "self-marks-failed defense-in-depth" therefore can't help org-import in that scenario — the workspace still sits in 'provisioning' until the sweeper. The behavior is correct (org-import IS SaaS-only in practice; no-backend deployments don't org-import) but the comment in workspace.go's Auto docstring overstates the safety net for this caller. Worth either:
  (a) Tightening the docstring to say "Auto's self-mark covers callers that DO call Auto on no-backend (Create); callers that gate on HasProvisioner first (org-import) intentionally skip both prep and the safety net," OR
  (b) Removing the org-import gate and letting Auto handle the no-backend mark (wastes prep work in the rare no-backend case but eliminates the 10-min-stuck class entirely).

**FYI:** `markProvisionFailed` runs on `context.Background()` while Create's UPSERT uses request ctx — on client cancel mid-request, the workspace gets marked failed but workspace_config never persists. Pre-existing split, slightly widened by this PR. Worth a follow-up if Config-tab-on-failed-workspace matters.

Not blocking the merge — the primary SaaS gate fix is correct and ships value. Will file the broadcast-assert as a follow-up issue.

PR #2822, opened 2026-05-05
fix(files-eic): silence ssh known-hosts warning that 500'd Hermes config load

HongmingWang-Rabbit:
## User-reported bug
\`GET /workspaces/cfc1fd02.../files/config.yaml\` on hongming.moleculesai.app's Hermes workspace returned **500** with body:
\`\`\`
ssh cat: exit status 1 (Warning: Permanently added '[127.0.0.1]:37951' (ED25519) to the list of known hosts.)
\`\`\`

## Root cause
ssh emits the "Permanently added" notice on every fresh tunnel connection, even with \`UserKnownHostsFile=/dev/null\` (that prevents persistence, not the warning). It lands on stderr, fooling \`readFileViaEIC\`'s file-not-found classifier:
\`\`\`go
if len(out) == 0 && stderr.Len() == 0 {
    return nil, os.ErrNotExist  // → 404
}
return nil, fmt.Errorf("ssh cat: %w (%s)", runErr, ...)  // → 500
\`\`\`
stderr was non-empty (the warning), so we returned the wrapped error → 500 from the HTTP layer instead of 404.

## Fix
Add \`-o LogLevel=ERROR\` to BOTH \`writeFileViaEIC\` and \`readFileViaEIC\` ssh invocations. Silences info+warning while keeping real auth/tunnel errors visible (those emit at ERROR level).

## Test
\`TestSSHArgs_LogLevelErrorBothSites\` pins the flag in both blocks via source regex match. **Mutation-tested:** stripping the flag from the source fails the gate.

Sibling to my earlier #2785 (the ReadFile via SSH-EIC PR that introduced the classifier).

### DIFF
--- a/workspace-server/internal/handlers/template_files_eic.go
+++ b/workspace-server/internal/handlers/template_files_eic.go
@@ -178,6 +178,16 @@ func writeFileViaEIC(ctx context.Context, instanceID, runtime, relPath string, c
 		"-i", keyPath,
 		"-o", "StrictHostKeyChecking=no",
 		"-o", "UserKnownHostsFile=/dev/null",
+		// LogLevel=ERROR silences the benign "Warning: Permanently
+		// added '[127.0.0.1]:NNNNN' to known hosts" notice that ssh
+		// emits on every fresh tunnel connection. Without this, the
+		// notice lands on stderr and fools readFileViaEIC's "empty
+		// stdout + empty stderr → file not found" classifier into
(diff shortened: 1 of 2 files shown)


hongmingwang-moleculeai (approved):
User-reported live bug. Root cause clear (ssh known-hosts warning fooling stderr-empty classifier), fix scoped (one ssh flag in both blocks), pinned with mutation-tested AST gate. Ship it.

PR #2826, opened 2026-05-05
feat(canvas/chat): lazy-load history — 10 newest on mount, 20 per scroll-up batch

HongmingWang-Rabbit:
## Summary

ChatTab now lazy-loads message history instead of fetching the newest 50 in one shot.

- **Initial load:** newest 10 messages.
- **Scroll up:** IntersectionObserver on a top sentinel (rootMargin 200px) fires \`loadOlder()\` which fetches the next 20 older via the existing \`/activity?before_ts=<ts>&limit=20\` cursor.
- **\`hasMore\`** turns false when the server returns < limit rows; sentinel unmounts, IO disconnects.
- **Scroll-anchor preservation:** \`scrollAnchorRef\` captures distance-from-bottom before the prepend; \`useLayoutEffect\` restores it after, so the user's reading position doesn't jump when older messages land.
- **Live append paths unchanged:** new agent push / user-send still scrolls to bottom (same effect, gated on \`scrollAnchorRef === null\`).

## Why
Pre-fix every chat-tab open paid 50× message-bubble paint + DOM cost up-front. On a workspace with hundreds of messages this shows up as a measurable jank on tab swap. User-visible request.

## Test plan
- [x] 4 new vitest cases in \`ChatTab.lazyHistory.test.tsx\`:
  - initial fetch carries \`limit=10\` (not the legacy 50)
  - hasMore=false hides the top sentinel when initial fetch returns fewer than the limit
  - exactly-the-limit page renders all 10 row-pairs (sentinel mount path doesn't crash)
  - retry button after failed initial load uses \`limit=10\`
- [x] Build + full vitest pass: 1254 tests / 88 files.
- [ ] Browser verify: tab into Chat on a workspace with > 30 messages; confirm initial load is the newest 10; scroll up → \"Loading older messages…\" + 20 prepend; viewport position preserved; live agent message still scrolls to bottom.
- [ ] Browser verify: tab into Chat on a workspace with < 10 messages; sentinel doesn't mount; no spurious fetch.

### DIFF
--- a/canvas/src/components/tabs/ChatTab.tsx
+++ b/canvas/src/components/tabs/ChatTab.tsx
@@ -1,6 +1,6 @@
 "use client";
 
-import { useState, useRef, useEffect, useCallback } from "react";
+import { useState, useRef, useEffect, useCallback, useLayoutEffect } from "react";
 import ReactMarkdown from "react-markdown";
 import remarkGfm from "remark-gfm";
 import { api } from "@/lib/api";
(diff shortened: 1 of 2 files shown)


hongmingwang-moleculeai (approved):
ChatTab lazy-load history. Initial 10 + 20-per-batch via existing /activity before_ts cursor. Top sentinel + IO observer (rootMargin 200px) wires loadOlder. scrollAnchorRef + useLayoutEffect preserves the user's reading position on prepend so they don't jump. hasMore tear-down disconnects IO when no more history. 4 new vitest cases pin the limit=10 initial + retry, plus the hasMore=false sentinel-hidden path. Reduced-motion / live-append paths unchanged. Ship it.

PR #2827, opened 2026-05-05
ci: e2e coverage matrix + branch-protection-as-code (closes #9)

HongmingWang-Rabbit:
Closes #9 — E2E coverage audit + hard-gate CI for E2E suite.

### 1. \`docs/e2e-coverage.md\` — source of truth

Matrix of every E2E suite, what it covers, when it runs, and whether it's required on staging/main. Today three suites are running but informational only on staging — that's how the org-import silent-drop bug shipped without a test catching it pre-merge.

### 2. \`tools/branch-protection/apply.sh\` — protection as code

Branch protection lives in repo settings, so changes are usually clicked through the UI and lost between admins. This script makes the config reproducible — diff it, change the file, rerun.

**This PR's net change**: add \`E2E API Smoke Test\` and \`Canvas tabs E2E\` as required on staging. Both already use the always-emit path-filter pattern (no-op step emits SUCCESS when paths weren't touched), so making them required can't deadlock unrelated PRs.

### 3. \`branch-protection-drift.yml\` — daily drift check

Cron at 14:00 UTC + on-PR-touch. Compares live state against \`apply.sh\`'s desired state. Fails the workflow when they drift; ops re-runs the script or updates it. Single source of truth.

## Why this matters now

The 2026-05-05 hongming org-import silent-drop incident (PR #2811 fixed) shipped because the E2E SaaS workflow that would have caught it was *running* on every staging push but *not required*. A passing org-import via canvas was the closest pre-merge signal we had — the workspace-server unit tests didn't exercise the OrgHandler gate code path that contained the bug.

Adding the existing well-shaped E2Es as required is the cheapest first step. The bigger E2E SaaS suite still needs refactoring to the always-emit pattern before it can be required (filed as follow-up — its current \`paths:\` filter would deadlock unrelated PRs).

## Out of scope (follow-ups)

- \`e2e-staging-saas\` + \`e2e-staging-external\` need always-emit refactor before they can be required.
- \`main\` branch protection mirrors staging here. If we want \`E2E Staging SaaS\` required on main only after canary-verify, that's a separate gating layer.

## Operator action after merge

\`\`\`
bash tools/branch-protection/apply.sh
\`\`\`

Drift check runs daily after that — first failure means someone clicked through the UI; investigate and either re-apply or update apply.sh.

## Test plan

- [x] Shellcheck clean
- [x] Dry-run produces expected payload
- [x] yaml validates
- [ ] Operator applies post-merge and confirms drift check passes
- [ ] Next PR after this lands sees the new required checks (proves the gate works)

### DIFF
--- a/.github/workflows/branch-protection-drift.yml
+++ b/.github/workflows/branch-protection-drift.yml
@@ -1,0 +1,42 @@
+name: branch-protection drift check
+
+# Catches out-of-band edits to branch protection (UI clicks, manual gh
+# api PATCH from a one-off ops session) by comparing live state against
+# tools/branch-protection/apply.sh's desired state every day. Fails the
+# workflow when they drift; the failure is the signal.
+#
+# When it fails: re-run apply.sh to put the live state back to the
+# script's intent, OR update apply.sh to encode the new intent and
+# commit. Either way the script is the source of truth.
+
+on:
+  schedule:
+    # 14:00 UTC daily. Off-hours for most teams; gives a fresh signal
+    # at the start of every working day.
+    - cron: '0 14 * * *'
+  workflow_dispatch:
+  pull_request:
+    branches: [staging, main]
+    paths:
+      - 'tools/branch-protection/**'
+      - '.github/workflows/branch-protection-drift.yml'
+
+permissions:
+  contents: read
+
+jobs:
+  drift:
+    name: Branch protection drift
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+      - name: Run drift check
+        env:
+          # GH_TOKEN_FOR_ADMIN_API — repo-admin scope, needed for the
+          # /branches/:b/protection endpoint. Falls back to GITHUB_TOKEN
+          # for read-only mode (gh api on the protection endpoint
+          # returns 200 to maintainers, 404 to read-only). Configure at
+          # Settings → Secrets and variables → Actions.
+          GH_TOKEN: ${{ secrets.GH_TOKEN_FOR_ADMIN_API || secrets.GITHUB_TOKEN }}
+        run: bash tools/branch-protection/drift_check.sh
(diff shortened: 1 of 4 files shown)


HongmingWang-Rabbit:
## ⚠️ Multi-model review found Critical — disabled auto-merge

Independent review (Model B) caught what my self-approval missed: `apply.sh` was authored from scratch, not generated from live state. Running it post-merge would **silently weaken staging+main protection**:

| Setting | Live | Script writes | Impact |
|---|---|---|---|
| `enforce_admins` | true | false | Admin enforcement disabled |
| `dismiss_stale_reviews` | false | true | Per memory `feedback_dismiss_stale_reviews_blocks_promote.md` — silently re-blocks every auto-promote PR (cost the user 2.5h once already) |
| `strict` | true | false | Drops require-up-to-date |
| `allow_fork_syncing` | false | true | Allowed |
| `bypass_pull_request_allowances` | HongmingWang-Rabbit + molecule-ai app | wiped | Bypass list nuked silently |

### Other findings worth addressing in this PR

**Required**
- **R1** drift_check.sh only diffs `required_status_checks.checks` — won't catch C1's silent rewrite of admin/review settings. Drift gate is checks-only despite "single source of truth" claim.
- **R2** Schedule-vs-dispatch hardening missing per `feedback_schedule_vs_dispatch_secrets_hardening.md`. Scheduled run with missing `GH_TOKEN_FOR_ADMIN_API` falls back to `GITHUB_TOKEN` → 404 → cryptic gh-error message instead of "missing admin secret".
- **R3** Check-name existence not validated. Typo ("Canvas Tabs E2E" vs "Canvas tabs E2E") → GH accepts → every PR blocked forever waiting for a never-emitted context. Add preflight against `/commits/{sha}/check-runs`.
- **R4** drift_check.sh `sed` extraction is fragile against future apply.sh format changes.

**Optional** — auto-apply on merge (currently manual, which is good safety until C1 fixed).

### What to do
1. Regenerate apply.sh payload from `gh api repos/Molecule-AI/molecule-core/branches/staging/protection` on live state
2. Diff old vs new — only diff should be the two new required check names
3. Add R3 check-name preflight
4. Address R1 (extend drift to admin/review settings) + R2 (hard-fail scheduled on missing secret)

Will hold this PR until those land. The docs/e2e-coverage.md piece is solid and could ship separately.

PR #2830, opened 2026-05-05
feat(canvas/terminal): "Not available" banner for runtimes without a TTY

HongmingWang-Rabbit:
## Summary
External-runtime workspaces don't have a shell endpoint, so opening the Terminal tab produced a spurious \"Connection failed\" with a Reconnect button. Now: render a clean \"Terminal not available\" banner with a big terminal-off icon, runtime name, and a pointer to the Chat tab.

## Behavior
- \`data.runtime === \"external\"\` (or any future entry in \`RUNTIMES_WITHOUT_TERMINAL\`) → early-return the banner; xterm + WebSocket are not initialised at all.
- \`claude-code\` and every other runtime → unchanged.
- \`data\` prop omitted → falls through to normal mount (back-compat).

## Test plan
- [x] 4 new vitest cases in \`TerminalTab.notAvailable.test.tsx\`:
  - external renders banner with runtime name
  - external does NOT open a WebSocket (load-bearing — we want zero spurious 404s on the workspace-server)
  - claude-code mounts normally (regression cover)
  - data omitted falls through
- [x] Build + full vitest pass: 1258 tests / 89 files.
- [ ] Browser verify: select an external workspace → click Terminal tab → see the banner with the slashed-terminal icon and \"This workspace runs the external runtime, which doesn't expose a shell\".

### DIFF
--- a/canvas/src/components/SidePanel.tsx
+++ b/canvas/src/components/SidePanel.tsx
@@ -283,7 +283,7 @@ export function SidePanel() {
         {panelTab === "skills" && <SkillsTab key={selectedNodeId} workspaceId={selectedNodeId} data={node.data} />}
         {panelTab === "activity" && <ActivityTab key={selectedNodeId} workspaceId={selectedNodeId} />}
         {panelTab === "chat" && <ChatTab key={selectedNodeId} workspaceId={selectedNodeId} data={node.data} />}
-        {panelTab === "terminal" && <TerminalTab key={selectedNodeId} workspaceId={selectedNodeId} />}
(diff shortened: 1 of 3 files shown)


hongmingwang-moleculeai (approved):
External-runtime workspaces have no /ws/terminal/<id> endpoint, so opening the Terminal tab on them was triggering a 404 → 'Connection failed' state with a Reconnect button — reading as a bug instead of an intentional 'no shell here'. RUNTIMES_WITHOUT_TERMINAL early-returns the not-available banner with a slashed terminal SVG + runtime name + pointer to Chat. Skips xterm + WS init entirely (verified by the WebSocket-constructor-not-called test). 4 vitest cases pin both branches + back-compat. Ship it.

Issue #2829, opened 2026-05-05
RFC: Async-first A2A delegation with durable task ledger


HongmingWang-Rabbit:
# RFC: Async-first A2A delegation with durable task ledger

## The bug we're fixing

Hongming's home hermes workspace, 2026-05-05 03:59 UTC. Caller delegated a multi-iteration task; the workspace was actively working ("iteration 14/90, receiving stream response" at 3:59:21), heartbeating every iteration. At 4:09:03 — **600s after `message/send` started** — the caller saw:

```
FAILED TO DELIVER TO HOME HERMES
TASK: home hermes failed: message/send queued but result not available after 600.0s timeout
UNDERLYING ERROR: message/send queued but result not available after 600.0s timeout
```

Same class as the 60s timeout fix shipped in #113 (memory `feedback_a2a_proxy_60s_idle_timeout_cancels_legit_slow_runs`), just at a higher cliff. Tuning the timeout doesn't change the shape — it picks where the cliff lives.

## Root cause

The platform's A2A proxy holds the caller's `message/send` HTTP connection open until the callee emits a final response part. Three problems with this design:

1. **`message/send` is synchronous-blocking by contract.** Caller waits for *final* response, not for status updates. Heartbeats keep the operator informed but don't satisfy the blocking call.
2. **Proxy treats heartbeats as informational, not keep-alive.** 600s timer counts from `message/send` to final part, not last-traffic. (Intentional — heartbeats can lie, "still working" forever on a stuck agent.)
3. **In-flight state lives in the HTTP connection.** Caller restart, callee restart, proxy timeout, network blip — all kill the delegation. Result emitted after the fail lands in `/dev/null`.

A 90-iteration external-runtime task on stream output will routinely exceed 10 min. Tuning to 1800s just delays the next incident.

## Constraints (from current implementation)

| # | Constraint | Source |
|---|---|---|
| 1 | `delegate_task_async` already returns `task_id` immediately | `workspace/builtin_tools/delegation.py:396` |
| 2 | `check_task_status` polls; heartbeat-aware on QUEUED state | `workspace/builtin_tools/delegation.py:447` |
| 3 | a2a_queue (PR #1870 Phase 1) handles target-busy enqueue, drains on heartbeat | `workspace-server/internal/handlers/a2a_queue.go` |
| 4 | Delegations are recorded to `activity_logs` (event stream), not a durable table | `delegations/record` + `delegations/update` endpoints |
[...]

PR #2836, opened 2026-05-05
feat(delegations): stuck-task sweeper with deadline + heartbeat-staleness rules (RFC #2829 PR-3)

HongmingWang-Rabbit:
Third of 4 PRs implementing [RFC #2829](https://github.com/Molecule-AI/molecule-core/issues/2829). Built on PR-1 (#2832, schema, merged).

## What this PR adds

A periodic sweeper (`DelegationSweeper`) over the `delegations` table:

1. **Deadline-exceeded** rows → marked `failed` (\"deadline exceeded by sweeper\")
2. **Heartbeat-stale** rows (no beat for >10min default) → marked `stuck`

## Why both rules

- Deadline catches forever-heartbeating wedged agents (alive-but-not-advancing class — agent loops on heartbeat call inside its main loop).
- Heartbeat-staleness catches OOM-killed and crashed agents that stop cold without graceful shutdown.

Either rule alone misses one of these classes.

## Order matters

Deadline checked first. A deadline-exceeded AND stale row is marked `failed` (give up), not `stuck` (operator can retry). Semantic difference matters.

## NULL heartbeat is a free pass

A delegation that's just been inserted but hasn't emitted its first heartbeat yet is NOT stuck-marked — gives the agent its first beat window. The deadline naturally catches true never-started rows.

## Concurrent-completion safety

When sweep races with UpdateStatus on a delegation that just completed, the ledger's terminal forward-only protection (PR-1's `ErrInvalidTransition`) holds. Sweeper logs + counts in `Errors`; the row stays correctly in `completed`.

## Configuration

- `DELEGATION_SWEEPER_INTERVAL_S` — tick cadence (default 5min)
- `DELEGATION_STUCK_THRESHOLD_S` — heartbeat-staleness threshold (default 10min)

Invalid input falls back to default — typo'd env shouldn't crash startup.

## Wiring

NOT wired into main.go in this PR — ships separately so the sweeper can be enabled/disabled independently of binary upgrade.

## Coverage

13 unit tests:

**Sweep semantics (8):**
- empty in-flight set → clean no-op
- deadline → failed
- heartbeat-stale → stuck
- NULL heartbeat is left alone (first-beat free pass)
- healthy row → no-op
- both-rule row → marked failed (deadline wins)
- mixed set → both rules fire on the right rows
- concurrent-completion race → forward-only protection holds

**Env override parsing (5):**
- default on missing env
- parses positive seconds
- falls back on garbage
- falls back on negative
- constructor picks up overrides

## Test plan
- [x] 13 unit tests pass
- [x] Existing delegation tests untouched
- [ ] PR wiring sweeper into main.go (follow-up)
- [ ] Staging E2E once sweeper is wired

Refs RFC #2829.

### DIFF
--- a/workspace-server/internal/handlers/delegation_sweeper.go
+++ b/workspace-server/internal/handlers/delegation_sweeper.go
@@ -1,0 +1,265 @@
+package handlers
+
+import (
+	"context"
+	"database/sql"
+	"log"
+	"os"
+	"strconv"
+	"time"
+
+	"github.com/Molecule-AI/molecule-monorepo/platform/internal/db"
+)
+
+// delegation_sweeper.go — RFC #2829 PR-3: stuck-task sweeper.
+//
+// What it does
+// ------------
+// Periodically scans the `delegations` table (PR-1 schema) for in-flight
+// rows that have either:
+//
+//   1. Blown past their `deadline` — agent claims to still be working but
+//      the hard ceiling fired. Mark `failed` with error_detail = "deadline
+//      exceeded".
+//   2. Stopped heartbeating for >stuckThreshold while still claiming
+//      in_progress. Mark `stuck` with error_detail = "no heartbeat for Ns".
+//
+// Why both rules
+// --------------
+// Deadline catches forever-heartbeating agents that never make progress
+// (a wedged agent looping on a heartbeat call inside its main work loop
+// looks "alive" by liveness signals but is not actually advancing).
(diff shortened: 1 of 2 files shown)


hongmingwang-moleculeai (approved):
Self-review: stuck-task sweeper foundation. 13 unit tests, both rules + concurrent-completion safety + env-override parsing covered. Not wired into main.go yet — that ships as separate enable PR.

Issue #12, opened 2026-04-14
Provisioner: preserve Claude session directory across workspace restart


HongmingWang-Rabbit:
## Problem

When a workspace container is restarted via `POST /workspaces/:id/restart`, the Postgres DB keeps `current_session_id` populated, but the container's filesystem — including `/root/.claude/sessions/` — is re-created fresh. On the next agent message, the SDK tries to resume the stored session ID, fails to find the session file, and raises:

```
No conversation found with session ID: <uuid>
Fatal error in message reader: Command failed with exit code 1
SDK agent [claude-code] transient error (attempt 1/3), retrying in 5s: ProcessError
```

Seen repeatedly in production today on PM and Dev Lead after auth-token-rotation-triggered restarts (see incident 2026-04-13).

## Proposed fix

**Default: preserve** (per CEO direction).

1. `provisioner.Start()`: create a named volume `ws-<id>-claude-sessions` if it doesn't exist.
2. Add bind mount to `/root/.claude/sessions/` in the container `HostConfig.Binds`.
3. Volume survives container restart/recreate by default.
4. Opt-out: if env `WORKSPACE_RESET_SESSION=1` is set (or a `?reset=true` query param on restart), skip mounting the volume so the session starts clean.

## Files likely touched

- `platform/internal/provisioner/provisioner.go` (add volume to `Start()`)
- `platform/internal/handlers/workspace_restart.go` (surface the opt-out)
- tests: `provisioner_test.go`

## Priority

High — this is the single most common error we saw across agents today.

## Status

Dev Lead has been briefed to draft design + implementation on branch `fix/preserve-claude-sessions`.

HongmingWang-Rabbit:
Picking this up — adding named volume ws-<id>-claude-sessions mounted at /root/.claude/sessions, with WORKSPACE_RESET_SESSION opt-out. Will open a draft PR within ~30 min.

PR #2846, opened 2026-05-05
feat(handlers): migrate Restart + Resume handlers to dispatchers — #2799 Phase 2 PR-A

HongmingWang-Rabbit:
First part of #2799 Phase 2 (PR-A). Sites 1+2+3 from the audit on PR #2843 are migrated to the SoT dispatchers.

## What changes

| Site | Before | After |
|---|---|---|
| Restart goroutine (1+2) | inline `if h.provisioner != nil { Stop } else if h.cpProv != nil { ... }` + provision branch | `h.RestartWorkspaceAutoOpts(ctx, id, ..., resetClaudeSession)` |
| Resume goroutine (3) | inline `if h.cpProv != nil { provisionWorkspaceCP } else { provisionWorkspace }` | `h.provisionWorkspaceAuto(ws.id, ...)` |

The inline goroutine in Restart was 11 lines of dispatch + comment; collapses to a single dispatcher call inside `go func() { ... }()` with all the rationale moved to the dispatcher's docstring.

## RestartWorkspaceAutoOpts — new variant

Phase 1's `RestartWorkspaceAuto` doesn't carry the `resetClaudeSession` flag (issue #12 — clears the in-container Claude session before restart). Adding it as a parameter:

\`\`\`go
RestartWorkspaceAuto(ctx, wsID, ...)            // wraps RestartWorkspaceAutoOpts(..., false)
RestartWorkspaceAutoOpts(ctx, wsID, ..., reset) // Docker-only knob; CP ignores
\`\`\`

Mirrors the existing \`provisionWorkspace\` / \`provisionWorkspaceOpts\` pair on the provision side.

## Why Resume uses provisionWorkspaceAuto, not RestartWorkspaceAuto

Resume transitions the workspace from \`paused\` (no live container) to \`provisioning\`. There's nothing to Stop. RestartWorkspaceAuto's stop-leg would fire a no-op Stop on a workspace that was never running, adding unnecessary latency on the SaaS path (cpStopWithRetry has up-to-31s of retry budget). \`provisionWorkspaceAuto\` is the right verb here.

## New source-level pins

- \`TestRestartHandler_UsesRestartWorkspaceAuto\` — asserts the Restart handler routes through the dispatcher.
- \`TestResumeHandler_UsesProvisionWorkspaceAuto\` — asserts the Resume handler routes through provisionWorkspaceAuto with \`ws.id\`.

Both prevent regression to the inline dispatch pattern.

## Drift-class fix bundled in

Pre-2026-05-05 the Restart handler inlined dispatch with **Docker-FIRST** ordering — different drift class from the silent-drop bugs PRs #2811/#2824 closed, but same kind of asymmetry from the convention. Migrating to the dispatcher enforces CP-FIRST ordering automatically (matches \`provisionWorkspaceAuto\` / \`StopWorkspaceAuto\` / the docs/architecture/backends.md doc).

## Out of scope (tracked on #2799)

- **Site 4** (\`runRestartCycle\`) — synchronous coordination model doesn't fit the fire-and-return dispatcher. Phase 2 PR-B.
- **Site 5** (Pause loop) — PAUSE doesn't reprovision; needs a new \`PauseWorkspaceAuto\` verb. Phase 3.

## Test plan

- [x] \`go vet ./...\` clean
- [x] All handler tests pass (\`go test ./internal/handlers/...\`)
- [x] Race detector clean (\`go test -race\`)
- [x] Two new source-level pins pass

### DIFF
--- a/workspace-server/internal/handlers/workspace.go
+++ b/workspace-server/internal/handlers/workspace.go
@@ -254,18 +254,34 @@ func (h *WorkspaceHandler) StopWorkspaceAuto(ctx context.Context, workspaceID st
 // runRestartCycle does this synchronously (line ~538) before delegating
 // — match that pattern in any new caller.
 func (h *WorkspaceHandler) RestartWorkspaceAuto(ctx context.Context, workspaceID, templatePath string, configFiles map[string][]byte, payload models.CreateWorkspacePayload) bool {
+	return h.RestartWorkspaceAutoOpts(ctx, workspaceID, templatePath, configFiles, payload, false)
+}
+
+// RestartWorkspaceAutoOpts is the variant that carries Docker-only
+// per-invocation knobs that don't fit on CreateWorkspacePayload. Today
+// the only such knob is resetClaudeSession (issue #12 — clears the
+// in-container Claude session before restart so the agent comes up
+// fresh). CP doesn't have a session-reset concept (each EC2 boots from
+// a fresh image), so the flag is silently ignored on the CP path.
+//
+// Most callers should call RestartWorkspaceAuto (resetClaudeSession=
+// false). The Restart HTTP handler is the one site that exposes the
+// flag to operators — it reads ?reset_session=true from the query
+// string when an operator wants to force a fresh session.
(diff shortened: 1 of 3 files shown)


hongmingwang-moleculeai (approved):
Phase 2 PR-A — clean migration of Sites 1+2+3. RestartWorkspaceAutoOpts shape mirrors provisionWorkspaceOpts; the resetClaudeSession plumbing is the right move. Pin tests gate the migration. Approving.

PR #2854, opened 2026-05-05
feat(delegations): wire ledger Insert+SetStatus from production paths (RFC #2829 #318)

### LINKED ISSUE #318: security: WorkspaceAuth fail-open allows unauthenticated access via non-existent workspace IDs

HongmingWang-Rabbit:
RFC #2829 PR-1 shipped the \`delegations\` table + \`DelegationLedger\` helper. PR-3 wired the sweeper. PR-4 wired the dashboard. **But no PR ever wired \`ledger.Insert\` from a production code path** — table stayed empty, sweeper had nothing to sweep, dashboard had nothing to show.

This PR closes that gap.

## What's wired

Behind feature flag \`DELEGATION_LEDGER_WRITE=1\` (default off):
- \`insertDelegationRow\` → \`ledger.Insert\` (queued)
- \`updateDelegationStatus\` → \`ledger.SetStatus\` on every transition (filtered to ledger-accepted vocabulary)
- \`executeDelegation\` completion path → \`ledger.SetStatus(completed, result_preview)\`
- \`Record\` handler → \`Insert + SetStatus(dispatched)\` (Record's first activity_logs row is dispatched but ledger Insert requires queued; promote on the next status call)

## Why a flag

The legacy flow has ~30 strict-sqlmock tests pinning exactly which SQL fires per handler. Always-on writes would force adding ExpectExec stanzas to each. Flag-off keeps existing tests green without churn; operators flip on staging to feed the sweeper + dashboard.

Default off → byte-identical to pre-#318 behavior.

## Status vocabulary mapping

activity_logs uses a freer status vocabulary than the ledger's CHECK constraint. The wiring filters via a switch to forward only known-good values, skipping anything else.

## Coverage

8 wiring tests:
- flag off → no SQL fired (rollout-safety contract)
- flag on → INSERT + UPDATE fire as expected
- flag rejects loose truthy values (true/yes/0/on/TRUE) — only \"1\" is the signal
- terminal-state replay swallows ErrInvalidTransition (legacy is authoritative)

All 30 existing delegation_test.go tests still pass.

## Test plan
- [x] 8 wiring tests pass
- [x] All 30 existing delegation tests pass
- [x] Build clean
- [ ] CI green
- [ ] Staging: flip \`DELEGATION_LEDGER_WRITE=1\` + create a delegation → \`/admin/delegations\` shows the row

Closes #318. Refs RFC #2829.

### DIFF
--- a/workspace-server/internal/handlers/delegation.go
+++ b/workspace-server/internal/handlers/delegation.go
@@ -269,6 +269,9 @@ func insertDelegationRow(ctx context.Context, c *gin.Context, sourceID string, b
 		VALUES ($1, 'delegation', 'delegate', $2, $3, $4, $5::jsonb, 'pending', $6)
 	`, sourceID, sourceID, body.TargetID, "Delegating to "+body.TargetID, string(taskJSON), idemArg)
 	if err == nil {
+		// RFC #2829 #318 — mirror to the durable delegations ledger
+		// (gated by DELEGATION_LEDGER_WRITE; default off → no-op).
+		recordLedgerInsert(ctx, sourceID, body.TargetID, delegationID, body.Task, body.IdempotencyKey)
 		return insertOK
 	}
 	// A unique-constraint hit means a concurrent request just took the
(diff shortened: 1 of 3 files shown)


hongmingwang-moleculeai (approved):
Self-review: ledger writes wired into 4 production sites behind DELEGATION_LEDGER_WRITE flag. 8 wiring tests + all 30 existing delegation tests still green. Default off keeps strict-sqlmock surface unchanged. Closes the dead-code gap from RFC #2829 PR-1; flag-on populates the dashboard.

PR #2886, opened 2026-05-05
test(rfc): poll-mode chat upload — phase 4 real-Postgres integration

HongmingWang-Rabbit:
## Summary

Phase 4 closes out the poll-mode chat upload rollout. Stack atop merged Phase 1+2 (#2877, #2881) and Phase 3 (#2885). Real-Postgres integration tests for the pendinguploads Storage contracts.

## Why these are NOT plain unit tests

Strict-sqlmock unit tests pin which SQL statements fire — they're fast and let us iterate without a DB. But sqlmock CANNOT detect bugs that depend on the actual row state after the SQL runs. In particular:

- The Sweep CTE depends on Postgres' `make_interval` function and the table's CHECK constraints; sqlmock would happily accept a hand-written SQL literal that Postgres rejects at runtime.
- The partial `idx_pending_uploads_unacked` index (Phase 1 migration) only catches a wrong WHERE predicate at real-query-plan time.
- Subtle predicate drift (e.g. a WHERE clause that filters by `acked_at IS NOT NULL` but uses BETWEEN incorrectly).

These tests close those gaps by booting a real Postgres, running the production helpers, and SELECTing the row to verify the observable state matches the expected outcome.

## Test cases

- `PutGetAckRoundTrip` — Put, Get, MarkFetched, Ack, idempotent re-Ack, Get-after-Ack returns ErrNotFound.
- `Sweep_DeletesAckedAfterRetention` — retention=1h preserves a just-acked row; retention=0 deletes it.
- `Sweep_DeletesExpiredUnacked` — backdated `expires_at` exercises the unacked-expired branch of the WHERE clause.
- `Sweep_DeletesBothCategoriesInOneCycle` — three rows (acked, expired, fresh); a single Sweep deletes the first two and leaves the third.
- `PutEnforcesSizeCap` — ErrTooLarge above MaxFileBytes.
- `GetIgnoresExpiredAndAcked` — Get filter predicate matches expected row state.

## Test plan

- [x] `go build -tags=integration ./...` clean
- [x] `go vet -tags=integration ./internal/handlers/` clean
- [ ] CI: `Handlers Postgres Integration` (the workflow that runs `go test -tags=integration` against a real PG service container) green on this PR

## RFC closure

After this PR merges:

| Phase | PR | What |
|------|----|------|
| 1 | #2877 ✅ | Schema + handler branch + endpoints |
| 2 | #2881 ✅ | Workspace inbox extension |
| 3 | #2885 | GC sweep + observability |
| 4 | this PR | Real-Postgres integration tests |

User-visible: chat upload to poll-mode external workspaces stops returning 422 and routes the file all the way to the agent's filesystem.

### DIFF
--- a/workspace-server/internal/handlers/pending_uploads_integration_test.go
+++ b/workspace-server/internal/handlers/pending_uploads_integration_test.go
@@ -1,0 +1,298 @@
+//go:build integration
+// +build integration
+
+// pending_uploads_integration_test.go — REAL Postgres integration
+// tests for the poll-mode chat upload flow (RFC: phases 1–3).
+//
+// Run with:
+//
+//   docker run --rm -d --name pg-integration \
+//     -e POSTGRES_PASSWORD=test -e POSTGRES_DB=molecule \
+//     -p 55432:5432 postgres:15-alpine
+//   sleep 4
+//   psql ... < workspace-server/migrations/20260505100000_pending_uploads.up.sql
+//   cd workspace-server
+//   INTEGRATION_DB_URL="postgres://postgres:test@localhost:55432/molecule?sslmode=disable" \
+//     go test -tags=integration ./internal/handlers/ -run Integration_PendingUploads
+//
+// CI (.github/workflows/handlers-postgres-integration.yml) runs this on
+// every PR that touches workspace-server/internal/handlers/** OR
+// workspace-server/migrations/**.
+//
+// Why these are NOT plain unit tests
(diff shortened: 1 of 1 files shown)


HongmingWang-Rabbit:
Status: `DIRTY/CONFLICTING`. Likely conflicts with #2877/#2881/#2885 phases that landed earlier. Needs rebase before queue can pick up.

PR #2939, opened 2026-05-05
fix(canvas): include peer-initiated rows in chat history (reno-stars data loss)

HongmingWang-Rabbit:
## Summary

Reported on production tenant **reno-stars** (CEO Ryan PC external workspace): \"received [a message] while the chat was open but it's not persisted.\" Live WebSocket delivery worked; close+reopen tab and full page refresh both lost the message.

## Root cause

\`ChatTab.loadHistory\` filters \`/workspaces/:id/activity\` with \`source=canvas\` (backend: \`source_id IS NULL\`), which returns ONLY canvas-user-initiated rows. **Peer-initiated** \`a2a_receive\` rows — another agent A2A-sending into this workspace, with \`source_id\` set to the peer's UUID — were excluded from history even though they ARE persisted in \`activity_log\`.

Net effect: a real message visible live (via WebSocket), gone on reload.

## Fix

Drop the \`source\` query param from the loadHistory call so both canvas-initiated AND peer-initiated rows return. \`activityRowToMessages\` already handles role assignment per-row from \`request_body\` / \`response_body\`.

## Test plan

- [x] \`npx vitest run src/components/tabs/__tests__/ChatTab\` — 8/8 pass after updating the discriminator (was matching \`source=canvas\`, now matches \"no source filter\" for MyChat)
- [x] \`tsc --noEmit\` clean
- [ ] CI green
- [ ] Approver: \`hongmingwang-moleculeai\`
- [ ] **Manual** post-deploy: send a peer message into a tenant workspace, refresh the chat tab, verify message persists

## Followup (separate PR)

Distinguish peer-initiated bubbles visually (e.g. \"via <peer name>:\") so users can tell their own messages from peer messages. Today both render as \"user\" bubbles — functionally correct (message belongs to this workspace's chat) but visually ambiguous.

### DIFF
--- a/canvas/src/components/tabs/ChatTab.tsx
+++ b/canvas/src/components/tabs/ChatTab.tsx
@@ -153,9 +153,26 @@ async function loadMessagesFromDB(
   beforeTs?: string,
 ): Promise<{ messages: ChatMessage[]; error: string | null; reachedEnd: boolean }> {
   try {
+    // Reported on reno-stars (production, 2026-05-05): "received [a
+    // message] while the chat was open but it's not persisted." Live
+    // delivery via WebSocket worked; refresh + reopen-tab dropped it.
+    //
+    // Root cause: this fetch was filtering `source=canvas` (source_id
+    // IS NULL), which only returns rows where the canvas-user is the
+    // initiator. Peer-initiated rows — another agent A2A-sending into
+    // this workspace — have source_id set to the peer's UUID and were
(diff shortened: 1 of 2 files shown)


HongmingWang-Rabbit:
Closing — diagnosis was wrong. The source=canvas filter is intentional to hide agent-to-agent traffic. Real bug is that the user-targeted messages from external agents aren't being recorded with source_id=NULL. Reopening as separate investigation.

Issue #2962, opened 2026-05-05
fix: byte-slice truncatePreview in delegation_ledger.go has same UTF-8 mid-codepoint bug


HongmingWang-Rabbit:
## Summary

Following self-review of PR #2949 + followup #2959, found the same byte-slice truncation bug in \`workspace-server/internal/handlers/delegation_ledger.go:58\`:

\`\`\`go
const previewCap = 4096

func truncatePreview(s string) string {
    if len(s) <= previewCap {
        return s
    }
    return s[:previewCap]
}
\`\`\`

Same shape as the agent_message bug fixed in #2959:
- \`len()\` is bytes, \`s[:previewCap]\` slices on byte boundary.
- Cap is 4096 bytes, so the bug only triggers when delegated tasks exceed 4096 bytes AND have a multi-byte codepoint at byte 4096.
- For ASCII delegations the bug doesn't fire; for CJK / emoji / mixed-script delegations it can produce invalid UTF-8 → Postgres JSONB rejects → ledger row not inserted → audit gap.

## Fix shape

Reuse \`truncatePreviewRunes\` from agent_message_writer.go (PR #2959), or extract a shared rune-safe truncate helper if more callers exist.

## Test addition

Mirror the table test from \`TestTruncatePreviewRunes_RuneBoundary\` on the delegation-ledger-side helper. Assert \`utf8.ValidString\` on every output.

## Lower-priority than the agent_message fix

The 4096-byte cap is much larger than the 80-byte agent_message cap, so the chance of a real-world hit is lower — most delegated task previews fit under 4KB. But the audit gap mode (silent INSERT failure for any task that triggers it) is identical, and the fix is mechanical.

Refs PR #2959, RFC #2945.

HongmingWang-Rabbit:
## Phase 1 — Investigation (audit)

The original delegation_ledger.go fix landed in commit a5903af4. Audit of every byte-slice truncation across `workspace-server/` reveals this is a class of bug, not a single instance.

| Location | Status | Shape |
|---|---|---|
| `internal/handlers/delegation_ledger.go:70` `truncatePreview` | ✅ SAFE | rune-walk via `for i := range s`, no ellipsis |
| `internal/handlers/agent_message_writer.go:70` `truncatePreviewRunes` | ✅ SAFE | rune-count cap, "…" ellipsis (PR #2959) |
[...]

PR #2964, opened 2026-05-05
fix(delegation_ledger): rune-safe preview truncation (#2962)

HongmingWang-Rabbit:
## Summary

Same bug class as #2959, different call site. The byte-slice form `s[:previewCap]` in `delegation_ledger.go` can split a multi-byte codepoint at byte 4096 → invalid UTF-8 → Postgres JSONB rejects → ledger row insert silently fails → audit gap on dashboards while activity_logs continues to record the event.

Fix: walk the string by rune index and stop at the last boundary that fits inside the cap. ASCII-only strings still hit the cap exactly; CJK/emoji strings stop slightly under, never over.

Mirrors the `truncatePreviewRunes` shape from #2959. Followup: deduplicate into a shared helper once both have landed.

## Test plan

- [x] `TestTruncatePreview_NeverProducesInvalidUTF8` — input is `"世" * (cap/3 + 10)`; assert `utf8.ValidString(got)` and `len(got) <= previewCap` and `strings.HasPrefix(in, got)`.
- [x] `TestTruncatePreview_MultiByteAtBoundary` — input is `"a"*4095 + "世"`; assert exact equality with `"a"*4095` (the 3-byte rune at byte 4095 must be entirely dropped, not partially included). This assertion is discriminating: the old impl would return 4096 bytes (4095 a's + first byte of 世) → test would fail.
- [x] `go test ./internal/handlers/... -run TestTruncatePreview -count=1` — all 5 pass (3 existing + 2 new).

### DIFF
--- a/workspace-server/internal/handlers/delegation_ledger.go
+++ b/workspace-server/internal/handlers/delegation_ledger.go
@@ -53,13 +53,35 @@ func NewDelegationLedger(handle *sql.DB) *DelegationLedger {
 // truncatePreview caps stored preview at 4KB. The full prompt/response is
 // already in activity_logs.{request,response}_body — this is the at-a-glance
 // view for the dashboard, not a forensic record.
+//
+// Rune-safe: previous byte-slice form (s[:previewCap]) split on a byte
+// boundary, which on a multi-byte codepoint at byte 4096 produced
+// invalid UTF-8 — Postgres JSONB rejects → ledger row not inserted →
+// audit gap. Issue #2962. Walks the string by rune, stops at the last
+// rune-boundary index that fits inside the cap. ASCII-only strings hit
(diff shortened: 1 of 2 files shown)


hongmingwang-moleculeai (approved):
Reviewed: 5 axes pass.

Correctness: `for i := range s` over a Go string yields rune-boundary byte indices. Loop walks until next index would exceed cap, last index assigned to `end` is the safe cut point. Edge cases covered — empty, ASCII at cap, multi-byte at boundary.

Tests: discriminating. The MultiByteAtBoundary case asserts exact equality with prefix (`"a"*4095`) which would fail the old byte-slice impl that returns 4096 bytes (4095 a's + 1 byte of 世). Both new tests use utf8.ValidString.

Architecture: matches truncatePreviewRunes in #2959. Followup deduplication tracked.

Approve.

